PR #9, opened 2015-02-21
Add metadata and plugins

zorkian:
This adds a very basic metadata (key, value) storage system, which can
be used by people who want to do things when accounts are set up, etc.
The plugins can be configured in the config_dir and apply to both the
API and FE servers.

### DIFF
--- a/grouper/fe/util.py
+++ b/grouper/fe/util.py
@@ -3,6 +3,7 @@ from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
 from expvar.stats import stats
 from jinja2 import Environment, PackageLoader
+import logging
 import pytz
 import re
 import smtplib
@@ -60,9 +61,9 @@ class GrouperHandler(tornado.web.RequestHandler):
             raise InvalidUser()
 
         try:
-            user = self.session.query(User).filter_by(username=username).first()
-            if not user:
-                user = User(username=username).add(self.session)
+            user, created = User.get_or_create(self.session, username=username)
+            if created:
+                logging.info("Created new user %s", username)
                 self.session.commit()
         except sqlalchemy.exc.OperationalError:
             # Failed to connect to database or create user, try to reconfigure the db. This invokes
@@ -140,7 +141,7 @@ class GrouperHandler(tornado.web.RequestHandler):
         msg.attach(html)
 
         if not settings["send_emails"]:
-            print msg.as_string()
+            logging.debug(msg.as_string())
             return
 
         smtp = smtplib.SMTP(settings["smtp_server"])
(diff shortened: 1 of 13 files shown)


zorkian:
Even has tests on the metadata functions ;)

zorkian:
@chet boom!

grouper/fe/util.py
```diff
@@ -60,9 +60,9 @@ def get_current_user(self):
             raise InvalidUser()
 
         try:
-            user = self.session.query(User).filter_by(username=username).first()
-            if not user:
-                user = User(username=username).add(self.session)
+            user, created = User.get_or_create(self.session, username=username)
+            if created:
+                print "Created new user %s" % (username, )
```
jathanism: `Should that `print still be in there? It seems out of place.

grouper/models.py
```diff
@@ -118,6 +130,17 @@ def get_db_engine(url):
...
 
 
+Plugins = []
+
+
+def load_plugins(plugin_dir):
+    """Load plugins from a directory"""
+    global Plugins
+    if Plugins:
+        raise Exception("Plugins already loaded; can't load twice!")
```
jathanism: I feel like this should have its own exception. Something like:

```python
raise PluginsAlreadyLoaded("Plugins already loaded; can't load twice!")
```

That way you can catch that exception? Base `Exception` can be hard to isolate.

zorkian: Is it general good practice to make lots of named classes for
exceptions? I don't know what "best" is :)

--
Mark Smith mark@qq.is


On Sat, Feb 21, 2015, at 03:33 PM, Jathan McCollum wrote:
> In grouper/models.py[1]:


> > @@ -118,6 +130,17 @@ def get_db_engine(url):
>      return create_engine(url, pool_recycle=300)
>
>
> +Plugins = [] + + +def load_plugins(plugin_dir):
> +    """Load plugins from a directory"""
> +    global Plugins
> +    if Plugins:
> +        raise Exception("Plugins already loaded; can't load twice!")
>          I feel like this should have its own exception. Something
>          like:


> raise PluginsAlreadyLoaded("Plugins already loaded; can't load
> twice!") That way you can catch that exception? Base Exception can be
> hard to isolate.


> — Reply to this email directly or view it on GitHub[2].

>



Links:

  1. https://github.com/dropbox/grouper/pull/9#discussion_r25127047
  2. https://github.com/dropbox/grouper/pull/9/files#r25127047

jathanism: In short: yes. That way for the error-handling you can be very specific.

zorkian:
Updated, thanks @jathanism !

PR #11, opened 2015-04-02
Add permissions method for UI

zorkian:
This promotes permissions to be represented like users/groups, i.e., to
have a little icon and a custom UI method to allow us to put styling
on these. (In the future we can do more like have JS popups and other
status information.)


zorkian:
@tmildorf ping!

PR #12, opened 2015-04-03
Let the frontend have the graph

zorkian:
It was confusing that the permissions view would only show direct
permissions, instead of letting you see all permissions. Instead of
walking the graph every time we needed this data, let's just load the
graph into the frontend worker with a customizable refresh.

This is potentially adding a lot of data/load to the frontend, and
now every fe worker has to load the graph. In my testing this seems
reasonable. There are things we can do to avoid the FE loading extra
metadata if we need to.

### DIFF
--- a/grouper/fe/settings.py
+++ b/grouper/fe/settings.py
@@ -22,6 +22,7 @@ settings = FeSettings.from_settings(base_settings, {
     "date_format": "%Y-%m-%d %I:%M %p",
     "cdnjs_prefix": "//cdnjs.cloudflare.com",
     "user_auth_header": "X-Grouper-User",
+    "refresh_interval": 60,
     "send_emails": True,
     "smtp_server": "localhost",
     "from_addr": "no-reply@grouper.local",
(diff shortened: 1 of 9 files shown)


zorkian:
@tmildorf ... pong!

grouper/fe/settings.py
```diff
@@ -22,6 +22,7 @@ def override_timezone(self, value):
     "date_format": "%Y-%m-%d %I:%M %p",
     "cdnjs_prefix": "//cdnjs.cloudflare.com",
     "user_auth_header": "X-Grouper-User",
+    "refresh_interval": 60,
```
tmildorf: This seems pretty infrequent; does this mean it'll be a full minute for each update when setting up new users?

zorkian: This is just the default. Settings are maintained in the config file, and we can set this to whatever we want there. In prod we set it to 10s right now, but could do w/e.

tmildorf:
Graph-in-frontend pulled with another commit.

PR #14, opened 2015-04-09
Make public keys viewable

zorkian:
This makes it possible to view the full contents of a public key from
the UI, making it possible for people who have manual implementations to
get keys.


zorkian:
@tmildorf another one, this is a higher priority one for users :)

PR #16, opened 2015-04-10
Add some more text + on/off button

zorkian:
This adds a little more descriptive text for auditing so people can tell
when they're trying to join an audited group, and when a permission is
audited. This also puts an enable/disable button for permissions admins
so that we can turn the functionality on or off for a permission.


zorkian:
@tmildorf anudder one

tmildorf:
CI test failures?

zorkian:
Wow, good to know. Fixed.

PR #19, opened 2015-04-14
Use bulk queries

zorkian:
This codepath was selecting public keys and user metadata for each user,
leading to doing hundreds of queries to refresh user data. This is
pretty slow.

### DIFF
--- a/grouper/graph.py
+++ b/grouper/graph.py
@@ -9,7 +9,7 @@ from sqlalchemy.orm import aliased
 from sqlalchemy.sql import label, literal
 
 from .models import (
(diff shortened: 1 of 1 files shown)


tmildorf:
Ack.

grouper/graph.py
```diff
@@ -113,6 +113,18 @@ def _get_user_metadata(session):
             User.enabled == True
         )
 
+        public_keys = {}
```
tmildorf: You could use a collections.defaultdict(list) for this pattern.

tmildorf: LGTM otherwise.

PR #21, opened 2015-04-16
Fill out FAQ-style help page.

tmildorf:
First cut, now without polluting upstream with development branches.

### DIFF
--- a/grouper/fe/templates/help.html
+++ b/grouper/fe/templates/help.html
@@ -10,13 +10,90 @@
 
 {% block content %}
 
-<!-- TODO(mildorf): make this more helpful...  -->
+<!-- TODO(mildorf): clean up the style here, it's not 1995. -->
+<!-- TODO(mildorf): keep this in sync with reality. -->
 
 <h3>What is Grouper?</h3>
-<p>This is a group membership and permission management platform. In other words, we are an
-authorization service you can use to answer such questions as: who's in what groups, and what
-permissions do they have?</p>
+<p>Grouper is a group membership and permission management platform. In other words, an
+authorization service you can use to answer such questions as: who's in what groups, and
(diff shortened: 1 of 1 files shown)


grouper/fe/templates/help.html
```diff
@@ -10,13 +10,90 @@
...
-authorization service you can use to answer such questions as: who's in what groups, and what
-permissions do they have?</p>
+<p>Grouper is a group membership and permission management platform. In other words, an
+authorization service you can use to answer such questions as: who's in what groups, and
+what permissions do they have? However, Grouper itself is just a system for permissions
+specfication. It's up to Grouper clients to actually implement things like access
+control.</p>
 
+<h3>How does group membership work?</h3>
+<p>Users and groups are the nodes of a directed acyclic graph.  The users are always leaf
```
zorkian: I'm not sure end users (non-engineers, managers a lot of the time) will know what a DAG is!

tmildorf: I was on the fence about adding a link to wikipedia.  Sounds like I should!

tmildorf: Will be cleaned up in separate diff.

zorkian:
Some general feedback,

This documentation needs to be designed for non-technical end users. We can't refer to things as nodes/leaves, DAGs, or anything like that. It has to be human/approachable for everybody, even the Design Internet who shows up and their only computer experience is Google Docs.

Second, we need to make it possible to link to our internal documentation. I was thinking that we could make that a tunable in the settings, and/or we could make it part of our internal build process (i.e. drop in our own help.html).

But mostly, we definitely need to approach this from an end-user perspective, and we should probably get croberson to look at it to make sure it seems OK with whatever verbiage we make.

PR #25, opened 2015-04-16
Answer some ?s, add links, and add the ability to list administrators and documents.

tmildorf:
Second cut.  Address excess jargon and add ways to insert current admins and site-specific documentation.

### DIFF
--- a/grouper/fe/templates/help.html
+++ b/grouper/fe/templates/help.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% from 'macros/ui.html' import permission with context %}
 
 {% block heading %}
     Help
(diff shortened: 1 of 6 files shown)


tmildorf:
Bump.

zorkian:
@tmildorf Sorry, was on vacation. Am back.

grouper/fe/templates/help.html
```diff
@@ -17,8 +18,95 @@
...
+<!-- TODO: change once auditing tooling is ready! -->
+<p>We're working on tooling to facilitate regular monitoring of both members of audited
+groups and groups with audited permissions.</p>
+
+<h3>How do I set up Grouper?</h3>
+<p>See the README that comes with Grouper. Basically, after a pip install, you need to
+configure grouper to find a SQL-style backing database and stand up processes to serve the
+read-write web UI and read-only programmatic API.</p>
+
+<h3>How do I set up the first groups and permissions?</h3>
```
zorkian: This should be in the README for admins, not on the web site. The web site should contain help for end-users, and whoever sets up Grouper already had to use grouper-ctl to, say, populate the DB etc. So, I don't think we should have this information on the FAQ.

grouper/fe/templates/help.html
```diff
@@ -17,8 +18,95 @@
...
+with at least one audited permission become audited groups.  Users with the
+{{permission(audit_perm)}} permission (any argument) are considered auditors.  Only
+auditors can be the manager or owner of an audited group.</p>
+
+<!-- TODO: change once auditing tooling is ready! -->
+<p>We're working on tooling to facilitate regular monitoring of both members of audited
+groups and groups with audited permissions.</p>
+
+<h3>How do I set up Grouper?</h3>
+<p>See the README that comes with Grouper. Basically, after a pip install, you need to
```
zorkian: Same comment here. We shouldn't have setup information on the help page. This is only viewable after you have the site set up.

grouper/fe/templates/help.html
```diff
@@ -17,8 +18,95 @@
...
+permission.</p>
+
+<h3>What if my questions aren't answered here?</h3>
+<p>This list isn't comprehensive. Please contact the admins of this site ({{
+  current_admins }}) for help and
+  {% if site_docs %}
+  see the documentation:
+  <ul>
+    {% for doc in site_docs %}
+    <li><a href="{{doc.url}}">{{doc.name}}</a> - {{doc.description}}.</li>
```
zorkian: The trailing period here is kind of weird, I suspect if they want a period they will put it in the description.

grouper/fe/templates/help.html
```diff
@@ -17,8 +18,95 @@
...
+<p>A <strong>group_admin</strong> has the power to manipulate the graph structure at will.
+For example, they can create any group, add any user to any group as any role, provided
+that auditing requirements aren't violated.  A <strong>user_admin</strong> has the power
+to disable/enable a user account but no special power with respect to group
+membership. Finally, a <strong>permission_admin</strong> has the power to create any
+permission.</p>
+
+<h3>What if my questions aren't answered here?</h3>
+<p>This list isn't comprehensive. Please contact the admins of this site ({{
+  current_admins }}) for help and
```
zorkian: So `current_admins` is defined as a string and just inserted here directly. Do you expect that string to contain HTML? I would expect it to be, say, an email address and then you could automatically linkify it here. Thoughts?

grouper/fe/templates/help.html
```diff
@@ -17,8 +18,95 @@
...
+'.' as a separator.</p>
+
+<h3>Who can grant permissions?</h3>
+<p>Generally, a user with the permission ({{permission(grant_perm)}}, prefix.*) can grant
+any permission starting with "prefix." with any argument.  As a special case, the
+permission ({{grant_perm.name}}, {{grant_perm.name}}/prefix.*) allows a user to
+grant the permission to grant ({{grant_perm.name}}, prefix.*).</p>
+
+<h3>How does auditing work?</h3>
+<!-- FIXME(mildorf): does the argument matter for grouper.auditor.permission? -->
```
zorkian: The argument means nothing for that permission.

grouper/fe/templates/help.html
```diff
@@ -17,8 +18,95 @@
...
+<p>Grouper permissions have a bit that defines whether they are audited or not.  Groups
```
zorkian: Non-programmer end-user says "what is a bit"?

zorkian:
Landing. I have the feedback above so it'd be nice to get a commit to address those, but it's substantially better than what we have, so

tmildorf:
Resolved conflicts.  Will address comments before landing.

grouper/fe/templates/help.html
tmildorf: No, it shouldn't contain HTML.  Is there a way to ensure that a string embeds as literal text?  The thought was that different organizations might have different ideas about reaching out for help, not necessarily by email, so that less is more in this case.  I'll change it to something more descriptive like "how_to_get_help" with the intent to configure it as something like "email user1@, user2@, and team@, or reach out on #chat."

PR #26, opened 2015-04-16
Require a '.' in UI-created permissions.

tmildorf:
Bugfix for the fight over /permissions/create by requiring some namespacing.

### DIFF
--- a/grouper/constants.py
+++ b/grouper/constants.py
@@ -31,6 +31,7 @@ RESERVED_NAMES = [
     r"^grouper",
     r"^admin",
     r"^test",
+    r"^[^.]*$",
 ]
 
 # Maximum length a name can be. This applies to user names and permission arguments.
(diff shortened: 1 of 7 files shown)


zorkian:
Can we add unit tests if we're going to update these regexes? They get scary, and it'd be good to make sure they work. Also, I believe this regex is used by other parts of the system besides permissions.

Anyway, in this particular case, I was thinking we would implement this in terms of adding to the `RESERVED_NAMES` list. I.e., make one that's like `r"^[^.]$"` or something.

tmildorf:
+10 to unit tests for anything regex-related.

+1 That's a much-less-intimidating regex to show a user.

zorkian:
This pull request seems to contain some extra commits? Can you rebase/clean it so it will be able to merge?

PR #27, opened 2015-04-21
Require a '.' in UI-created permissions.

tmildorf:
Add r"^[^.]*$" to the list of reserved names to block UI-created permissions that lack a period because namespacing is good.  Also add some unit tests.

### DIFF
--- a/tests/test_permissions.py
+++ b/tests/test_permissions.py
@@ -1,3 +1,7 @@
+import unittest
+
+import grouper.fe.util
+
 from fixtures import standard_graph, graph, users, groups, session, permissions  # noqa
 from util import get_group_permissions, get_user_permissions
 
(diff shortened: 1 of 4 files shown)


tests/test_permissions.py
```diff
@@ -23,3 +27,10 @@ def test_basic_permission(standard_graph, session, users, groups, permissions):
     assert sorted(get_user_permissions(graph, "zorkian")) == [
         "audited:", PERMISSION_AUDITOR + ":", "ssh:*", "sudo:shell"]
     assert sorted(get_user_permissions(graph, "testuser")) == []
+
+class PermissionTests(unittest.TestCase):
+    def test_reject_bad_permission_names(self):
+        self.assertEquals(len(grouper.fe.util.test_reserved_names("permission_lacks_period")), 1)
```
zorkian: No positive test to make sure a name is valid?

PR #30, opened 2015-04-24
Make permission granting check audited owners

zorkian:
This brings in the lgoic used to determine whether or not a group (and
all subgroups) are controlled by auditors into the permission granting
process. With this change, you can only grant an audited permission to a
group that is controlled entirely by auditors.

### DIFF
--- a/grouper/audit.py
+++ b/grouper/audit.py
@@ -25,7 +25,49 @@ def user_is_auditor(username):
     return False
 
 
-def can_join(group, user_or_group, role="member"):
+def assert_controllers_are_auditors(group):
+    """Return whether not all owners/managers in a group (and below) are auditors
+
+    This is used to ensure that all of the people who can control a group (owners, managers)
+    and all subgroups (all the way down the tree) have audit permissions.
+
+    Raises:
+        UserNotAuditor: If a user is found that violates the audit training policy, then this
+            exception is raised.
+
+    Returns:
+        bool: True if the tree is completely controlled by auditors, else it will raise as above.
+    """
+    graph = Graph()
+    checked, queue = set(), [group.name]
+    while queue:
+        cur_group = queue.pop()
+        if cur_group in checked:
+            continue
+        details = graph.get_group_details(cur_group)
+        for chk_user, info in details["users"].iteritems():
+            if chk_user in checked:
+                continue
+            # Only examine direct members of this group, because then the role is accurate.
+            if info["distance"] == 1:
+                if info["rolename"] == "member":
+                    continue
+                if user_is_auditor(chk_user):
+                    checked.add(chk_user)
+                else:
+                    raise UserNotAuditor(
+                        "User {} has role {} of the {} group, but is not an auditor.".format(
+                            chk_user, info["rolename"], cur_group))
+        # Now put subgroups into the queue to examine.
+        for chk_group, info in details["subgroups"].iteritems():
+            if info["distance"] == 1:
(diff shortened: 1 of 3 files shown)


tmildorf:
I was planning to check for this manually.  Bravo!

grouper/audit.py
```diff
@@ -25,6 +25,48 @@ def user_is_auditor(username):
...
+
+    This is used to ensure that all of the people who can control a group (owners, managers)
+    and all subgroups (all the way down the tree) have audit permissions.
+
+    Raises:
+        UserNotAuditor: If a user is found that violates the audit training policy, then this
+            exception is raised.
+
+    Returns:
+        bool: True if the tree is completely controlled by auditors, False if not.
```
tmildorf: False is never returned.

grouper/audit.py
```diff
@@ -25,6 +25,48 @@ def user_is_auditor(username):
...
+            if chk_user in checked:
+                continue
+            # Only examine direct members of this group, because then the role is accurate.
+            if info["distance"] == 1:
+                if info["rolename"] == "member":
+                    continue
+                if user_is_auditor(chk_user):
+                    checked.add(chk_user)
+                else:
+                    raise UserNotAuditor(
```
tmildorf: Would it be better to report all of the violations at once?  Conceivably there could be a large number of subgroups and it would be pesky to find out about them one by one.

grouper/audit.py
```diff
@@ -36,6 +78,10 @@ def can_join(group, user_or_group, role="member"):
         user (models.User): The user attempting to join.
         role (str): The role being tested.
 
+    Raises:
+        UserNotAuditor: If a user is found that violates the audit training policy, then this
+            exception is raised.
+
     Returns:
         bool: True if the user should be allowed per policy, False if not.
```
tmildorf: Also never returns False.

grouper/audit.py
```diff
@@ -25,6 +25,48 @@ def user_is_auditor(username):
...
+                continue
+            # Only examine direct members of this group, because then the role is accurate.
+            if info["distance"] == 1:
+                if info["rolename"] == "member":
+                    continue
+                if user_is_auditor(chk_user):
+                    checked.add(chk_user)
+                else:
+                    raise UserNotAuditor(
+                        "User {} is a {} of the {} group, but is not an auditor.".format(
```
tmildorf: nit: grammar.  This can generate "is a owner", which makes me sad.  You could delete " a" to make it somewhat better, or conditionally switch to " an" for best effect.

tmildorf:
LGTM apart from the minor style/grammar points.

grouper/audit.py
zorkian: Fixed.

grouper/audit.py
zorkian: Updated.

grouper/audit.py
zorkian: Changed names.

grouper/audit.py
zorkian: Yeah, I tried this originally but the UI kind of does weird things when the error message is really long, so I took it out. I'll try to come up with something better.

tmildorf: Maybe scribble a list into the assertion, then convert said list into HTML <ul> and <li> elements?  Does that get handled OK.

zorkian: I will try in an upcoming patch!

PR #32, opened 2015-05-08
First pass of implementing auditing

zorkian:
This creates an "audits" framework that allows someone to start a global
audit. When this is done we take a point in time snapshot of all the
members of all audited groups at that time and then set up an audit for
each group.

These audits are then displayed to group owners when they visit their
groups, and they're forced to go through a process whereby they
approve/remove everybody in their group.

Once done, they mark the audit as complete and the statuses are checked
and then enforced.

### DIFF
--- a/grouper/fe/handlers.py
+++ b/grouper/fe/handlers.py
@@ -1,28 +1,31 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 import operator
 
 from expvar.stats import stats
 from tornado.web import RequestHandler
 
-from sqlalchemy import union_all, or_, and_
+from sqlalchemy import union_all
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql import label, literal
 
 import sshpubkey
 
 from ..audit import assert_controllers_are_auditors, assert_can_join, UserNotAuditor
-from ..constants import PERMISSION_GRANT, PERMISSION_CREATE, PERMISSION_AUDITOR
+from ..constants import (
+    PERMISSION_GRANT, PERMISSION_CREATE, PERMISSION_AUDITOR, AUDIT_MANAGER, AUDIT_VIEWER
+)
 
 from .forms import (
     GroupCreateForm, GroupEditForm, GroupJoinForm, GroupAddForm, GroupRemoveForm,
     GroupRequestModifyForm, PublicKeyForm, PermissionCreateForm,
-    PermissionGrantForm, GroupEditMemberForm,
+    PermissionGrantForm, GroupEditMemberForm, AuditCreateForm,
 )
 from ..graph import NoSuchUser, NoSuchGroup
 from ..models import (
     User, Group, Request, PublicKey, Permission, PermissionMap, AuditLog, GroupEdge, Counter,
     GROUP_JOIN_CHOICES, REQUEST_STATUS_CHOICES, GROUP_EDGE_ROLES, OBJ_TYPES,
-    get_all_groups, get_all_users, get_user_or_group,
+    get_all_groups, get_all_users,
+    get_user_or_group, Audit, AuditMember, AUDIT_STATUS_CHOICES,
 )
 from .settings import settings
 from .util import GrouperHandler, Alert, test_reserved_names
@@ -365,7 +368,7 @@ class PermissionsView(GrouperHandler):
     Controller for viewing the major permissions list. There is no privacy here; the existence of
     a permission is public.
     '''
-    def get(self):
+    def get(self, audited_only=False):
         offset = int(self.get_argument("offset", 0))
         limit = int(self.get_argument("limit", 100))
         audited_only = bool(int(self.get_argument("audited", 0)))
@@ -374,7 +377,7 @@ class PermissionsView(GrouperHandler):
 
         permissions = self.graph.get_permissions(audited=audited_only)
         total = len(permissions)
-        permissions = permissions[offset:offset+limit]
+        permissions = permissions[offset:offset + limit]
 
         can_create = self.current_user.my_creatable_permissions()
 
@@ -502,6 +505,7 @@ class GroupView(GrouperHandler):
             "group.html", group=group, members=members, groups=groups,
             num_pending=num_pending, alerts=alerts, permissions=permissions,
             log_entries=log_entries, grantable=grantable, audited=audited,
+            statuses=AUDIT_STATUS_CHOICES,
         )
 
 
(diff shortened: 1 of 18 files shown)


zorkian:
The tests don't pass. This is still missing functionality like notification emails. But this is a start and I wanted to get it in front of you just so you can start asking questions or poking if there are critical implementation issues, etc. DO NOT MERGE.

grouper/models.py
```diff
@@ -898,6 +905,100 @@ def __repr__(self):
...
+        elif self.edge.member_type == 1:  # Group
+            return Group.get(self.session, pk=self.edge.member_pk)
+        raise Exception("invalid member_type in AuditMember!")
+
+
+class Audit(Model):
+    """An Audit is applied to a group for a particular audit period
+
+    This contains all of the state of a given audit including each of the members who were
+    present in the group at the beginnign of the audit period.
```
tmildorf: Typo.

grouper/fe/handlers.py
```diff
@@ -717,6 +718,136 @@ def get(self, group_id=None, name=None):
...
+            )
+
+        # TODO: Make it so only certain users can start audits
+        user = self.get_current_user()
+
+        # Step 1, detect if there are non-completed audits and close them.
+        open_audits = self.session.query(Audit).filter(
+            Audit.complete == False).all()
+        for audit in open_audits:
+            audit.complete = True
```
tmildorf: If I understand correctly, this is recording that "the last round of group audits finished normally" even though there may be pending and/or unactioned statuses.  Should we record something more detailed about the status of such a past group audits?

grouper/fe/handlers.py
```diff
@@ -717,6 +718,136 @@ def get(self, group_id=None, name=None):
         )
 
 
+class AuditsComplete(GrouperHandler):
```
tmildorf: I'm not generally a fan of the overloading of "auditing", but if it's standard parlance I can learn to live with it.  To me, "an audit" should refer unambiguously to the top-level concept that applies when auditors visit an institution.  As such, I'm in favor of replacing "group audit" with a synonym like "probe" or "review", depending on how scary/invasive we want it to sound.  It's not so obvious from its name whether this model is going to refer to a view of past top-level audits, the status of current group audits, or something else.

grouper/fe/templates/audit-create.html
```diff
@@ -0,0 +1,62 @@
...
+{% block content %}
+<div class="col-md-6 col-md-offset-3">
+    <div class="panel panel-default">
+        <div class="panel-heading">
+            <h3 class="panel-title">Start Global Audit</h3>
+       </div>
+        <div class="panel-body">
+            <form class="form-horizontal" role="form"
+                  method="post" action="/audits/create">
+                    <p>Submitting this form will create a new global audit. This will force
```
tmildorf: I like that this page explains what's going on.  I think it would be better to say that an audit comprises separate "reviews" rather than "audits" for each group.

grouper/fe/handlers.py
```diff
@@ -717,6 +718,136 @@ def get(self, group_id=None, name=None):
...
+        if not audit.completable:
+            return self.redirect('/groups/{}'.format(audit.group.name))
+
+        # Complete audits have to be "enacted" now. This means anybody marked as remove has to
+        # be removed from the group now.
+        for member in audit.my_members():
+            if member.status == "remove":
+                audit.group.revoke_member(self.current_user, member.member,
+                                          "Revoked as part of audit.")
+                AuditLog.log(self.session, self.current_user.id, 'remove_member',
```
tmildorf: Should we record "Membership confirmed as part of audit." for members who do belong?  Mostly we'd expect audits to find "things are as they should be", at least at a properly-managed institution.

grouper/fe/handlers.py
zorkian: This is what they're called, however; the organization performs an internal audit, and then it exposes the results to the external auditors.

I mean, I don't care too much, but it at least aligns the two processes so it's clear they're related?

grouper/fe/handlers.py
zorkian: I thought about that, but I think it's going to balloon the logs in the steady state, and it doesn't add any value. No state changed, so we can just implicitly say "user X was a member during the audit period, so they were audited". (I ran this by Tolga and he was happy with it.)

grouper/fe/handlers.py
zorkian: Actually we should probably just remove this. It's helpful in testing but I can make that a CLI option; in the real world, we'd probably be in trouble for "voiding" an ongoing audit?

grouper/fe/templates/audit-create.html
zorkian: Will revise.

grouper/models.py
zorkian: Eagle eyes!

grouper/fe/handlers.py
tmildorf: Agreed.  An ongoing audit should not quietly be finished.

grouper/fe/handlers.py
tmildorf: That sounds fair.

grouper/fe/handlers.py
tmildorf: Agreed, though the best information there comes from what we tell people on the audit-create template.  We can update the FAQ correspondingly.

grouper/models.py
```diff
@@ -472,6 +474,9 @@ class Group(Model):
     canjoin = Column(Enum(*GROUP_JOIN_CHOICES), default="canask")
     enabled = Column(Boolean, default=True, nullable=False)
 
+    audit_id = Column(Integer, ForeignKey("audits.id"), nullable=True)
```
tmildorf: By distinguishing "[global] audits" and "[group] reviews", one way to resolve the circular dependency would be to have groups point at audits, reviews point at groups and audits, and a relationship with backref between reviews and audits.

zorkian:
@mildorf We should chat about notifications. I don't want to build some sort of notification queueing system if you've already got one in mind, etc.

tmildorf:
Big todo: I didn't spot an obvious links to see a list of current audits or where to create a new audit, so I ended up peeking in routes.py for a hint.

Small todo:  we should update the FAQ with appropriate references to grouper.audit.view and grouper.audit.manage.

Small bug: I can remove group owners during an audit, leaving an empty group!

There might have been something else related to getting into an audited group after having disabled one audited permission but not another, I'll keep exploring. It's tricky to test UIs manually!

zorkian:
Updated now that we have async emails to send emails. Also cancels pending emails when the user finishes the audit.

This still doesn't add a link to Audits, as it's weirdly non-trivial to do so. We can't add it to the base navigation layer because that layer doesn't have user information, and it felt weird to add it to a profile page? I suppose we could but that's kind of gross.

tmildorf:
Overall it looks good, but there are lots of little things:

* bug: can *add* (but not edit to) an owner who doesn’t have grouper.permissions.auditors as owner of audited group (they do get appropriately 403’d when trying to do the audit for the group)
* bug: 500 on Start Global Audit button when there’s an audit in progress: 
```Traceback (most recent call last):
  File "/Users/mildorf/src/grouper/.env/lib/python2.7/site-packages/tornado/web.py", line 1218, in
_when_complete
    callback()
  File "/Users/mildorf/src/grouper/.env/lib/python2.7/site-packages/tornado/web.py", line 1239, in
_execute_method
    self._when_complete(method(*self.path_args, **self.path_kwargs),
  File "/Users/mildorf/src/grouper/grouper/fe/handlers.py", line 820, in post
    raise Exception("Sorry, there are audits in progress.")
Exception: Sorry, there are audits in progress.```
* bug: user with grouper.audit.view but not grouper.audit.manage can see the Start Global Audit button, which just serves them 403.
* nit: `/audits` Ends column should be empty or “Completed at XXX” once completed.
* nit: display time zone of the due date (everything is UTC, but I see ends at 5pm, which is just pacific for UTC midnight)
* nit: “remove” is an option for self when filling out the form (though it is appropriately ignored)

zorkian:
Updated with requested changes, except the TZ one. That's in the config file, and it seems OK in prod?

zorkian:
Ping?

PR #33, opened 2015-05-15
Add CLI methods create, add_public_key

zorkian:
This allows us to both create users and add public keys for users from
the command line.

Also fix a bug where deletes were logged on the wrong account.

### DIFF
--- a/bin/grouper-ctl
+++ b/bin/grouper-ctl
@@ -7,6 +7,7 @@ import BaseHTTPServer
 import code
 import getpass
 import logging
+import sshpubkey
 from mrproxy import UserProxyHandler
 from pprint import pprint
 from sqlalchemy.exc import IntegrityError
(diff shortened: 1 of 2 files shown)


bin/grouper-ctl
```diff
@@ -194,6 +228,15 @@ def main():
     user_parser.set_defaults(func=user_command)
     user_subparser = user_parser.add_subparsers(dest="subcommand")
 
+    user_create_parser = user_subparser.add_parser(
```
tmildorf: I'm having trouble getting this help info for sub commands to display. I've tried grouper-ctl --help, grouper-ctl help, grouper-ctl --help sync_db, grouper-ctl help sync_db, grouper-ctl sync_db --help; am I missing the right permutation?

tmildorf:
$ bin/grouper-ctl -v -c config/dev.yaml user add_public_key alex@example.com AAAA**SNIP**==
No such user alex@example.com

bin/grouper-ctl
zorkian: Just type whatever and append `-h`:

```
bin/grouper-ctl user -h
bin/grouper-ctl user set_metadata -h
```

etc.

zorkian:
Did you create the user? If the user doesn't exist, then you can't set a key on them...

tmildorf:
I succeeded by using two separate commands.

zorkian:
Yeah; `create` is the only way to create a user. I wanted it to be an explicit command, rather than having someone typo a username when trying to add a public key.

tmildorf:
My impression from skimming the code was that I should be able to figure out how to specify both `create` and `add_public_key` subcommands but that may not be supported.  It's a minor point.  I approve.

tmildorf:
Making the help a little easier to get through can be addressed separately.  The desired functionality is the key here.

PR #34, opened 2015-05-15
Async notifications

zorkian:
This is an initial implementation of an AsyncNotification class.
Basically, you can call send_async_email and specify a time/key and then
an email will be scheduled to go out later.

This adds a thread to the API server that causes it to send emails from
the queue every minute or so and then mark them sent. (Note: this has a
race condition, if two API servers happen to be on the exact same thread
cycle... we might send duplicate emails.)

Submitting this now to get feedback to see if this is what @mildorf is
thinking.


zorkian:
Is this what you were thinking @mildorf?

bin/grouper-api
```diff
@@ -29,6 +30,34 @@ class Application(tornado.web.Application):
...
 
+class SendEmailThread(Thread):
+    def run(self):
+        while True:
+            logging.debug("Sending emails...")
+            try:
+                now = datetime.utcnow()
+                session = Session()
+                emails = session.query(AsyncNotification).filter(
+                    AsyncNotification.sent == False).all()
```
tmildorf: To avoid a race, chain with (send_after < now) and preemptively update to sent=True, reverting to sent=False for those emails that hit the exception handler.

bin/grouper-api
```diff
@@ -29,6 +30,34 @@ class Application(tornado.web.Application):
...
+                for email in emails:
+                    if email.send_after < now:
+                        try:
+                            send_email_raw(settings, email.email, email.body)
+                            email.sent = True
+                        except:
+                            # Any sort of error with sending the email and we want to move on to
+                            # the next email. TODO: Is there a way to enumerate all of the errors
```
tmildorf: Re TODO:  I'm not aware of a way to enumerate all possible exceptions.

bin/grouper-api
```diff
@@ -29,6 +30,34 @@ class Application(tornado.web.Application):
         super(Application, self).__init__(*args, **kwargs)
 
 
+class SendEmailThread(Thread):
+    def run(self):
+        while True:
+            logging.debug("Sending emails...")
+            try:
+                now = datetime.utcnow()
```
tmildorf: For unit testing, I think it's preferable to split out the contents of this loop to a separate function that takes time as an argument.

grouper/fe/handlers.py
```diff
@@ -1139,7 +1139,7 @@ def post(self, user_id=None, name=None):
 
         AuditLog.log(self.session, self.current_user.id, 'add_public_key',
                      'Added public key: {}'.format(pubkey.fingerprint),
-                     on_user_id=self.current_user.id)
+                     on_user_id=user.id)
```
tmildorf: Rebase?

tmildorf:
Yup, this is pretty much what I had in mind.

bin/grouper-api
zorkian: Yeah, I'm not sure how that would avoid the race. They won't error -- it's perfectly valid to update a column to the same value twice?

Anyway, I was being lazy, I can update this. There's a method of doing this I think that will work better, it just requires a bunch of back and forths with the db.

bin/grouper-api
zorkian: Will do.

bin/grouper-api
zorkian: I'm not sure what you mean... have a function like `send_email(now)` and have the main loop call it with `datetime.utcnow()`?

grouper/fe/handlers.py
zorkian: Yeah, will do.

bin/grouper-api
tmildorf: Yes.  That way we can have tests that simulate time.

zorkian:
Updated for feedback.

zorkian:
Ping.

bin/grouper-api
```diff
@@ -29,6 +31,66 @@ class Application(tornado.web.Application):
...
+            int: Number of emails that were sent.
+        """
+        emails = session.query(AsyncNotification).filter(
+            AsyncNotification.sent == False,
+            AsyncNotification.send_after < now_ts,
+        ).all()
+        for email in emails:
+            # For atomicity, attempt to set the sent flag on this email to true if
+            # and only if it's still false.
+            # try:
```
tmildorf: nit: leftover?

bin/grouper-api
```diff
@@ -29,6 +31,66 @@ class Application(tornado.web.Application):
...
+                AsyncNotification.sent == False
+            ).update({"sent": True})
+
+            # If it's 0, someone else won the race. Bail.
+            if update_ct == 0:
+                continue
+
+            try:
+                send_email_raw(settings, email.email, email.body)
+                email.sent = True
```
tmildorf: Does this line or line 65 actually change the persisted state?  Should email.sent = False be set in the exception handler?

zorkian: Nothing should change until `commit` is called, AIUI.

bin/grouper-api
```diff
@@ -29,6 +31,66 @@ class Application(tornado.web.Application):
...
 
 
+class SendEmailThread(Thread):
+    """Background thread for sending emails from the async pool
+
+    This class thread will exist on multiple servers in a standard Grouper production environment
+    so we need to ensure that it's race-safe.
+    """
+
+    def send_emails(self, session, now_ts):
```
tmildorf: This goes part of the way to mocking up the state for testing.  I think we want a dry_run parameter as well so that we can test the notification system.

zorkian: Would `dry_run` update the db? send an email? neither?

tmildorf: Do everything apart from actually sending email:  write to db, return number of would-be sent emails, even build the email to send but not send it.

zorkian:
Added dry_run and removed nit.

bin/grouper-api
```diff
@@ -29,6 +31,70 @@ class Application(tornado.web.Application):
...
 
 
+class SendEmailThread(Thread):
+    """Background thread for sending emails from the async pool
+
+    This class thread will exist on multiple servers in a standard Grouper production environment
+    so we need to ensure that it's race-safe.
+    """
+
+    def send_emails(self, session, now_ts, dry_run=False):
```
tmildorf: It makes sense to run the email-sending thread in the grouper api server(s), but the DB-updating send_emails method should be importable, probably in grouper.util?

zorkian:
Updated to move thread to separate file for easier use.

PR #35, opened 2015-05-23
Implement typeahead for users.

mattspitz:
This implements typeahead using the [Chosen](http://harvesthq.github.io/chosen/) library.

It adds all users as a `<select>` field, since those are the valid options, anyway. This leaks the list of all users to someone who's using Grouper, but if we care about what users an owner is allowed to know about, we can easily fix that by filtering the list of options. For now, I think the usability win is worthwhile.

The other TODO is to use CSS for Chosen that is more compatible with Bootstrap. These exist** but not on cdnjs.com, so I hesitated before adding them as static resources to stick with what's done elsewhere. Please advise if we should do something different.

Here's [a video](https://www.dropbox.com/s/wayseg9t7u9qjiv/GrouperAutocompleteDemo.mov?dl=0) of the whole flow in action.

** e.g. https://github.com/alxlit/bootstrap-chosen, https://github.com/dbtek/chosen-bootstrap


tmildorf:
To my knowledge, we're not currently concerned with filtering users, and anyone can see a full list at /users.  As you point out, the different style is mildly suboptimal.  Is there a significant downside to distributing the appropriate CSS as static content?

mattspitz:
I don't feel strongly about copying the CSS and distributing it with Grouper, but this would be the first library not distributed via CDNJS and would sign us up for keeping that CSS up-to-date.

@zorkian, any comments?

zorkian:
Bundling CSS is a time honored tradition in dependency management. It's kind of one of the only ways of doing it in the pre-cdnjs days. So while we have cdnjs and should use it if we can, if we don't have it then I don't think that's necessary an indication that we can't use a particular library.

My suggestion would be to bundle like `grouper/fe/static/ext/LIBRARY-VERSION.VERSION/...` and make those directories [WORM](http://en.wikipedia.org/wiki/Write_once_read_many).

mattspitz:
https://www.dropbox.com/s/yt3je5ytav9kqzy/Screenshot%202015-05-26%2014.49.35.png?dl=0

Looks nicer now!

mattspitz:
My mistake. I previously didn't allow selecting groups!

At some point, we should perhaps differentiate between users and group (maybe with a nice little icon while we render the selection widget), but in practice right now, users have an @ sign.

PR #37, opened 2015-05-29
Support removing users from groups.

mattspitz:
Added a 'remove' button next to the 'edit' button that removes a user
from a group via modal.

Users must be owners to remove members from groups. Users may not remove
themselves (use "leave group" instead).

### DIFF
--- a/grouper/fe/handlers.py
+++ b/grouper/fe/handlers.py
@@ -14,8 +14,9 @@ from ..audit import assert_controllers_are_auditors, assert_can_join, UserNotAud
 from ..constants import PERMISSION_GRANT, PERMISSION_CREATE, PERMISSION_AUDITOR
 
 from .forms import (
-    GroupForm, GroupJoinForm, GroupAddForm, GroupRequestModifyForm, PublicKeyForm,
-    PermissionCreateForm, PermissionGrantForm, GroupEditMemberForm,
+    GroupForm, GroupJoinForm, GroupAddForm, GroupRemoveForm,
+    GroupRequestModifyForm, PublicKeyForm, PermissionCreateForm,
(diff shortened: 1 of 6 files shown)


mattspitz:
https://www.dropbox.com/s/6e0gt3qvwiae6nw/GrouperRemoveFromGroupDemo.mov?dl=0

grouper/fe/handlers.py
```diff
@@ -841,6 +842,40 @@ def post(self, group_id=None, name=None):
...
 
 
+class GroupRemove(GrouperHandler):
+    def post(self, group_id=None, name=None):
+        group = Group.get(self.session, group_id, name)
+        if not group:
+            return self.notfound()
+
+        members = group.my_members()
+        if self.current_user.my_role(members) != "owner":
```
tmildorf: Managers see remove buttons for members, but when they attempt to remove a member, their efforts fail here.

mattspitz: Good catch. My mistake. Fixed!

grouper/fe/handlers.py
```diff
@@ -869,6 +870,40 @@ def post(self, group_id=None, name=None):
...
+
+        removed_member = get_user_or_group(self.session, member_name, user_or_group=member_type)
+
+        if self.current_user == removed_member:
+            return self.send_error(
+                status_code=400,
+                reason="Can't remove yourself. Leave group instead."
+            )
+
+        group.revoke_member(self.current_user, removed_member, "Removed by owner.")
```
tmildorf: Might be a manager rather than owner.

mattspitz: Sorted!

PR #38, opened 2015-06-10
Add /audited-permissions, /audited-groups.

tmildorf:
This adds buttons to the /groups and /permissions pages that toggle between complete lists and audited-only lists; the audited-only lists are served at /audited-groups and /audited-permissions respectively.  This should help with scoping an audit.

### DIFF
--- a/grouper/fe/handlers.py
+++ b/grouper/fe/handlers.py
@@ -6,7 +6,7 @@ import operator
 from expvar.stats import stats
 from tornado.web import RequestHandler
 
-from sqlalchemy import union_all
+from sqlalchemy import union_all, or_, and_
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql import label, literal
 
@@ -21,7 +21,7 @@ from .forms import (
     PermissionGrantForm, GroupEditMemberForm,
 )
 from ..models import (
-    User, Group, Request, PublicKey, Permission, PermissionMap, AuditLog, GroupEdge,
+    User, Group, Request, PublicKey, Permission, PermissionMap, AuditLog, GroupEdge, Counter,
     GROUP_JOIN_CHOICES, REQUEST_STATUS_CHOICES, GROUP_EDGE_ROLES, OBJ_TYPES,
     get_all_groups, get_all_users, get_user_or_group,
 )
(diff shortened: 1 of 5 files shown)


grouper/fe/handlers.py
```diff
@@ -732,6 +743,16 @@ def get(self):
             .filter(Group.enabled == True)
             .order_by(Group.groupname)
         )
+        if audited_only:
+            audited_permissions = (
+                self.session.query(Permission)
+                .filter(Permission._audited == True).all()
+            )
+            groups = groups.filter(
```
zorkian: This won't catch inheritance, will it? I.e. permission on group `x` makes it audited, but `y` is a member of `x`, so `y` should also be shown as audited.

grouper/fe/templates/groups.html
```diff
@@ -2,11 +2,18 @@
...
+      Audited
+    {% endif %}
     Groups
 {% endblock %}
 
 {% block subheading %}
-    {{total}} group(s)
+    {{total}}
+    {% if audited_groups %}
+      audited
```
zorkian: nit: Capitalize

grouper/fe/handlers.py
tmildorf: Good catch!  We need to expand groups to all group descendants.

zorkian:
BTW you have to comment on a PR after you push an update, else there's no notification that it is ready to re-review.

grouper/fe/handlers.py
zorkian: I see you updated this, but why are you doing the queries manually instead of using the graph which contains all of this data calculated? Granted, the graph means you're 10s delayed or whatever the reload interval is, but that's probably better than doing this query manually?

tmildorf: Touche.  That would be much more efficient, higher availability, etc. etc.

tmildorf:
In principle, neither the GroupsView nor the PermissionsView should be doing SQL queries at all.  I'll switch them up.

tmildorf:
Updated: no-longer query SQL when viewing groups/permissions.

grouper/graph.py
```diff
@@ -26,6 +26,14 @@
 def Graph():  # noqa
     return GroupGraph()
 
+# A GroupGraph caches permissions and groups as these objects which are intended
+# to behave like the corresponding models but without any connection to SQL
+# backend.
+PermissionTuple = namedtuple("CachedTuple",
```
zorkian: Isn't the first argument to `namedtuple` supposed to be the name? I.e. `PermissionTuple` here?

tmildorf: Yup.  Oversight on my part.  I had originally named them CachedTuples then quickly realized that that was far too vague and renamed them, but missed this.

PR #40, opened 2015-06-15
Fix bug affecting startup.

tmildorf:
The Grouper UI sometimes 500s when because it goes looking for self._rgraph to serve a user's personal view before the graph has been initialized.  This is an easy way to fix that.

### DIFF
--- a/grouper/fe/handlers.py
+++ b/grouper/fe/handlers.py
@@ -1,3 +1,5 @@
+# TODO(mildorf): we should refresh the graph after every write operation so that
+# redirects at the end of posts will include their effects.
 from datetime import datetime
 import operator
 
--- a/grouper/fe/util.py
+++ b/grouper/fe/util.py
@@ -68,6 +68,10 @@ class GrouperHandler(tornado.web.RequestHandler):
             if created:
                 logging.info("Created new user %s", username)
                 self.session.commit()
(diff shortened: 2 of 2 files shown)


mattspitz:
Are there other places where we might try to access the graph before it's initialized? Might be worth wrapping this.

tmildorf:
Good question.  I'm not sure.  I haven't figured out exactly why this sometimes happens.

tmildorf:
Ok.  This affects the first time the first user visits Grouper.  They are added to the db in a roundabout way and an updates table is populated, but the graph is not refreshed at that time, and it so happens that the graph is not initialized at all if the updates table is empty.  The two combine for the buggy behavior.

A better solution would be to resync the graph structure in this situation (the auto-adding for users appears to be unique).

Separately, we should resync the graph structure after each post method that ends in a redirect, since users will expect to see the effects of their writes.

mattspitz:
Great! It'd be nice to have a comment to that effect (and perhaps a TODO with your suggestions above). Feel free to merge it after you add the comment.

And I agree with refreshing the graph on commit(), particularly for redirects. Is refreshing the graph threadsafe? Should we just do it after every commit()? Seems like that'd be easier than trying to whack-a-mole every other case. The write volume should be pretty low, and you could write a nice little wrapper (or override session.commit()) that commits and refreshes the database.

PR #43, opened 2015-06-18
Allow for non-technical owners, fix manager add choices.

tmildorf:
Create new role "No-Permissions Owner" (a.k.a. "np-owner") that has full group ownership but does not provide permission inheritance via the group.  A user serving as np-owner in a group can still obtain permissions if they have membership in other groups.

Bugfixes: allow managers to remove other managers, don't allow managers to add owners.

### DIFF
--- a/grouper/fe/handlers.py
+++ b/grouper/fe/handlers.py
@@ -273,7 +273,7 @@ class PermissionsGrant(GrouperHandler):
         # If the permission is audited, then see if the subtree meets auditing requirements.
         if permission.audited:
             fail_message = ("Permission is audited and this group (or a subgroup) contains " +
-                            "owners or managers who have not received audit training.")
+                            "owners, np-owners, or managers who have not received audit training.")
             try:
                 permission_ok = assert_controllers_are_auditors(group)
             except UserNotAuditor as e:
@@ -507,7 +507,7 @@ class GroupEditMember(GrouperHandler):
 
         members = group.my_members()
         my_role = self.current_user.my_role(members)
-        if my_role not in ("manager", "owner"):
+        if my_role not in ("manager", "owner", "np-owner"):
             return self.forbidden()
 
         member = members.get((member_type.capitalize(), name2), None)
@@ -525,9 +525,10 @@ class GroupEditMember(GrouperHandler):
 
         form = GroupEditMemberForm(self.request.arguments)
         form.role.choices = [["member", "Member"]]
-        if my_role == "owner":
+        if my_role in ("owner", "np-owner"):
             form.role.choices.append(["manager", "Manager"])
             form.role.choices.append(["owner", "Owner"])
+            form.role.choices.append(["np-owner", "No-Permissions Owner"])
 
         form.role.data = edge.role
         form.expiration.data = edge.expiration.strftime("%m/%d/%Y") if edge.expiration else None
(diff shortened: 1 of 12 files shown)


zorkian:
`Bugfixes: allow managers to remove other managers`

That's not a bug. That's a design choice, because managers are not supposed to be able to control other managers. Did you really mean to change this?

grouper/fe/handlers.py
```diff
@@ -899,7 +912,7 @@ def post(self, group_id=None, name=None):
                 reason="Can't remove yourself. Leave group instead."
             )
 
-        group.revoke_member(self.current_user, removed_member, "Removed by owner/manager")
+        group.revoke_member(self.current_user, removed_member, "Removed by owner/np-owner/manager")
         AuditLog.log(self.session, self.current_user.id, 'remove_from_group',
                      '{} was removed from the group.'.format(removed_member.name),
                      on_group_id=group.id)
```
zorkian: This should really have an `on_user_id` for the person being removed. I know you didn't edit this code, but I just saw it and don't recall reviewing it, so.

grouper/fe/templates/macros/ui.html
```diff
@@ -152,9 +153,9 @@ <h3 class="panel-title">Groups</h3>
     <tr>
         <td>
             {% if member.name != current_user.name and
-                (current_user.my_role(members) == "owner" or
+                (current_user.my_role(members) in ("owner", "np-owner") or
                  (current_user.my_role(members) == "manager" and
-                  ROLES[member.role] == "member"))
+                  ROLES[member.role] in ("member", "manager")))
```
zorkian: As mentioned, the idea of managers was that they can only control members. The management structure is controlled by the owners. (This can obviously change, but I don't want it changed because someone thought it was a bug -- it's a deliberate decision.)

grouper/graph.py
```diff
@@ -362,13 +362,37 @@ def get_group_details(self, groupname, cutoff=None, show_permission=None):
...
+            # User permissions are inherited from all groups for which their
+            # role is not "np-owner".  User groups are all groups in which a
+            # user is a member by inheritance, except for ancestors of groups
+            # where their role is "np-owner", unless the user is a member of
+            # such an ancestor via a non-"np-owner" role in another group.
+            rpaths = {}
+            for group in self._rgraph.neighbors(user):
+                role = self._rgraph[user][group]["role"]
+                if GROUP_EDGE_ROLES[role] == "np-owner":
+                    print "skipping group", group
```
zorkian: Debugging

tests/fixtures.py
```diff
@@ -78,7 +80,7 @@ def graph(session):
 def users(session):
     users = {
         username: User.get_or_create(session, username=username)[0]
-        for username in ("gary", "zay", "zorkian", "oliver", "testuser")
+        for username in ("gary", "zay", "zorkian", "oliver", "testuser", "figurehead")
```
zorkian: Hahah, I love the name of the user.

zorkian:
Finished review. Overall, LGTM. My one comment was around the change to how managers work. Let me know what you want to do there (if anything) and then this can be merged.

grouper/fe/handlers.py
tmildorf: Will do.

grouper/fe/templates/macros/ui.html
tmildorf: Ok.

grouper/graph.py
tmildorf: Good catch!

tests/fixtures.py
tmildorf: I considered that for the name of the role, but decided that its connotations were a bit too nebulous.

tmildorf:
I'm happy as long as every change is immediately-reversible by its actioner.  This diff still includes a fix for manager add choices (they can add members only, rather than any role).

PR #44, opened 2015-06-22
Add group creator role choice; refresh before redirecting

tmildorf:
Allow group creators to choose either regular Owner or No-Permissions Owner for their role in the new group.  Refresh the graph just before redirects so the results of POSTs are immediately visible.


zorkian:
Seems reasonable, good catch!

PR #45, opened 2015-06-23
Populate the graph with both enabled and disabled groups and users.

tmildorf:
Include disabled users and groups in the graph.  Disabled groups have no members and don't belong to other groups, but still display their description and direct permissions; -disable/+enable are inverses for groups.  In this implementation, re-enabled users start with no group memberships, so -disable/+enable are not inverses for users.


zorkian:
I'm not sure this does what you want. Since we don't remove membership edges in the database, if you stop filtering on enabled here, won't this grant disabled users their permissions?

PR #46, opened 2015-06-24
Keep disabled users and groups out of the in-memory graph.

tmildorf:
This is a safer solution to the 500s we were seeing in the UserView handler.  It's preferable to keep disabled users and groups out of the in-memory graph in order to simplify UX, correctness of clients, etc.

### DIFF
--- a/grouper/fe/handlers.py
+++ b/grouper/fe/handlers.py
@@ -20,6 +20,7 @@ from .forms import (
     GroupRequestModifyForm, PublicKeyForm, PermissionCreateForm,
     PermissionGrantForm, GroupEditMemberForm,
 )
+from ..graph import NoSuchUser, NoSuchGroup
 from ..models import (
     User, Group, Request, PublicKey, Permission, PermissionMap, AuditLog, GroupEdge, Counter,
     GROUP_JOIN_CHOICES, REQUEST_STATUS_CHOICES, GROUP_EDGE_ROLES, OBJ_TYPES,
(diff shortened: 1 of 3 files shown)


grouper/fe/handlers.py
```diff
@@ -96,8 +98,9 @@ def get(self, user_id=None, name=None):
 
         try:
             user_md = self.graph.get_user_details(user.name)
-        except KeyError:
-            # User is probably very new, so they have no metadata yet.
+        except (KeyError, NetworkXError):
```
mattspitz: Seems like we should be more explicit with the error states we're allowing.

In particular,
- Can we be more explicit than a `KeyError`? This is unrelated to this change, but we should have a specific exception if we're going to catch it, as `KeyError`s aren't specific to getting user details and might catch more exceptions than we mean to.
- Can we be more specific than a `NetworkXError`? Is there a more specific exception (or set of exceptions) that we should allow? The answer may be no, but I figured I'd ask. The comment seems very specific to disabled users but the exception handling is general.

mattspitz:
This feels like it's brittle to me. I don't understand all the different places where we query for disabled or not_disabled users or where the best place to make this change is, but I don't feel confident enough in my knowledge of the code to review it effectively. @zorkian / @gmjosack, any opinions?

grouper/fe/handlers.py
tmildorf: I'm not 100% sure where the KeyErrors come from but I'd guess from the comment that it's graph.py line 507 in the get_user_details function.  Likewise, NetworkXError is coming from line 475.  Maybe we should just make that function do the right thing instead.

tmildorf:
This is actually reverting a recent change to _get_nodes_from_db, and it adds a filter to _get_group_tuples to make it behave like the other DB-syncing functions (i.e. ignoring disabled entities.)

PR #48, opened 2015-06-26
Include disabled users in API server /users, serve like frontend.

tmildorf:
Include disabled users in API server /users route.  Serve disabled users with no permission, no groups, just their public keys.  Add a field that distinguishes disabled users from enabled users with no permission and no groups.

### DIFF
--- a/grouper/api/handlers.py
+++ b/grouper/api/handlers.py
@@ -46,19 +46,22 @@ class Users(GraphHandler):
         with self.graph.lock:
             if not name:
                 return self.success({
-                    "users": [
-                        user
-                        for user in self.graph.users
-                    ],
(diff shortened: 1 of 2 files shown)


grouper/api/handlers.py
```diff
@@ -46,19 +46,22 @@ def get(self, name=None):
         with self.graph.lock:
             if not name:
                 return self.success({
-                    "users": [
-                        user
-                        for user in self.graph.users
-                    ],
+                    "users": sorted(list(self.graph.users)+list(self.graph.disabled_users.keys())),
```
MarkChuCarroll: Nit, but there should be spaces around the +

MarkChuCarroll:
Modulo one tiny nit, it looks OK to me.

PR #49, opened 2015-06-26
Increase the rows returned by frontend.  Raise the hard limit.

tmildorf:
Return 10x as many results from the frontend.  Allow users to enter much large values manually.


mattspitz:
Never trust a HAL 9000.

https://www.youtube.com/watch?v=wT6vlfpReuE

PR #53, opened 2015-07-02
API servers treat disabled users like enabled users.

tmildorf:
There's no reason to treat disabled users and enabled users differently on an API server.  Reuse code.

### DIFF
--- a/grouper/api/handlers.py
+++ b/grouper/api/handlers.py
@@ -46,22 +46,16 @@ class Users(GraphHandler):
         with self.graph.lock:
             if not name:
                 return self.success({
-                    "users": sorted(list(self.graph.users) + list(self.graph.disabled_users.keys())),
+                    "users": sorted(self.graph.user_metadata.keys()),
                 })
 
-            if name not in self.graph.users and name not in self.graph.disabled_users:
-                return self.notfound("User (%s) not found." % name)
-
-            out = {"user": {"name": name, "enabled": not name in self.graph.disabled_users}}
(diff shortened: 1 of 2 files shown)


grouper/api/handlers.py
```diff
@@ -46,22 +46,20 @@ def get(self, name=None):
         with self.graph.lock:
             if not name:
                 return self.success({
-                    "users": sorted(list(self.graph.users) + list(self.graph.disabled_users.keys())),
+                    "users": sorted(list(self.graph.users) + \
```
zorkian: Is the trailing `\` necessary? I thought as long as you're within a parenthesis, you can just drop it.

grouper/graph.py
```diff
@@ -130,29 +130,12 @@ def _get_checkpoint(session):
...
-            out[user.username] = [{
-                "public_key": key.public_key,
-                "fingerprint": key.fingerprint,
-                "created_on": str(key.created_on),
-            } for key in user.my_public_keys()]
-        return out
-
-    @staticmethod
-    def _get_user_metadata(session):
+    def _get_user_metadata(session, enabled=True):
```
zorkian: Why don't we get all metadata? The query is more efficient to run once, and since there's no index on the column, it's a full table scan either way.

tmildorf:
Ping.

zorkian:
LGTM. This request can't be merged, so if you rebase it, you can go ahead and merge it.

PR #54, opened 2015-07-03
Add refresh requests to certain redirects.

tmildorf:
Use an argument when serving redirect to ask the possibly-different process/host serving the follow-up get to sync its graph before serving the page.


zorkian:
Good idea, I like it.

PR #56, opened 2015-07-04
Add asynchronous config-refreshing.

tmildorf:
Keep Grouper configuration fresh by reloading the config file every few seconds.

### DIFF
--- a/grouper/settings.py
+++ b/grouper/settings.py
@@ -1,9 +1,16 @@
+import logging
+import threading
+import time
 import yaml
 
+from expvar.stats import stats
 
 class Settings(object):
     def __init__(self, initial_settings):
         self.settings = initial_settings
+        self.lock = threading.Lock()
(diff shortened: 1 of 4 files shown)


grouper/settings.py
```diff
@@ -32,16 +39,31 @@ def update_from_config(self, filename, section=None):
...
 
+            self[key] = value
+
+
+    def update_from_config(self, filename, section=None, refresh_config_seconds=10):
+        self._update_from_config(filename, section=section)
+        def refresh_config_loop():
+            while True:
+                time.sleep(refresh_config_seconds)
+                self._update_from_config(filename, section=section)
```
mattspitz: If this fails once, the thread dies. Should we catch and log an exception so it can try again later?

tmildorf: Yup!  Really we should expose the stats via the /debug/stats route like the other asynchronous methods.

PR #60, opened 2015-07-07
Avoid race that 2x's work on refresh.

tmildorf:
The recently-developed refresh argument causes an update race that ends up building the new graph twice.  Avoid that race by adding a mutex specifically for the update method.

### DIFF
--- a/grouper/graph.py
+++ b/grouper/graph.py
@@ -45,7 +45,8 @@ class GroupGraph(object):
         logging.info('Created graph object.')
         self._graph = None
         self._rgraph = None
-        self.lock = RLock()
+        self.lock = RLock() # Graph structure.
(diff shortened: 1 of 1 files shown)


zorkian:
What happens if it can't get the lock? Does it pause and wait for the lock to become available? (At which point we just do two serial updates..)

tmildorf:
The race is caused by POST+GETs.  Generally, the follow-up GET starts a refresh, which currently takes 2-3 seconds (rather slow, a separate issue that should be improved), then the async refresh comes in and piles on extra SQL work as well (or vice versa).  What should happen now is one update that does actual work; the other blocks on the update lock, then just tests the checkpoint and finds out that it's already up to date in a few milliseconds.

PR #62, opened 2015-07-20
Make the settings thread into a daemon thread.

tmildorf:
Because the settings-loading thread is not a daemon thread, grouper-ctl stalls when it should just exit.


MarkChuCarroll:
LGTM.

PR #64, opened 2015-07-21
Identify role users; API servers only serve them if asked

tmildorf:
Role users can be pretty confusing in certain contexts. Make them opt-in for API server clients, via GET argument include_role_users=yes, and identify them on the front end.


MarkChuCarroll:
Looks good to me.

PR #66, opened 2015-07-28
Add a plugin method for site-specific error handling

tmildorf:
Add an exception handler to the base plugin.  Make the API servers return code 500 rather than no response when an internal error comes up in a handler.


MarkChuCarroll:
LGTM.

PR #68, opened 2015-08-06
add page that aggregates all a user's requests across groups

herb:


tmildorf:
Oh, exciting!  There has been interest in this for some time.

grouper/fe/templates/user-requests.html
```diff
@@ -0,0 +1,66 @@
...
+                                <strong>Role:</strong> {{ROLES[request.role]}}<br />
+                                <strong>Reason:</strong> <em>{{request.reason}}</em>
+                            </div>
+                        </td>
+                    </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+        {% else %}
+        <div class="center">nothing pending</div>
```
tmildorf: nit: Nothing pending.

grouper/fe/templates/user-requests.html
```diff
@@ -0,0 +1,66 @@
...
+                    <th class="col-sm-2">Status</th>
+                    <th class="col-sm-2">Requested At</th>
+                    <th class="col-sm-2">Expiration</th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for request in requests %}
+                    <tr>
+                        <td>
+                            <a href="/groups/{{request.groupname}}/requests/{{request.id}}"
```
tmildorf: This is certainly convenient, but it currently has the unfortunate side-effect that you don't end up back on the aggregated requests page after actioning a request.  Could you add a from argument so that the GroupRequestUpdate handler can redirect back to the appropriate page?

tmildorf:
I didn't spot a link to /users/requests anywhere.  Could you add something like the requests button from the group page to a user's home page and direct it to that new path?

grouper/fe/templates/user-requests.html
herb: fixed

grouper/fe/templates/user-requests.html
herb: done

herb:
added the link from user page and added some basic tests for requests.

PR #71, opened 2015-08-12
Remove duplicate README

lfaraone:


gmjosack:
You likely would want to remove the reference to this README in the setup.py but as the comment there says:

    # PyPi, despite not parsing markdown, will prefer the README.md to the
    # standard README. Explicitly read it here.

Only having a markdown config will result in seeing literal markdown on PyPI unless they changed that.

You also likely wouldn't need the Makefile anymore since that's just used to create the rst version from the markdown version.

gmjosack:
MANIFEST.in also refers to this file. You might want to try a `python setup.py sdist` to make sure you haven't broken the building of a package.

lfaraone:
If we delete `README.md` and rename `README` to `README.rst` (and clean up the relevant references), we'd only have one copy of the file and get pretty rendering on both PyPI and GitHub. [See example branch for how GitHub shows `README.rst`](https://github.com/lfaraone/grouper/tree/test-with-no-readme). 

Thoughts?

gmjosack:
Yeah, I'm not against that. This mostly existed just because I prefer to write markdown to restructured text but that's not a great reason for the extra complication.

lfaraone:
Abandoning, see #72.

PR #72, opened 2015-08-14
Drop README in favour of README.md

lfaraone:


setup.py
```diff
@@ -29,9 +29,8 @@ def get_package_data(package, base_dir):
     "package_data": package_data,
     "scripts": ["bin/grouper-api", "bin/grouper-fe", "bin/grouper-ctl"],
     "description": "Self-service Nested Group Management Server.",
-    # PyPi, despite not parsing markdown, will prefer the README.md to the
-    # standard README. Explicitly read it here.
-    "long_description": open("README").read(),
+    # TODO(lfaraone): Check whether this is still needed for PyPI.
```
lfaraone: If we don't specify a `long_description`, will PyPI DTRT and use the `README.rst`?

gmjosack: I believe that it will, though I'm not completely positive.

gmjosack:
Can you kill the Makefile now too?

lfaraone:
Sure, will do.

PR #75, opened 2015-08-24
do not strip name input prior to validation. this ensures leading/tra…

herb:
…iling whitespace does not make it into name values. this extra whitespace gets escaped in subsequent links which routes doesn't know what to do with.

### DIFF
--- a/grouper/fe/forms.py
+++ b/grouper/fe/forms.py
@@ -17,7 +17,7 @@ class ValidateRegex(object):
         self._re = re.compile(self.regex)
 
     def __call__(self, form, field):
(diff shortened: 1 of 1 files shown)


lfaraone:
Can you amend the commit message so the first line isn't wrapped?

This can be fixed by adding two newlines after the first period.

herb:
can i do that without submitting a new pull request? alternatively can we just do that in the merge?

lfaraone:
You can amend and force-push the branch; github will pick it up
automatically.

herb:
okie dokie. amended.

PR #76, opened 2015-08-25
add an endpoint to audit all user ssh_keys

herb:
the endpoint is behind Yet Another Built In Permission(tm).

### DIFF
--- a/grouper/constants.py
+++ b/grouper/constants.py
@@ -16,6 +16,7 @@ ARGUMENT_VALIDATION = r"(?P<argument>|\*|[\w=+/.:-]+\*?)"
 PERMISSION_GRANT = "grouper.permission.grant"
 PERMISSION_CREATE = "grouper.permission.create"
 PERMISSION_AUDITOR = "grouper.permission.auditor"
+AUDIT_SECURITY = "grouper.audit.security"
 AUDIT_MANAGER = "grouper.audit.manage"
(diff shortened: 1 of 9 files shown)


grouper/constants.py
```diff
@@ -24,6 +25,7 @@
     (PERMISSION_CREATE, "Ability to create permissions within Grouper."),
     (PERMISSION_GRANT, "Ability to grant a permission to a group."),
     (PERMISSION_AUDITOR, "Ability to own or manage groups with audited permissions."),
+    (AUDIT_SECURITY, "Ability to audit security related activity on the system."),
```
lfaraone: Should be `security-related`. And is this going to be just for SSH keys, or will the scope be expanded later on?

grouper/fe/routes.py
```diff
@@ -25,6 +25,7 @@
     ),
     (r"/search", handlers.Search),
     (r"/users", handlers.UsersView),
+    (r"/users/public-keys", handlers.UsersPublicKey),
```
lfaraone: These are specifically SSH keys, right? Will this endpoint return, e.g., X.509 public keys if Grouper grows support for that in the future?

herb: It's specific to ssh keys but so is the rest of the endpoints that have 'public-key' in the url path. I'd expect any future upgrade to support other types of public keys should include expanding this endpoint to support it.

PR #86, opened 2015-09-02
sort member to be audited

herb:


bismuth-taco:
i approve this message

PR #88, opened 2015-09-04
log rates for basic metrics

herb:


rra:
Would it make sense to add some simple tests for this, just to be sure the new stats show up in the stats view?  Or is there not a test suite for that stuff yet?

herb:
it would make lots of sense. but, yeah, no test suite for handlers yet.

PR #90, opened 2015-09-11
only allow group owner/np-owner to complete an audit for that group

herb:


grouper/fe/handlers.py
```diff
@@ -821,6 +821,12 @@ def post(self, audit_id):
             return self.forbidden()
 
         audit = self.session.query(Audit).filter(Audit.id == audit_id).one()
+
+        # only owners can complete
+        owner_ids = {member.id for n,member in audit.group.my_owners().items()}
```
rra: This seems a bit odd to me, since I'm immediately wondering what the values of that dict are (and `n` is unused).  How about:

```
owner_ids = {member.id: True for _, member in audit.group.my_owners().items()}
```

rra: Or, actually, since you only care about the values:

```
owner_ids = {member.id: True for member in audit.group.my_owners().values()}
```

herb: i was intending to help document my_owners() since it doesn't do what i expect but i guess this was the totally wrong approach to that. :)

this is also a set vs a dict.

i updated to use `values()` instead so there's no confusion about `n`.

rra:
Oh, ack, right, I completely forgot about that set construction notation.  Sorry about that!

PR #101, opened 2015-09-29
Send email when group join requests are processed

rra:
When a group join request has been either actioned or cancelled,
send email to the affected person letting them know and including
the reason field.  Assumes that the user name is an email address,
but that appears to be a common assumption in Grouper already.

### DIFF
--- a/grouper/fe/handlers.py
+++ b/grouper/fe/handlers.py
@@ -1224,6 +1224,20 @@ class GroupAdd(GrouperHandler):
                          member.name, form.data["role"]),
                      on_group_id=group.id)
 
+        if member.type == "User":
+            self.send_email(
+                [member.name],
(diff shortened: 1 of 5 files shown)


rra:
I'm not at all sure how to test this effectively.  Is there a simple way to stand up a local instance of Grouper?

herb:
Yes, a local instance is possible. If you follow the instructions in README and use virtualenv to pull down the dependencies it should Just Work(tm).

rra:
Oh, doh, there are instructions right there.  Thanks!

rra:
I've now tested this via a local instance with a modified dev configuration to enable sending email notifications.  Tested both the actioned and canceled states, with and without an expiration set, and the email looked good in all cases.

PR #103, opened 2015-09-30
Support a bind address for the FE and API servers

rra:
Add a command-line argument and configuration option to set the
bind address of the API and front-end servers.  Change the dev
configuration to bind both to localhost by default, rather than
listening to the public interfaces of development machines.

### DIFF
--- a/bin/grouper-fe
+++ b/bin/grouper-fe
@@ -66,6 +66,8 @@ def main(argv):
     parser.add_argument("-V", "--version", action="version",
                         version="%%(prog)s %s" % grouper.__version__,
                         help="Display version information.")
+    parser.add_argument("-a", "--address", type=str, default="",
+                        help="Override address in config.")
     parser.add_argument("-p", "--port", type=int, default=None,
                         help="Override port in config.")
     parser.add_argument("-n", "--deployment-name", type=str, default="",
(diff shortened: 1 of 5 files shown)


bin/grouper-fe
```diff
@@ -66,6 +66,8 @@ def main(argv):
     parser.add_argument("-V", "--version", action="version",
                         version="%%(prog)s %s" % grouper.__version__,
                         help="Display version information.")
+    parser.add_argument("-a", "--address", type=str, default="",
```
herb: nit: `default=None`?

rra: The method to my madness here is that the default in the config file is `""`, so I made this match.  But `None` would be better.  Is the `null` value supported in the YAML that we use, so that I can document that people can set `address: null`?  Or am I just making this all too complicated?

rra: Oh, I see, you just leave the key blank.  New diff coming.

herb: i believe it does support `null` but yeah, the convention seems to be leaving it blank. and i'm the one who's nit picking so if anything i'm making this all too complicated. :)

rra:
New diff added, and it seems a lot cleaner.  Thanks!

PR #107, opened 2015-10-01
consolidate public addition logic and add simple end-to-end tests

herb:
apparently `grouper-ctl` uses a different implementation of the public key addition logic. :cry:


herb:
mind taking a look at this when you get a chance? no rush.

PR #110, opened 2015-10-07
add some simple user/group validation to grouper-ctl

herb:


bin/grouper-ctl
```diff
@@ -32,6 +35,30 @@ def make_session():
     return models.Session()
 
 
+def ensure_valid_username(f):
+    @wraps(f)
+    def wrapper(args):
+        if not re.match("^{}$".format(USERNAME_VALIDATION), args.username):
+            print "Invalid username {}".format(args.username)
```
rra: One nit for future work (since I see that grouper-ctl currently does this all over the place): error messages from command-line utilities should really go to standard error instead of standard out.

PR #118, opened 2015-11-07
correctly validate more usernames

herb:


rra:
I think Grouper the open source project probably wants this change, but at Dropbox, we prohibited periods in usernames and wanted to catch those as early as possible, which makes me think that internally we could still want this restriction?  That's part of the general problem of being sure people use the same username everywhere.

herb:
indeed. i'll work on generalizing this a bit and allowing it to be set from
config. follow up separately about what prompted this change.

On Fri, Nov 6, 2015 at 5:58 PM, Russ Allbery <notifications@github.com>
wrote:

> I think Grouper the open source project probably wants this change, but at
> Dropbox, we prohibited periods in usernames and wanted to catch those as
> early as possible, which makes me think that internally we could still want
> this restriction? That's part of the general problem of being sure people
> use the same username everywhere.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dropbox/grouper/pull/118#issuecomment-154597455>.
>



-- 

herb

rra:
Replaced by https://github.com/dropbox/merou/pull/596.

PR #119, opened 2015-11-13
Consolidate email logic

herb:


rra:
I mostly just scanned this and am relying on the test suite, but it looks good to me!

PR #122, opened 2015-11-18
add flake8 to travis ci and fix all offending issues

herb:
made some exclusions based on personal preference + limiting scope of changes. feedback welcome.


setup.py
```diff
@@ -1,19 +1,21 @@
 #!/usr/bin/env python
 
 import os
+import setuptools  # noqa
 
-import setuptools
 from distutils.core import setup
 
-execfile('grouper/version.py')
+from grouper.version import __version__
```
gmjosack: This explicitly uses execfile so that grouper doesn't have to be importable and on the path in order to build grouper.

herb: ah, ok. well, i'll revert the behavior. maybe need to exclude `setup.py` from flake8 or just `# noqa` that line that references `__version__` then.

gmjosack: You might be able to use something else. i've seen things like `pkg_resources.get_distribution` be suggested. I've always used execfile out of simplicity. In general though it's not recommended to import your package in your setup.py.

herb: gotcha. tried and true is fine by me. reverted the change and added a comment.

PR #128, opened 2015-11-25
Replace capabilities with normal Grouper permissions

lfaraone:
This allows for us to audit those to whom these permissions are granted.

There is no migration included in this PR.


README.rst
```diff
@@ -76,14 +76,9 @@ via the following commands:
...
-    bin/grouper-ctl -vv \
-        capabilities add $USER@example.com user_admin
+    bin/grouper-ctl -vv\
+        user create $USER@example.com
 
-    # Allow someone to create permissions.
+    # Give the user administrative access to the Grouper instance
     bin/grouper-ctl -vv \
-        capabilities add $USER@example.com permission_admin
+        group add_member $USER@example.com
```
rra: Does this command need to specify what group to which to add the user?

grouper/constants.py
```diff
@@ -19,6 +19,9 @@
 PERMISSION_GRANT = "grouper.permission.grant"
 PERMISSION_CREATE = "grouper.permission.create"
 PERMISSION_AUDITOR = "grouper.permission.auditor"
+PERMISSION_ADMIN = "grouper.admin.permissions"
+USER_ADMIN = "grouper.admin.users"
+GROUP_ADMIN = "grouper.admin.groups"
```
lfaraone: I ported the semantics over directly from the old capabilities model. We can make things more fine-grained as we desire.

I put these under `grouper.admin.*`  to make it obvious that they're dangerous. But happy to swap the last two elements in the hierarchy for each.

README.rst
lfaraone: Fixed, thanks!

grouper/constants.py
rra: This looks good to me!

grouper/capabilities.py
```diff
@@ -1,11 +0,0 @@
-import bittle
```
herb: looks like this is the only usage of `bittle`, mind nuking it from All The Places?

grouper/constants.py
```diff
@@ -28,6 +31,9 @@
     (PERMISSION_CREATE, "Ability to create permissions within Grouper."),
     (PERMISSION_GRANT, "Ability to grant a permission to a group."),
     (PERMISSION_AUDITOR, "Ability to own or manage groups with audited permissions."),
+    (PERMISSION_ADMIN, "Ability to manipulate any permission."),
+    (USER_ADMIN, "Ability to to disable/enable any User account and modify its attributes."),
+    (GROUP_ADMIN, "Ability to approve/deny/revoke membership to any group."),
```
herb: Might be worth keeping these separate and preserving the existing semantic that `SYSTEM_PERMISSIONS` aren't automagically granted during `sync_db`.

It 'feels' cleaner to have grouper-administrators only get the admin permissions but maybe you can talk about why you merged them?

grouper/ctl/sync_db.py
```diff
@@ -5,6 +5,7 @@
 from grouper.models import (
         get_db_engine,
         Model,
+        Group,
```
herb: nit: alphabetical

grouper/ctl/sync_db.py
```diff
@@ -17,10 +18,36 @@ def sync_db_command(args):
...
+                canjoin="nobody",
+        )
+
+        try:
+            admin_group.add(session)
+            session.flush()
+        except IntegrityError:
+            session.rollback()
+            raise Exception('Failed to create group: grouper-administrators')
+    admin_group_permission_names = [perm[1] for perm in admin_group.my_permissions()]
```
herb: There's `Model.get_or_create()` which could replace this code but has the unfortunately side effect that if anyone changed the `description` or `canjoin` attributes would fail trying to create the group again.

Do we want to enforce either of these attributes as immutable? Seems silly but just asking the question. :)

grouper/ctl/sync_db.py
```diff
@@ -17,10 +18,36 @@ def sync_db_command(args):
...
+        if permission and name not in admin_group_permission_names:
+            try:
+                admin_group.grant_permission(permission)
+                session.commit()
+            except IntegrityError:
+                # The permission is already extant in the group, carry on.
+                session.rollback()
+            continue
+        elif permission:
             continue
```
herb: May be worth adding a quick test to verify `sync_db` creates the group and adds the correct permissions.

grouper/capabilities.py
lfaraone: Done in 65632c7.

grouper/ctl/sync_db.py
lfaraone: Fixed and squashed.

grouper/ctl/sync_db.py
lfaraone: I modified the code in 0fd2a16 to only take action on `grouper-administrators` if it doesn't already exist. This gives the site administrator the ability to disable, &c the group w/o `sync_db` overriding them.

grouper/constants.py
lfaraone: I've kept them in `SYSTEM_PERMISSIONS` but simply enumerate the permissions to automatically grant during `sync_db` in 0fd2a16.

grouper/ctl/sync_db.py
lfaraone: Done in c992df0.

lfaraone:
Ready for re-review.

PR #131, opened 2015-11-26
Send notifications and log to audit log for expired edges

rra:
Previously, expired edges were just excluded by SQL query but were
otherwise left untouched.  This meant that edge expiration was not
an actual event, so we didn't log anything to the audit log and didn't
send notification.

Add a background processing task that looks for all expired but active
edges and sends expiration notification, logs an audit log message,
and then deactivates the edge.

Before deployment, we'll need to run a one-time job to mark all existing
expired edges inactive.

### DIFF
--- a/grouper/audit.py
+++ b/grouper/audit.py
@@ -1,4 +1,4 @@
-from constants import PERMISSION_AUDITOR
+from .constants import PERMISSION_AUDITOR
 
 from grouper.graph import Graph
 from grouper.models import Audit
--- a/grouper/background.py
+++ b/grouper/background.py
@@ -3,10 +3,11 @@ import logging
 from threading import Thread
 from time import sleep
 
+from sqlalchemy import and_
 from sqlalchemy.exc import OperationalError
 
-from grouper.email_util import process_async_emails
-from grouper.models import get_db_engine, Session
+from grouper.email_util import notify_edge_expiration, process_async_emails
+from grouper.models import get_db_engine, GroupEdge, Session
 from grouper.perf_profile import prune_old_traces
 from grouper.util import get_database_url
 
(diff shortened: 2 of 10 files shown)


grouper/background.py
```diff
@@ -34,19 +35,20 @@ def capture_exception(self):
...
+                logging.debug("Expiring edges....")
+                expire_edges(session)
+                logging.debug("Pruning old traces....")
                 prune_old_traces(session)
                 session.commit()
                 session.close()
             except OperationalError:
                 Session.configure(bind=get_db_engine(get_database_url(self.settings)))
                 logging.critical("Failed to connect to database.")
-                self.capture_exception()
```
herb: this duplication is intentional so we can log OperationalError to sentry as well

rra: Yeah, this was a merge error on my part.  Now fixed.

grouper/background.py
```diff
@@ -34,19 +35,20 @@ def capture_exception(self):
...
         while True:
-            logging.debug("Sending emails...")
             try:
                 session = Session()
+                logging.debug("Sending emails...")
                 process_async_emails(self.settings, session, datetime.utcnow())
+                logging.debug("Expiring edges....")
+                expire_edges(session)
+                logging.debug("Pruning old traces....")
                 prune_old_traces(session)
```
herb: this is maybe a separate change but i'm wondering if we should be `stats.set_gauge` around each of the operations here (async emails, expiring, and pruning old perf traces). then we can alert on it similar to how we alert on db refresh and config refresh now.

grouper/models.py
```diff
@@ -1406,17 +1406,76 @@ def apply_changes(self, request):
                 if value:
                     expiration = datetime.strptime(value, "%m/%d/%Y")
                     setattr(self, key, expiration)
-                    AsyncNotification.add_expiration(self.session,
-                                                     expiration,
-                                                     group_name,
-                                                     member_name,
-                                                     recipients=recipients,
-                                                     member_is_user=member_is_user)
+                    # Avoid sending notifications for expired edges.
```
herb: nice

grouper/models.py
```diff
@@ -1406,17 +1406,76 @@ def apply_changes(self, request):
...
+        }
+        from grouper.fe.settings import settings as fe_settings
+        send_email(
+            session=self.session,
+            recipients=recipients,
+            subject="Membership in {} expired".format(group_name),
+            template="expiration",
+            settings=fe_settings,
+            context=email_context,
+        )
```
herb: i suspect we'll want to remove `send_email` and `async_send_email` dependencies within models. this will get us out of the circular dependency hole we're in now.

it's a little more annoying but sending the email would be cleaner in `audit.expire_edges` and will be one less dependency to remove when the time comes.

for the `async_send_email` dependency i was thinking of passing a callback into the model method. that could work here though it's simple enough you might be able to detach it entirely. your call.

grouper/background.py
rra: Let's make that a separate change if just because I think we want to restructure the exception handling so that we can log success and failure properly.  Right now, there isn't an exception block where we can clearly determine what failed.

rra:
Updated to split the audit and notification handling into `grouper.email_util` instead and avoid circular dependencies.

grouper/audit.py
```diff
@@ -1,4 +1,4 @@
-from constants import PERMISSION_AUDITOR
+from .constants import PERMISSION_AUDITOR
```
herb: can we make that `from grouper.constants import PERMISSION_AUDITOR` relative imports are confusing. :)

actually, i'll do it and merge it.

PR #133, opened 2015-12-09
log user removals referencing removed user correctly

herb:
also added end to end tests. which, for the record, i spent way too much time working on. :/


rra:
The tests are great!  Thank you.  I think that stuff is totally worth the time investment.

PR #139, opened 2015-12-18
fix cached graph to not include disabled groups

herb:


rra:
Wow, nice catch!

gmjosack:
:disappointed: Thanks for cleaning up my blunders!

PR #143, opened 2016-01-08
move group membership removal from user disable to user enable

herb:


grouper/fe/templates/user.html
```diff
@@ -108,20 +108,23 @@ <h3 class="panel-title">Open Audits</h3>
...
-                <button type="button" class="btn btn-default"
-                        data-dismiss="modal">Close</button>
-                <form action="/users/{{user.id}}/enable" method="post"
-                      style="display: inline;">
+            <form action="/users/{{user.id}}/enable" method="post">
+                <div class="modal-body">
+                    <p>Are you sure you want to enable this user?</p>
+                    <label class="checkbox" style="font-weight: normal">
+                        <i>preserve any pre-existing group membership (you don't want this)</i>
+                        <input type="checkbox" name="preserve_membership" value="true"/>
```
rra: Doesn't this make preserving membership the default?  I assume we'd want it the other way around.

grouper/models.py
```diff
@@ -514,7 +522,7 @@ def my_grantable_permissions(self):
                                    grantable[1] if len(grantable) > 1 else '*', ))
         return sorted(result, key=lambda x: x[0].name + x[1])
 
-    def my_groups(self):
+    def my_groups(self, ignore_enabled=False):
```
rra: Nit: I find `ignore_enabled` unintuitive, since I first thought that would filter out groups that were enabled.  Maybe `include_disabled` instead?

grouper/fe/templates/user.html
herb: this is actually default not checked (cause the 'checked' attribute is not present). is there a better way to write this or maybe a comment to make this not as confusing?

grouper/models.py
herb: yeah, this is essentially ignoring the fact the user is disabled. i renamed it to `ignore_user_disabled`. does that help?

grouper/fe/templates/user.html
rra: Oh, right, no, this is just me not remembering HTML.  Just ignore me.  :)

grouper/models.py
rra: Yup, sounds good!

PR #144, opened 2016-01-12
improvements to one-offs: (1) an enforced dry_run mode, (2) key=value arguments passed from grouper-ctl

herb:


rra:
Nice.  Looks good!  However, tests fail.

herb:
fixed flake8. checks pass now.

PR #152, opened 2016-01-26
test to verify role changes reflect in graph

herb:


lfaraone:
We should also verify that the checkpoint gets incremented.

PR #156, opened 2016-02-05
permission requests (and actions)

herb:
this implements everything on the request side. servicing the request is pending as is a real plugin.

requesting endpoint:
https://www.dropbox.com/s/f4jrbyc6s7gp4wu/Screenshot%202016-02-05%2015.10.52.png?dl=0

new group view with 'request permission' button and pending requests shown:
https://www.dropbox.com/s/kchrvdp724zrrvc/Screenshot%202016-02-05%2015.11.02.png?dl=0

### DIFF
--- a/grouper/fe/handlers.py
+++ b/grouper/fe/handlers.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+import json
 import operator
 
 from expvar.stats import stats
@@ -8,6 +9,7 @@ from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql import label, literal
 
 from .. import perf_profile
+from .. import permissions
 from .. import public_key
 from ..audit import assert_controllers_are_auditors, assert_can_join, get_audits, UserNotAuditor
 from ..constants import (
@@ -21,6 +23,7 @@ from .forms import (
     GroupEditForm,
     GroupEditMemberForm,
     GroupJoinForm,
+    GroupPermissionRequestForm,
     GroupRemoveForm,
     GroupRequestModifyForm,
     PermissionCreateForm,
@@ -33,12 +36,30 @@ from .forms import (
 from ..email_util import cancel_async_emails, send_email, send_async_email
 from ..graph import NoSuchUser, NoSuchGroup
 from ..models import (
-    User, Group, Request, PublicKey, Permission, PermissionMap, AuditLog, GroupEdge, Counter,
-    GROUP_JOIN_CHOICES, REQUEST_STATUS_CHOICES, GROUP_EDGE_ROLES, OBJ_TYPES,
-    get_all_groups, get_all_users,
-    get_user_or_group, Audit, AuditMember, AUDIT_STATUS_CHOICES, AuditLogCategory,
+    AUDIT_STATUS_CHOICES,
+    Audit,
+    AuditLog,
+    AuditLogCategory,
+    AuditMember,
+    Counter,
+    GROUP_EDGE_ROLES,
+    GROUP_JOIN_CHOICES,
+    Group,
+    GroupEdge,
+    OBJ_TYPES,
+    OWNER_ROLE_INDICES,
+    Permission,
+    PermissionMap,
+    PublicKey,
+    REQUEST_STATUS_CHOICES,
+    Request,
+    User,
     UserToken,
+    get_all_groups,
+    get_all_users,
+    get_user_or_group,
 )
+from ..permissions import get_grantable_permissions, get_pending_request_by_group
 from .settings import settings
 from .util import ensure_audit_security, GrouperHandler, Alert, test_reserved_names
 from ..util import matches_glob
@@ -566,9 +587,11 @@ class GroupView(GrouperHandler):
         members = group.my_members()
         groups = group.my_groups()
         permissions = group_md.get('permissions', [])
+        permission_requests_pending = get_pending_request_by_group(self.session, group)
         audited = group_md.get('audited', False)
         log_entries = group.my_log_entries()
         num_pending = group.my_requests("pending").count()
+        is_owner = self.current_user.my_role_index(members) in OWNER_ROLE_INDICES
 
         # Add mapping_id to permissions structure
         my_permissions = group.my_permissions()
@@ -588,7 +611,8 @@ class GroupView(GrouperHandler):
             "group.html", group=group, members=members, groups=groups,
             num_pending=num_pending, alerts=alerts, permissions=permissions,
             log_entries=log_entries, grantable=grantable, audited=audited,
-            statuses=AUDIT_STATUS_CHOICES,
+            statuses=AUDIT_STATUS_CHOICES, is_owner=is_owner,
+            permission_requests_pending=permission_requests_pending
         )
 
 
@@ -856,6 +880,93 @@ class GroupRequests(GrouperHandler):
             members=members, status=status, statuses=REQUEST_STATUS_CHOICES,
             offset=offset, limit=limit, total=total
         )
+
+
+class GroupPermissionRequest(GrouperHandler):
+    def _prep_form(self, form, args_by_perm):
+        form.permission_name.choices = [("", "")] + sorted([(p, p) for p in args_by_perm.keys()])
+        form.argument_select.choices = [("", "")]
+
+        return form
+
+    def _prep_form_for_display(self, form):
+        # hack: we don't want validation to puke if one is undefined but we
+        # want them to show up in the UI as such
+        form.argument_select.flags.required = True
+        form.argument_text.flags.required = True
+
+        return form
+
+    def get(self, group_id=None, name=None):
+        group = Group.get(self.session, group_id, name)
+        if not group:
+            return self.notfound()
+
+        form = GroupPermissionRequestForm()
+        args_by_perm = get_grantable_permissions(self.session)
+        form = self._prep_form_for_display(self._prep_form(form, args_by_perm))
+
+        self.render("group-permission-request.html", form=form, group=group,
+                args_by_perm_json=json.dumps(args_by_perm))
+
+    def post(self, group_id=None, name=None):
+        group = Group.get(self.session, group_id, name)
+        if not group:
+            return self.notfound()
+
+        role_index = self.current_user.my_role_index(group.my_members())
+        if role_index not in OWNER_ROLE_INDICES:
+            return self.forbidden()
+
+        form = GroupPermissionRequestForm(self.request.arguments)
+        args_by_perm = get_grantable_permissions(self.session)
+
+        form = self._prep_form(form, args_by_perm)
+
+        if not form.validate():
+            form = self._prep_form_for_display(form)
+            return self.render(
+                    "group-permission-request.html", form=form, group=group,
+                    args_by_perm_json=json.dumps(args_by_perm),
+                    alerts=self.get_form_alerts(form.errors),
+                    )
+
+        if form.argument_select.data:
+            argument = form.argument_select.data
+        else:
+            argument = form.argument_text.data
+        permission = Permission.get(self.session, form.permission_name.data)
+
+        calc_args = args_by_perm.get(permission.name, [])
+        if not permission or (argument not in calc_args and "*" not in calc_args):
+            return self.render(
+                    "group-permission-request.html", form=form, group=group,
+                    args_by_perm_json=json.dumps(args_by_perm),
+                    alerts=[
+                        Alert("danger", "no such permission or permission+argument pair"),
+                        ],
+                    )
+
+        # save off request
+        try:
+            permissions.create_request(self.session, self.current_user, group,
+                    permission, argument, form.reason.data)
+        except permissions.RequestAlreadyExists:
+            alerts = [Alert("danger", "request for permission + argument already exists")]
+        except permissions.NoOwnersAvailable:
+            alerts = [Alert("danger", "no owners available for requested permission + argument")]
+        else:
+            alerts = None
+
+        if alerts:
+            form = self._prep_form(form, args_by_perm)
+            return self.render(
+                    "group-permission-request.html", form=form, group=group,
+                    args_by_perm_json=json.dumps(args_by_perm),
+                    alerts=alerts,
+                    )
+        else:
+            return self.redirect("/groups/{}".format(group.name))
 
 
 class AuditsComplete(GrouperHandler):
--- a/grouper/fe/handlers.py
+++ b/grouper/fe/handlers.py
@@ -989,19 +989,17 @@ class GroupPermissionRequest(GrouperHandler):
 
 
 class PermissionsRequests(GrouperHandler):
-    """Allow a user to review a list of permission requests that they have both
-    un-actioned and previously actioned."""
+    """Allow a user to review a list of permission requests that they have."""
     def get(self):
         form = PermissionRequestsForm(self.request.arguments)
         form.status.choices = [("", "")] + [(k, k) for k in REQUEST_STATUS_CHOICES]
 
-        requests = []
-        alerts = []
-        total = 0
-
         if not form.validate():
             alerts = self.get_form_alerts(form.errors)
+            request_tuple = None
+            total = 0
         else:
+            alerts = []
             request_tuple, total = permissions.get_requests_by_owner(self.session,
                     self.current_user, status=form.status.data,
                     limit=form.limit.data, offset=form.offset.data)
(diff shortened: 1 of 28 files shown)


grouper/permissions.py
```diff
@@ -0,0 +1,245 @@
...
+            "permission_name": permission.name,
+            "argument": argument,
+            "reason": reason,
+            }
+
+    # TODO: would be nicer if it told you which group you're an approver of
+    # that's causing this notification
+    mail_to = []
+    for owner in owners:
+        mail_to += [u.name for u in owner.my_users() if u.role in APPROVER_ROLE_INDICIES]
```
rra: I think you want all the group members here, not just the ones in `APPROVER_ROLE_INDICIES`.  My mental model is that the Grouper permission to grant permissions applies to the entire group and any member can do the grant, so the email notification should go to all the members of the group.  That will be noisy for groups with lots of members, but I think groups with lots of members generally shouldn't have the ability to grant other permissions anyway.

rra:
Read through this all and looked at the screen shots and this generally looks good to me for the request side and for the UI.  I like that the pending permissions are shown greyed out in the group screen.  (Tests are failing, though.)

grouper/permissions.py
herb: good call. i fixed this.

herb:
this is ready for a more serious review. here's the review/action flow:
screenshot 2016-02-18 23 11 46

screenshot 2016-02-18 23 12 03

grouper/fe/handlers.py
```diff
@@ -858,6 +886,188 @@ def get(self, group_id=None, name=None):
...
+                        "group-permission-request.html", form=form, group=group,
+                        args_by_perm_json=json.dumps(args_by_perm),
+                        alerts=[
+                            Alert("danger", "this group already has this permission+argument pair"),
+                            ],
+                        )
+
+        # make sure there's an owner for this
+        calc_args = args_by_perm.get(permission.name, [])
+        if not permission or (argument not in calc_args and "*" not in calc_args):
```
rra: If `permission` is `None` (so the first part of this conditional fails), won't the previous line explode?

grouper/fe/handlers.py
```diff
@@ -858,6 +886,188 @@ def get(self, group_id=None, name=None):
...
+                    args_by_perm_json=json.dumps(args_by_perm),
+                    alerts=self.get_form_alerts(form.errors),
+                    )
+
+        # load requested permission
+        if form.argument_select.data:
+            argument = form.argument_select.data
+        else:
+            argument = form.argument_text.data
+        permission = Permission.get(self.session, form.permission_name.data)
```
rra: Does this gracefully handle the case that the permission doesn't exist?

grouper/fe/handlers.py
```diff
@@ -858,6 +886,188 @@ def get(self, group_id=None, name=None):
...
+                    )
+
+        # save off request
+        try:
+            permissions.create_request(self.session, self.current_user, group,
+                    permission, argument, form.reason.data)
+        except permissions.RequestAlreadyExists:
+            alerts = [Alert("danger", "request for permission + argument already exists")]
+        except permissions.NoOwnersAvailable:
+            alerts = [Alert("danger", "no owners available for requested permission + argument")]
```
rra: If we're checking again at `create_request` time if the group already has the permission or if there are no owners available, does it make sense to do the explicit checks above?  I assume you're catching these exceptions because you could still be racing another caller, in which case maybe it would make sense to just try to create the request and catch the exceptions as the primary flow rather than pre-checking?  It just feels like the code is doing the same thing twice.

grouper/fe/templates/email/permission_request_cancelled_html.tmpl
```diff
@@ -0,0 +1,19 @@
+{% extends "email/base_html.tmpl" %}
+
+{% block subject %}Request for Permission Cancelled{% endblock %}
```
rra: Nit: You could probably combine this with the actioned template with a bit more templating.  Not sure if it's worth it.

grouper/permissions.py
```diff
@@ -0,0 +1,409 @@
...
+        {models.Permission: [arg1, arg2, ...], ...}
+    """
+    owners_by_arg_by_perm = get_owners_by_grantable_permission(session)
+    args_by_perm = defaultdict(list)
+    for permission, owners_by_arg in owners_by_arg_by_perm.items():
+        for argument in owners_by_arg:
+            args_by_perm[permission].append(argument)
+
+    def _reduce_args(args):
+        return "*" if len(args) > 1 and "*" in args else args
```
rra: Nit: I think this should be `["*"]`.  Because of Python's treatment of strings as sequences, it ends up working either way, but it confused me a bit when I realized this didn't return a list.

grouper/fe/handlers.py
herb: maybe not but this is a great point. but i'm also going to remove this code block per a later comment. so moot.

grouper/fe/handlers.py
herb: yeah, not really. but arguably it shouldn't since we're seeding the permission and we don't delete permissions ever. going to add an assert here just in case.

grouper/fe/handlers.py
herb: yeah, this is good feedback. i'll consolidate these checks in `create_request`.

grouper/fe/templates/email/permission_request_cancelled_html.tmpl
herb: yeah, i wasn't sure either. i left them separate in case the messaging needs to diverge more (e.g. the current messaging isn't different enough and folks get confused).

grouper/permissions.py
herb: good catch. fixed.

grouper/fe/handlers.py
```diff
@@ -858,6 +886,172 @@ def get(self, group_id=None, name=None):
...
 
+class GroupPermissionRequest(GrouperHandler):
+    def _prep_form(self, form, args_by_perm):
+        form.permission_name.choices = [("", "")] + sorted([(p, p) for p in args_by_perm.keys()])
+        form.argument_select.choices = [("", "")]
+
+        return form
+
+    def _prep_form_for_display(self, form):
+        # hack: we don't want validation to puke if one is undefined but we
```
lfaraone: Why not?

grouper/fe/handlers.py
```diff
@@ -858,6 +886,172 @@ def get(self, group_id=None, name=None):
...
+        form = self._prep_form_for_display(form)
+
+        # load requested permission
+        if form.argument_select.data:
+            argument = form.argument_select.data
+        else:
+            argument = form.argument_text.data
+
+        permission = Permission.get(self.session, form.permission_name.data)
+        assert permission, "our preseeded permission should exist or we have problems"
```
lfaraone: `preseeded`?

grouper/fe/handlers.py
```diff
@@ -858,6 +886,172 @@ def get(self, group_id=None, name=None):
...
+
+        # save off request
+        try:
+            permissions.create_request(self.session, self.current_user, group,
+                    permission, argument, form.reason.data)
+        except permissions.RequestAlreadyGranted:
+            alerts = [Alert("danger", "this group already has this  permission + argument")]
```
lfaraone: Extra space: `this  permission`

grouper/fe/handlers.py
```diff
@@ -858,6 +886,172 @@ def get(self, group_id=None, name=None):
...
+        # check for request existence
+        request = permissions.get_request_by_id(self.session, request_id)
+        if not request:
+            return self.notfound()
+
+        # check that this user should be actioning this request
+        user_requests, total = permissions.get_requests_by_owner(self.session,
+                self.current_user, status="pending", limit=None, offset=0)
+        user_request_ids = [ur.id for ur in user_requests.requests]
+        if request.id not in user_request_ids:
```
lfaraone: Rather than getting all pending requests for the user and then doing a linear scan, why not just query the database with the above constraints plus the ID?

grouper/fe/handlers.py
herb: this form allows selecting an argument either from a dropbown or a free form text field, depending on what grants are available. so one of these fields should be undefined/empty. is there a better way to express this mutual exclusivity within wtform itself?

grouper/fe/handlers.py
herb: we seed a dropdown with possible permissions the user can select. is there a term more likely to be understood than `preseed` to describe this?

grouper/fe/handlers.py
herb: fixed

grouper/fe/handlers.py
herb: personal preference: my belief is that code (essentially what the existing membership request flow does) is more cryptic.

grouper/fe/handlers.py
lfaraone: `prefill`

herb: fixed.

grouper/fe/handlers.py
herb: ah, apparently the google says using two forms is the best approach. let me give that a whirl

herb: done. though the code got a bit more complicated. not sure if it was a win.

herb:
any other comments? i fixed a bug with grants on grants with argument wildcards. unclear if there's more functionality i'm missing but getting to the point some real usage would help make this more shallow. :)

rra:
I'm good to merge.  Luke, any more comments on your side?

grouper/fe/templates/group-permission-request.html
```diff
@@ -0,0 +1,100 @@
...
+                    $argument_select.append($option);
+                });
+            }
+        }
+
+        {# prime it #}
+        update_args();
+
+        $perm_fields.change(function() {
+            {# keep permission values sync'd #}
```
lfaraone: `synced` is both correct and will avoid breaking code highlighting :)

grouper/fe/handlers.py
```diff
@@ -858,6 +888,172 @@ def get(self, group_id=None, name=None):
         )
 
 
+class GroupPermissionRequest(GrouperHandler):
+    def _get_forms(self, args_by_perm, get_request_data=False):
+        if get_request_data:
```
lfaraone: This behaviour seems a bit magic. I'd prefer that this be refactored into a `@staticmethod` that always took the data as a parameter (or `None`)

grouper/fe/handlers.py
```diff
@@ -858,6 +888,172 @@ def get(self, group_id=None, name=None):
...
+
+        argument_type = self.request.arguments.get("argument_type")
+        if argument_type and argument_type[0] == "text":
+            form = text_form
+        elif argument_type and argument_type[0] == "dropdown":
+            form = dropdown_form
+            form.argument.choices = [(a, a) for a in args_by_perm[form.permission_name.data]]
+        else:
+            # someone messing with the form
+            return self.forbidden()
```
lfaraone: It'd be good to log this too, so we can look into bugs :)

grouper/fe/handlers.py
```diff
@@ -858,6 +888,172 @@ def get(self, group_id=None, name=None):
...
+
+        if not form.validate():
+            return self.render(
+                    "group-permission-request.html", dropdown_form=dropdown_form,
+                    text_form=text_form, group=group, args_by_perm_json=json.dumps(args_by_perm),
+                    alerts=self.get_form_alerts(form.errors),
+                    )
+
+        permission = Permission.get(self.session, form.permission_name.data)
+        assert permission, "our prefilled permission should exist or we have problems"
```
lfaraone: `permission is not None` is more explicit

grouper/fe/handlers.py
```diff
@@ -858,6 +888,172 @@ def get(self, group_id=None, name=None):
...
+
+        # save off request
+        try:
+            permissions.create_request(self.session, self.current_user, group,
+                    permission, form.argument.data, form.reason.data)
+        except permissions.RequestAlreadyGranted:
+            alerts = [Alert("danger", "this group already has this permission + argument")]
```
lfaraone: If this is shown to the user as-is (is not embedded into a larger sentence), capitalise the first letter of each fragment, spell out "and" instead of `+`.

It'd also be great if we told users what to do for each of these error conditions.

grouper/fe/handlers.py
```diff
@@ -858,6 +888,172 @@ def get(self, group_id=None, name=None):
...
+            alerts = None
+
+        if alerts:
+            return self.render(
+                    "group-permission-request.html", dropdown_form=dropdown_form,
+                    text_form=text_form, group=group, args_by_perm_json=json.dumps(args_by_perm),
+                    alerts=alerts,
+                    )
+        else:
+            return self.redirect("/groups/{}".format(group.name))
```
lfaraone: Do we show the user some confirmation that their request succeeded?

grouper/fe/handlers.py
lfaraone: You're asking the database "does this thing exist" — it seems perfectly reasonable to perform a maximally-specific query.

grouper/fe/templates/group-permission-request.html
```diff
@@ -0,0 +1,100 @@
...
+        var $argument_select = $('#dropdown_form #argument');
+        var $argument_text = $('#text_form #argument');
+
+        var $dropdown_form_div = $('#dropdown_form');
+        var $text_form_div = $('#text_form');
+
+        var $perm_fields = $('.input-permission_name');
+        var $reason_fields = $('.input-reason');
+
+        function update_args() {
```
lfaraone: Improve the commenting of the various cases in this function. Currently, they just cover the "what"; what's the "why"?

grouper/group.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    now = datetime.utcnow()
+    return session.query(Group, GroupEdge).filter(
+            GroupEdge.group_id == Group.id,
+            GroupEdge.member_pk == user.id,
+            GroupEdge.member_type == 0,
+            GroupEdge.active == True,
+            Group.enabled == True,
+            or_(GroupEdge.expiration > now, GroupEdge.expiration == None),
+            ).all()
```
lfaraone: Pull down the `)` to match the function indent

grouper/models.py
```diff
@@ -1680,6 +1627,48 @@ def my_log_entries(self):
...
+
+
+class PermissionRequestStatusChange(Model):
+    """Tracks changes to each permission grant request."""
+    __tablename__ = "permission_request_status_changes"
+
+    id = Column(Integer, primary_key=True)
+
+    request_id = Column(Integer, ForeignKey("permission_requests.id"), nullable=False)
+    request = relationship(PermissionRequest)
```
lfaraone: Worth adding a `backref` here?

grouper/fe/templates/group-permission-request.html
herb: fixed

grouper/fe/handlers.py
herb: done

grouper/fe/handlers.py
herb: done

grouper/fe/handlers.py
herb: done

grouper/fe/handlers.py
herb: good point. done.

grouper/fe/handlers.py
herb: yes, but it's probably not great right now. we show requests at the bottom as pending in the permission panel.

i have a todo to make this better (threading messages through redirects).

grouper/fe/templates/group-permission-request.html
herb: attempted

grouper/group.py
herb: can you elaborate? do you mean
```
return session.query(...
).all()
```
?

grouper/models.py
herb: ¯\_(ツ)_/¯

the membership request schema doesn't include it either. but mostly because we haven't needed to build the functionality of getting back from a random comment to a request. i suppose if we ever build a comment viewer it makes sense but i'm inclined to leave it off for now.

grouper/permissions.py
```diff
@@ -0,0 +1,431 @@
...
+    return owners_by_arg_by_perm
+
+
+def get_grantable_permissions(session):
+    """Returns all grantable permissions and their possible arguments. This
+    function attempts to reduce a permission's arguments to the most
+    permissive, i.e. if a wildcard argument exists, everything else is
+    discarded.
+
+    TODO: this only does a simplistic reduction and doesn't really understand
```
lfaraone: TODO should be in comments, not the docs

herb:
going to merge this after an out of band 'ok' from luke

PR #157, opened 2016-02-10
Database model for user access tokens

lfaraone:


grouper/models.py
```diff
@@ -631,6 +635,55 @@ def my_open_audits(self):
         ).all()
 
 
+def _make_secret():
+    return os.urandom(20).encode("hex")
+
+
+class UserToken(Model):
```
lfaraone: All of this needs some test cases...

tests/test_users.py
```diff
@@ -44,6 +44,28 @@ def test_basic_metadata(standard_graph, session, users, groups, permissions):  #
     assert len(md) == 1, "One metadata item"
 
 
+def test_usertokens(standard_graph, session, users, groups, permissions):  # noqa
```
lfaraone: Should I be doing something different to ensure I don't mutate shared test state?

tests/test_users.py
```diff
@@ -44,6 +44,28 @@ def test_basic_metadata(standard_graph, session, users, groups, permissions):  #
...
+    user = users["zorkian@a.co"]
```
lfaraone: Should I be doing something different to ensure I don't mutate shared test state?

grouper/models.py
```diff
@@ -1,3 +1,6 @@
+import hmac
+
+import os
```
herb: nit: alphabetical please

grouper/models.py
```diff
@@ -631,6 +635,59 @@ def my_open_audits(self):
...
+
+    __tablename__ = "user_tokens"
+
+    id = Column(Integer, primary_key=True)
+
+    user_id = Column(Integer, ForeignKey("users.id"))
+    name = Column(String(length=16), nullable=False)
+
+    ctime = Column(DateTime, default=datetime.utcnow, nullable=False)
+    dtime = Column(DateTime, default=None, nullable=True)
```
herb: i'm not a huge personal fan of the naming convention that exists but can we make these two more consistent with it? so `ctime` should be `created_on` or `created_at` and 'dtime' should be `disabled_on` or `disabled_at`

tests/test_users.py
herb: no, you're good. each test gets a new database and associated fixtures.

grouper/models.py
lfaraone: Fixed.

grouper/models.py
lfaraone: Fixed.

PR #159, opened 2016-02-17
Frontend for creating user tokens

lfaraone:
Screenshots of flow:

grouper_tok_addconf
grouper_tok_create
grouper_tok_disable
grouper_tok_disable_conf

### DIFF
--- a/grouper/fe/handlers.py
+++ b/grouper/fe/handlers.py
@@ -126,6 +126,7 @@ class UserView(GrouperHandler):
         log_entries = user.my_log_entries()
         self.render("user.html", user=user, groups=groups, public_keys=public_keys,
                     can_control=can_control, permissions=permissions,
+                    user_tokens=user.tokens,
                     log_entries=log_entries, num_pending_requests=num_pending_requests,
                     open_audits=open_audits)
 
--- a/grouper/fe/handlers.py
+++ b/grouper/fe/handlers.py
@@ -28,6 +28,7 @@ from .forms import (
     PublicKeyForm,
     UserEnableForm,
     UsersPublicKeyForm,
+    UserTokenForm,
 )
 from ..email_util import cancel_async_emails, send_email, send_async_email
 from ..graph import NoSuchUser, NoSuchGroup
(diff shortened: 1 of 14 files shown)


grouper/fe/handlers.py
```diff
@@ -1686,6 +1689,89 @@ def get(self):
...
+        if not form.validate():
+            return self.render(
+                "user-token-add.html", form=form, user=user,
+                alerts=self.get_form_alerts(form.errors),
+            )
+
+        try:
+            token = UserToken(name=form.data["name"], user=user)
+            token.add(self.session)
+            print token.secret
```
herb: is this debugging that can be removed?

grouper/fe/templates/email/user_tokens_changed_html.tmpl
```diff
@@ -0,0 +1,14 @@
...
+
+{% block content %}
+
+<p>User token {{ action }} for <strong><a href="{{url}}/users/{{changed_user}}">{{ changed_user }}</a></strong>.</p>
+
+{% if changed_user != actioner %}
+    <p>This edit was made by
+    <a href="{{url}}/users/{{actioner}}">{{ actioner }}</a>.</p>
+{% endif %}
+
```
herb: this will never happen, right? are you planning for administrative access to change these on another user's behalf?

herb:
can we add a simple test that creates and then disables the token? it not the best coverage-wise but it'll be better than nothing and gives us some assurance nothing basic is broken in the future.

grouper/fe/handlers.py
lfaraone: Yep, will nix.

grouper/fe/templates/email/user_tokens_changed_html.tmpl
lfaraone: Administrators can disable user tokens.

lfaraone:
> can we add a simple test that creates and then disables the token?

`tests/test_users.py:test_usertokens()`?

PR #161, opened 2016-02-22
Store secret hash, rather than secret

lfaraone:
`self.secret` is set to `None` by default. When `self.add()` is called, `self.secret` (not stored) is set to a randomly generated string, and the SHA256 of the secret is placed in `hashed_secret` (which is database-backed)

Afterwards, comparisons are performed by hashing the user-specified secret and comparing against the hashed secret in the database.

### DIFF
--- a/grouper/models.py
+++ b/grouper/models.py
@@ -653,9 +653,7 @@ class UserToken(Model):
     created_at = Column(DateTime, default=datetime.utcnow, nullable=False)
     disabled_at = Column(DateTime, default=None, nullable=True)
 
-    hashed_secret = Column('secret', String(length=64), unique=True, nullable=False)
-
-    secret = None
+    hashed_secret = Column(String(length=64), unique=True, nullable=False)
 
     user = relationship("User", back_populates="tokens")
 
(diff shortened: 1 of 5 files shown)


grouper/models.py
```diff
@@ -652,7 +653,9 @@ class UserToken(Model):
     created_at = Column(DateTime, default=datetime.utcnow, nullable=False)
     disabled_at = Column(DateTime, default=None, nullable=True)
 
-    secret = Column(String(length=32), default=_make_secret, unique=True, nullable=False)
+    hashed_secret = Column('secret', String(length=64), unique=True, nullable=False)
```
herb: any reason not to rename the column to 'hashed_secret' as well? seems to make it self-documenting in the schema and more consistent with the code.

grouper/models.py
```diff
@@ -652,7 +653,9 @@ class UserToken(Model):
...
+
+    secret = None
```
herb: this is a little confusing because most (all?) attributes of models in grouper are actual columns in the db. can we maybe make this a private attribute and hide the accessor behind a property function?

alternative solution, and maybe much cleaner, is to have the `add` method return both the model and the unhash secret in the return. that keeps the contract (attributes are columns) and makes it explicit the unhashed secret is ephemeral.

grouper/models.py
lfaraone: Done

grouper/models.py
lfaraone: Done.

PR #164, opened 2016-02-24
fix perm request

herb:
various fixes for the permission request flow.

### DIFF
--- a/grouper/fe/templates/group-permission-request.html
+++ b/grouper/fe/templates/group-permission-request.html
@@ -12,6 +12,7 @@
 <script type="text/javascript">
     $(function () {
         var args_by_perm = {{ args_by_perm_json }};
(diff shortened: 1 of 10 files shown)


grouper/fe/templates/group-permission-request.html
```diff
@@ -12,6 +12,7 @@
 <script type="text/javascript">
     $(function () {
         var args_by_perm = {{ args_by_perm_json }};
+        console.log('args_by_perm', args_by_perm);
```
rra: Stray debugging code?

herb: thanks for catching that. removed.

herb:
@lfaraone @rra can one of you guys take a look at this? it's ready for review now.

PR #166, opened 2016-02-24
more detail to permission request edit page

herb:


herb:
before:
screenshot 2016-02-24 14 04 23

after:
screenshot 2016-02-24 14 04 00

grouper/fe/templates/permission-request-update.html
```diff
@@ -30,8 +30,12 @@
                 </tr>
                 <tr>
                     <td colspan="5">
-                        <span style="padding-left: 10px;">
-                            <strong>Reason:</strong> <em>{{comment.comment}}</em>
+                        <dl class="dl-horizontal" style="margin-bottom: 0px;">
+                            <dt>Permission:<dt>
+                            <dd><a href="/permissions/{{ request.permission.name }}">
+                                {{ request.permission.name }}</a></dd>
```
lfaraone: We don't have a linter for our templates, but I think ideal style is to put closing tags on a separate line from content if the opening tags were.

grouper/fe/templates/permission-request-update.html
```diff
@@ -30,8 +30,12 @@
                 </tr>
                 <tr>
                     <td colspan="5">
-                        <span style="padding-left: 10px;">
-                            <strong>Reason:</strong> <em>{{comment.comment}}</em>
+                        <dl class="dl-horizontal" style="margin-bottom: 0px;">
```
lfaraone: This isn't tabular data, so it shouldn't be nested in a table. Using tables for layout is bad news bears.

grouper/fe/templates/permission-request-update.html
herb: fixed

grouper/fe/templates/permission-request-update.html
herb: i copied/pasted from the join request flow. so yeah, probably a TODO for cleanup.

PR #169, opened 2016-03-02
allow for selectively excluding wildcard perm ownership from display

herb:


grouper/permissions.py
```diff
@@ -109,13 +109,16 @@ def get_owners_by_grantable_permission(session):
...
 
-def get_grantable_permissions(session):
+def get_grantable_permissions(session, restricted_ownership_permissions):
     """Returns all grantable permissions and their possible arguments. This
     function attempts to reduce a permission's arguments to the least
     permissive possible.
 
     Args:
         session(sqlalchemy.orm.session.Session): database session
+        restricted_ownership_permissions([str]): list of permissions for which
```
lfaraone: Should be `List[str]`.

lfaraone:
Otherwise LGTM to merge.

grouper/permissions.py
herb: thanks. forced an update. will merge in a tad.

PR #179, opened 2016-03-22
provide optional help text for perm request argument when dropdown is present

herb:
screenshot:
screenshot 2016-03-22 11 35 09

### DIFF
--- a/grouper/fe/handlers.py
+++ b/grouper/fe/handlers.py
@@ -973,7 +973,9 @@ class GroupPermissionRequest(GrouperHandler):
         dropdown_form, text_form = GroupPermissionRequest._get_forms(args_by_perm, None)
 
         self.render("group-permission-request.html", dropdown_form=dropdown_form,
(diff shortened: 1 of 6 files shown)


grouper/fe/handlers.py
```diff
@@ -973,7 +973,9 @@ def get(self, group_id=None, name=None):
         dropdown_form, text_form = GroupPermissionRequest._get_forms(args_by_perm, None)
 
         self.render("group-permission-request.html", dropdown_form=dropdown_form,
-                text_form=text_form, group=group, args_by_perm_json=json.dumps(args_by_perm))
+                text_form=text_form, group=group, args_by_perm_json=json.dumps(args_by_perm),
+                dropdown_help=settings.permission_request_dropdown_help,
+                text_help=settings.permission_request_text_help)
```
lfaraone: We don't have a linter for this, but it'd help minimize diffs in the future if we always ended lists with a trailing comma and put the closing `)` on the next line, as below.

PR #182, opened 2016-03-23
split handlers into individual modules

herb:


herb:
oops, one more change to make first.

PR #188, opened 2016-03-31
Add list_users, enable and disable functionality to grouper-ctl

santoshankr:
### New Subcommands

To list all users and their account status (enabled / disabled), run

`bin/grouper-ctl -vv user list`

The `enable` and `disable` subcommands use the same syntax as `create`. One or more usernames must be provided. 
```$> bin/grouper-ctl -vv user disable username [ username ... ]```
```$> bin/grouper-ctl -vv user disable username [ username ... ]```

`enable` also allows an additional `--preserve-membership` flag that reactivates a user without revoking their access to any groups. This flag is set once per command, and applies to all accounts being actioned.
```$> bin/grouper-ctl -vv user enable username [ username ... ] --preserve-membership```

### DIFF
--- a/grouper/ctl/user.py
+++ b/grouper/ctl/user.py
@@ -10,7 +10,7 @@ def handle_command(args):
         session = make_session()
         all_users = get_all_users(session)
         for user in all_users:
-            user_enabled = "enabled" if user.is_enabled else "disabled"
+            user_enabled = "enabled" if user.enabled else "disabled"
             logging.info("{} has status {}".format(user.name, user_enabled))
         return
 
@@ -38,7 +38,7 @@ def user_command(args):
             user = User.get(session, name=username)
             if not user:
                 logging.info("{}: No such user. Doing nothing.".format(username))
-            elif not user.is_enabled:
+            elif not user.enabled:
                 logging.info("{}: User already disabled. Doing nothing.".format(username))
             else:
                 logging.info("{}: User found, disabling...".format(username))
@@ -51,7 +51,7 @@ def user_command(args):
             user = User.get(session, name=username)
             if not user:
                 logging.info("{}: No such user. Doing nothing.".format(username))
-            elif user.is_enabled:
+            elif user.enabled:
                 logging.info("{}: User not disabled. Doing nothing.".format(username))
             else:
                 logging.info("{}: User found, enabling...".format(username))
@@ -96,7 +96,7 @@ def add_parser(subparsers):
     user_parser.set_defaults(func=handle_command)
     user_subparser = user_parser.add_subparsers(dest="subcommand")
 
-    user_list_parser = user_subparser.add_parser(
+    user_subparser.add_parser(
         "list", help="List all users and their account statuses")
 
     user_create_parser = user_subparser.add_parser(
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: grouper/ctl/user.py [hunk 1]
```diff
@@ -2,7 +2,15 @@
 
 from grouper import public_key
 from grouper.ctl.util import ensure_valid_username, make_session
-from grouper.models import AuditLog, User
+from grouper.models import AuditLog, User, get_all_users
```
lfaraone: @herb remind me, do we alphabetize imports as case-insensitive, or case-sensitive?

grouper/ctl/user.py
```diff
@@ -53,6 +84,10 @@ def user_command(args):
 
 
 def add_parser(subparsers):
+    list_users_parser = subparsers.add_parser(
+        "list_users", help="List all users and their account statuses")
```
lfaraone: should probably instead be a subcommand of `user`

grouper/ctl/user.py
```diff
@@ -64,6 +99,16 @@ def add_parser(subparsers):
     user_create_parser.add_argument("--role-user", default=False, action="store_true",
                                     help="If given, identifies user as a role user.")
 
+    user_disable_parser = user_subparser.add_parser(
+        "disable", help="Disable a user account")
```
lfaraone: same here

grouper/ctl/user.py
```diff
@@ -64,6 +99,16 @@ def add_parser(subparsers):
...
+    user_disable_parser.add_argument("username", nargs="+")
+
+    user_enable_parser = user_subparser.add_parser(
+        "enable", help="(Re-)enable a user account")
```
lfaraone: same here

grouper/models.py
```diff
@@ -253,6 +253,10 @@ def name(self):
     def type(self):
         return "User"
 
+    @property
+    def is_enabled(self):
+        return self.enabled
```
lfaraone: why not just directly reference `enabled`? if we need to wrap it later, we can always make it a property.

grouper/ctl/user.py
```diff
@@ -18,7 +24,32 @@ def user_command(args):
                 session.commit()
             else:
                 logging.info("{}: Already exists. Doing nothing.".format(username))
+        return
```
lfaraone: explicit return isn't needed since we're only using one if/elif block

grouper/models.py
santoshankr: More future-proof if we ever want to do more here, set this value from outside the class or change the attribute name.

lfaraone: If we want to do that in the future, you rename `enabled` to `_enabled`, and make `enabled` a property.

Since right now we have a bunch of existing code referencing `enabled` (I assume), we end up with two different ways of checking the same thing that can get out of sync.

santoshankr: I thought it wasn't used anywhere outside of the User class, but on closer inspection, both the graph and the fe handlers read it directly, so it makes sense to keep that consistent. Fixing!

grouper/ctl/user.py
herb: i believe we're inconsistent. :/ fwiw, i've started doing whatever `:sort` does in VI. ;)

PR #189, opened 2016-04-05
Gradual refactor models

herb:


herb:
80% of the models have been pulled out. the remaining 20% is pretty tightly tangled. I've started in on disentangling Permission but ran out of time. will keep that diff around for when free time presents itself.

PR #194, opened 2016-04-14
After joining a Grouper group you should still be able to join a group you own

lfaraone:


lfaraone:
@rra: Can you test this to make sure it works? I'm having weird UI consistency issues when testing locally.

grouper/fe/templates/group.html
```diff
@@ -40,7 +40,7 @@
 
     <!-- Show the join button if the user isn't already a member and the group
          isn't invite-only. -->
-    {% if not current_user.my_role(members) and group.canjoin != "nobody" %}
+    {% if not current_user.my_role(members) in ("manager", "np-owner", "owner") and group.canjoin != "nobody" %}
```
herb: you should check `current_user_role['is_approver']` instead.

grouper/fe/templates/group.html
```diff
@@ -40,7 +40,7 @@
 
     <!-- Show the join button if the user isn't already a member and the group
          isn't invite-only. -->
-    {% if not current_user.my_role(members) and group.canjoin != "nobody" %}
+    {% if not current_user['is_approver'] and group.canjoin != "nobody" %}
```
herb: huh. you want `current_user_role` vs `current_user`. i.e. `current_user_role['is_approver']`. i guess selenium would have caught this...

lfaraone: Oops, fixed, thanks.

PR #196, opened 2016-04-15
Allow members to request permissions

lfaraone:


herb:
we should probably pre-calc that 'is_member' check but that's not your problem. :)

PR #198, opened 2016-04-25
revoking permission should update the graph

herb:


grouper/fe/handlers/permissions_grant.py
```diff
@@ -59,6 +59,7 @@ def post(self, name=None):
             if perm[0].name == permission.name:
                 if matches_glob(perm[1], form.data["argument"]):
                     allowed = True
+                    break
```
lfaraone: is this change related?

lfaraone: (confirmed no in person, but this is fine)

PR #202, opened 2016-05-26
Make minor tweaks to the Public Key and Counter schemas to fix MySQL …

Acidity:
…support

Attempting to autogenerate the SQL tables using grouper-ctl sync_db fails on MySQL systems.
This is due to the lack of length for the name column of the Counter model (this field is
a String type which translates to a SQL VARCHAR, which must have a length for MySQL). We
simply add a sufficiently large length (255) to this column in order to resolve the issue.
Additionally, the public_key column (type Text) of the PublicKey model was marked as unique.
MySQL doesn't support unique constraints on columns that do not have a specified length, so
we remove the unique constraint on the public_key column and instead move it to the fingerprint
column, which does have a maximum length, and should also be unique.

Signed-off-by: Tyler O'Meara <tyleromeara@dropbox.com>

### DIFF
--- a/grouper/models/public_key.py
+++ b/grouper/models/public_key.py
@@ -15,6 +15,6 @@ class PublicKey(Model):
 
     key_type = Column(String(length=32))
     key_size = Column(Integer)
-    public_key = Column(Text, nullable=False, unique=True)
-    fingerprint = Column(String(length=64), nullable=False)
+    public_key = Column(Text, nullable=False)
+    fingerprint = Column(String(length=64), nullable=False, unique=True)
     created_on = Column(DateTime, default=datetime.utcnow, nullable=False)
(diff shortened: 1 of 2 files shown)


grouper/models/public_key.py
```diff
@@ -15,6 +15,6 @@ class PublicKey(Model):
 
     key_type = Column(String(length=32))
     key_size = Column(Integer)
-    public_key = Column(Text, nullable=False, unique=True)
-    fingerprint = Column(String(length=64), nullable=False)
+    public_key = Column(Text, nullable=False)
+    fingerprint = Column(String(length=64), nullable=False, unique=True)
```
lfaraone: This is fine, since we're relying on `fingerprint` to avoid collisions.

PR #203, opened 2016-05-26
Add request update emails for group approvers

Acidity:
Whenever an approver updates a group request, grouper now alerts
all approvers for that group (except the approver who updated the
request) of the changed status of the request.

### DIFF
--- a/grouper/fe/handlers/group_request_update.py
+++ b/grouper/fe/handlers/group_request_update.py
@@ -3,7 +3,7 @@ from grouper.email_util import send_email
 from grouper.fe.forms import GroupRequestModifyForm
 from grouper.fe.settings import settings
 from grouper.fe.util import GrouperHandler, Alert
-from grouper.model_soup import Group, GroupEdge, Request
+from grouper.model_soup import Group, GroupEdge, Request, GROUP_EDGE_ROLES
 from grouper.models.base.constants import REQUEST_STATUS_CHOICES
 from grouper.models.audit_log import AuditLog
 
@@ -92,6 +92,31 @@ class GroupRequestUpdate(GrouperHandler):
         edge = self.session.query(GroupEdge).filter_by(
             id=request.edge_id
         ).one()
+
+        approver_mail_to = [
+            user.name
+            for user in group.my_users()
+            if GROUP_EDGE_ROLES[user.role] in ("manager", "owner", "np-owner") and
+            user.name != self.current_user.name and user.name != request.requester.username
+        ]
+
+        send_email(
+            self.session,
+            approver_mail_to,
+            "Request to join {} by {} has been {}".format(group.groupname,
+                request.requester.name, form.data['status']),
+            "approver_request_updated",
+            settings,
+            {
+                'group': group.name,
+                'requester': request.requester.username,
+                'changed_by': self.current_user.name,
+                'status': form.data['status'],
+                'role': edge.role,
+                'reason': form.data['reason'],
+            },
+        )
+
         if form.data['status'] == 'actioned':
             send_email(
                 self.session,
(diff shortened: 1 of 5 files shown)


grouper/fe/handlers/group_request_update.py
```diff
@@ -92,6 +92,31 @@ def post(self, request_id, group_id=None, name=None):
         edge = self.session.query(GroupEdge).filter_by(
             id=request.edge_id
         ).one()
+
+        approver_mail_to = [
+            user.name
+            for user in group.my_users()
+            if GROUP_EDGE_ROLES[user.role] in ("manager", "owner", "np-owner") and
```
herb: you should use the `APPROVER_ROLE_INDICIES` set in `model_soup` instead of defining the set of roles inline. you'd need to use `user._role` instead.

herb:
can you also flesh out the test `test_requests.test_basic_request` to check the number of emails sent for both the initial request and the one you just added?

there should be reasonable examples elsewhere where we just check # of notifications sent.

Acidity:
Fixed herb's issues :sun_with_face:

grouper/model_soup.py
```diff
@@ -66,7 +66,7 @@
 )
 OWNER_ROLE_INDICES = set([GROUP_EDGE_ROLES.index("owner"), GROUP_EDGE_ROLES.index("np-owner")])
 APPROVER_ROLE_INDICIES = set([GROUP_EDGE_ROLES.index("owner"), GROUP_EDGE_ROLES.index("np-owner"),
-        GROUP_EDGE_ROLES.index("manager")])
+        GROUP_EDGE_ROLES.index("manager")])  # TODO: Fix spelling
```
lfaraone: What's the tpyo? :P

tests/test_fe_handlers.py
```diff
@@ -100,3 +100,72 @@ def test_usertokens(session, users, http_client, base_url):
             headers={'X-Grouper-User': user.username})
     assert resp.code == 200
     assert "Disabled token: myFoobarToken" in resp.body
+
+@pytest.mark.gen_test
+def test_request_emails(graph, groups, permissions, session, standard_graph, users, base_url,
+                        http_client):
+    tech = groups["tech-ops"]
```
herb: thanks for doing this!

any reason you decided to do an entirely new test rather than adding email count checks to `test.test_requests.test_basic_request`?

tests/test_fe_handlers.py
```diff
@@ -100,3 +100,72 @@ def test_usertokens(session, users, http_client, base_url):
...
+    resp = yield http_client.fetch(fe_url, method="POST",
+            body=urlencode({"reason": "Test Request Please Ignore", "member": "User: {}".format(user.name)}),
+            headers={'X-Grouper-User': user.username})
+    assert resp.code == 200
+
+    assert session.query(AsyncNotification).count() == 2, "Only approvers for the requested group should receive an email"
+    assert session.query(AsyncNotification).filter_by(email="zay@a.co").count() == 1, "Owners should receive exactly one email per canask request"
+    assert session.query(AsyncNotification).filter_by(email="figurehead@a.co").count() == 1, "NP-Owners should receive exactly one email per canask request"
+
+    assert session.query(Request).count() == before_reqs + 1, "There should only be one added request"
```
herb: these checks that just select count against `AsyncNotification` are a tad brittle (can't refactor by moving code blocks around).

better options are
1. calling `grouper.email_util.process_async_emails` directly (which returns a count of sent emails). or
2. doing something similar to `tests.test_permissions._get_unsent_and_mark_as_sent_emails` that takes a username.

(2) probably is more inline with what you're doing.

tests/test_fe_handlers.py
Acidity: Emails are sent via fe code, which the requests tests don't touch (because they use the test.util functions which in turn directly use the Group methods)

grouper/model_soup.py
Acidity: The comment is supposed to be for the line above, APPROVER_ROLE_INDICIES has indices misspelled.

Acidity:
Made the tests better

tests/test_fe_handlers.py
herb: ah, right. looks good.

PR #204, opened 2016-05-27
Fix group url typos in the request update email templates

Acidity:
I opted to retain the use of "group" as the template variable
for the requested group name rather than the "requested" name
used in the pending_request template because I think group is
more readable. It might be a good idea to change the name of
the variable in pending_request to sync it with these
templates.


herb:
agreed we should sync the variable names. want to add that to this PR? either `group` or more explicitly `requested_group` feel right to me.

Acidity:
Updated the PR with the template variables all changed to group_name

PR #207, opened 2016-05-27
Add support for default expiration times for groups

Acidity:
Adds the auto_expire attribute to Group. When a user requests to
join a group (only via the front end), if they do not provide an
expiration date, their expiration is automatically set to auto_expire
days in the future. This happens for both canask and canjoin requests,
but does not apply to when a user is added to a group via a non
front end mechanism, or when an approver directly adds a member to
the group.

Ensures that the value of Group.auto_expire is correctly used where
appropriate (when a user submits a request and doesn't provide an
expiration time) and is not used where inappropriate (a user submits
a request and does provide an expiration time, an approver adds the
user via the add member feature, or the approver edits a member in
the group and leaves the expiration blank)

### DIFF
--- a/grouper/fe/forms.py
+++ b/grouper/fe/forms.py
@@ -42,7 +42,7 @@ class ValidateDate(object):
 class DaysTimeDeltaField(IntegerField):
 
     def process_data(self, value):
-        """Converting from TimeDelta notation to a simple integer"""
+        """Converts from TimeDelta notation to a simple integer"""
         # Do the check because groups with no expiration will result in value being None
         if hasattr(value, "days"):
             self.data = value.days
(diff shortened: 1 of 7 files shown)


grouper/fe/forms.py
```diff
@@ -39,6 +39,25 @@ def __call__(self, form, field):
                 "{} does not match format 'MM/DD/YYYY'".format(field.data))
 
 
+class DaysTimeDeltaField(IntegerField):
+
+    def process_data(self, value):
+        """Converting from TimeDelta notation to a simple integer"""
```
lfaraone: Docstrings should be written in imperative style. `"""Convert from…"""`

(I personally prefer the `"""Converts from """` style, but it looks like we mostly use the former. Except in cases where I wrote the code, it looks like. :P )

Acidity:
Updated docstring

lfaraone:
LGTM

PR #208, opened 2016-06-03
Add support for public key tags

Acidity:
Allows public keys to only have a subset of the permissions that the public key's owner has, in order to enable having separate public keys for different use cases (and therefore limit the danger of credential compromise)

### DIFF
--- a/grouper/fe/handlers/permissions_grant_tag.py
+++ b/grouper/fe/handlers/permissions_grant_tag.py
@@ -1,11 +1,10 @@
-from sqlalchemy.exc import IntegrityError
+from grouper.constants import TAG_EDIT
 from grouper.fe.forms import PermissionGrantTagForm
 from grouper.fe.util import GrouperHandler
 from grouper.models.audit_log import AuditLog
-from grouper.permissions import grant_permission_to_tag
 from grouper.models.permission import Permission
-from grouper.constants import TAG_EDIT
 from grouper.models.public_key_tag import PublicKeyTag
+from grouper.permissions import grant_permission_to_tag
 
 
 class PermissionsGrantTag(GrouperHandler):
@@ -51,20 +50,17 @@ class PermissionsGrantTag(GrouperHandler):
         if not permission:
             return self.notfound()  # Shouldn't happen.
 
-        try:
-            grant_permission_to_tag(self.session, tag.id, permission.id,
+        success = grant_permission_to_tag(self.session, tag.id, permission.id,
                                     argument=form.data["argument"])
-        except IntegrityError:
+
+        if not success:
             form.argument.errors.append(
                 "Permission and Argument already mapped to this tag."
             )
-            self.session.rollback()
             return self.render(
                 "permission-grant-tag.html", form=form, tag=tag,
                 alerts=self.get_form_alerts(form.errors),
             )
-
-        self.session.commit()
 
         AuditLog.log(self.session, self.current_user.id, 'grant_permission_tag',
                      'Granted permission with argument: {}'.format(form.data["argument"]),
--- a/grouper/fe/handlers/permissions_grant_tag.py
+++ b/grouper/fe/handlers/permissions_grant_tag.py
@@ -5,6 +5,7 @@ from grouper.models.audit_log import AuditLog
 from grouper.models.permission import Permission
 from grouper.models.public_key_tag import PublicKeyTag
 from grouper.permissions import grant_permission_to_tag
+from grouper.user_permissions import user_has_permission
 
 
 class PermissionsGrantTag(GrouperHandler):
@@ -13,7 +14,7 @@ class PermissionsGrantTag(GrouperHandler):
         if not tag:
             return self.notfound()
 
-        if not self.current_user.has_permission(TAG_EDIT, tag.name):
+        if not user_has_permission(self.session, self.current_user, TAG_EDIT, tag.name):
             return self.forbidden()
 
         form = PermissionGrantTagForm()
@@ -31,7 +32,7 @@ class PermissionsGrantTag(GrouperHandler):
         if not tag:
             return self.notfound()
 
-        if not self.current_user.has_permission(TAG_EDIT, tag.name):
+        if not user_has_permission(self.session, self.current_user, TAG_EDIT, tag.name):
             return self.forbidden()
 
         form = PermissionGrantTagForm(self.request.arguments)
--- a/grouper/fe/handlers/permissions_grant_tag.py
+++ b/grouper/fe/handlers/permissions_grant_tag.py
@@ -1,0 +1,73 @@
+from sqlalchemy.exc import IntegrityError
+from grouper.fe.forms import PermissionGrantTagForm
+from grouper.fe.util import GrouperHandler
+from grouper.models.audit_log import AuditLog
+from grouper.permissions import grant_permission_to_tag
+from grouper.models.permission import Permission
+from grouper.constants import TAG_EDIT
+from grouper.models.public_key_tag import PublicKeyTag
+
+
+class PermissionsGrantTag(GrouperHandler):
+    def get(self, name=None):
+        tag = PublicKeyTag.get(self.session, None, name)
+        if not tag:
+            return self.notfound()
+
+        if not self.current_user.has_permission(TAG_EDIT, tag.name):
+            return self.forbidden()
+
+        form = PermissionGrantTagForm()
+        form.permission.choices = [["", "(select one)"]]
+
+        for perm in self.session.query(Permission).all():
+            form.permission.choices.append([perm.name, "{} (*)".format(perm.name)])
+
+        return self.render(
+            "permission-grant-tag.html", form=form, tag=tag,
+        )
+
+    def post(self, name=None):
+        tag = PublicKeyTag.get(self.session, None, name)
+        if not tag:
+            return self.notfound()
+
+        if not self.current_user.has_permission(TAG_EDIT, tag.name):
+            return self.forbidden()
+
+        form = PermissionGrantTagForm(self.request.arguments)
+        form.permission.choices = [["", "(select one)"]]
+
+        for perm in self.session.query(Permission).all():
+            form.permission.choices.append([perm.name, "{} (*)".format(perm.name)])
+
+        if not form.validate():
+            return self.render(
+                "permission-grant-tag.html", form=form, tag=tag,
+                alerts=self.get_form_alerts(form.errors)
+            )
+
+        permission = Permission.get(self.session, form.data["permission"])
+        if not permission:
+            return self.notfound()  # Shouldn't happen.
+
+        try:
+            grant_permission_to_tag(self.session, tag.id, permission.id,
+                                    argument=form.data["argument"])
+        except IntegrityError:
+            form.argument.errors.append(
+                "Permission and Argument already mapped to this tag."
+            )
+            self.session.rollback()
+            return self.render(
+                "permission-grant-tag.html", form=form, tag=tag,
+                alerts=self.get_form_alerts(form.errors),
+            )
+
+        self.session.commit()
+
+        AuditLog.log(self.session, self.current_user.id, 'grant_permission_tag',
+                     'Granted permission with argument: {}'.format(form.data["argument"]),
+                     on_permission_id=permission.id, on_tag_id=tag.id)
+
+        return self.redirect("/tags/{}?refresh=yes".format(tag.name))
--- a/grouper/fe/handlers/permissions_revoke_tag.py
+++ b/grouper/fe/handlers/permissions_revoke_tag.py
@@ -1,0 +1,39 @@
+from grouper.fe.util import GrouperHandler
+from grouper.models.audit_log import AuditLog
+from grouper.models.counter import Counter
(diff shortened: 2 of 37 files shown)


grouper/fe/handlers/permissions_grant_tag.py
```diff
@@ -0,0 +1,73 @@
+from sqlalchemy.exc import IntegrityError
+from grouper.fe.forms import PermissionGrantTagForm
```
herb: nit: blank space between 3rd party and grouper modules

grouper/fe/handlers/permissions_grant_tag.py
```diff
@@ -0,0 +1,73 @@
...
+from grouper.fe.util import GrouperHandler
+from grouper.models.audit_log import AuditLog
+from grouper.permissions import grant_permission_to_tag
+from grouper.models.permission import Permission
+from grouper.constants import TAG_EDIT
```
herb: nit: alphabetical sort imports please, probably should get flake8 to check this ([extra credit](https://github.com/public/flake8-import-order))

grouper/fe/handlers/permissions_grant_tag.py
```diff
@@ -0,0 +1,73 @@
...
+            form.argument.errors.append(
+                "Permission and Argument already mapped to this tag."
+            )
+            self.session.rollback()
+            return self.render(
+                "permission-grant-tag.html", form=form, tag=tag,
+                alerts=self.get_form_alerts(form.errors),
+            )
+
+        self.session.commit()
```
herb: is this commit superfluous? i want to say `grant_pemrission_to_tag` will commit as well.

grouper/fe/handlers/permissions_revoke_tag.py
```diff
@@ -0,0 +1,39 @@
+from grouper.fe.util import GrouperHandler
+from grouper.models.audit_log import AuditLog
+from grouper.models.counter import Counter
+from grouper.models.tag_permission_map import TagPermissionMap
+from grouper.constants import TAG_EDIT
```
herb: nit: alphabetical sort imports please

grouper/models/public_key.py
```diff
@@ -2,6 +2,63 @@
 from sqlalchemy import Column, Integer, String, Text, ForeignKey, DateTime
 from sqlalchemy.orm import relationship
 from grouper.models.base.model_base import Model
+from grouper.models.public_key_tag_map import PublicKeyTagMap
+
+
+def _perm_list_to_dict(perms):
+    # type: List[Permission] -> Dict[str, Dict[str, Permission]]
```
herb: i believe this will fail `mypy` since
1. Permission is not imported
2. and what's returned is actually a `sqlalchemy.engine.RowProxy` more like a Permission Tuple. :/

grouper/models/public_key.py
```diff
@@ -2,6 +2,63 @@
...
+    stores the Permission object
+
+    Args:
+        perms: a list containing Permission objects
+
+    Returns:
+        a dictionary with the permission names as keys, and has as values another dictionary
+        with permission arguments as keys and Permission objects as values
+    """
+    ret = dict()
```
herb: nit: prefer using the literal, i.e. `ret = {}`

better yet, use `defaultdict` and avoid line 24.

grouper/models/public_key.py
```diff
@@ -2,6 +2,63 @@
...
+            continue
+        # According to Group.has_permission, None as an argument counts as a wildcard too
+        if None in pdict_b[perm.name]:
+            ret.add(perm)
+            continue
+        # If this permission is a wildcard, we add all permissions with the same name from
+        # the other set
+        if perm.argument == "*" or perm.argument is None:
+            ret |= {p for p in pdict_b[perm.name].values()}
+    return ret
```
herb: I think I get why you put these guys here (they're only used within this module) but they feel like they belong with the rest of permission code since it's whole-y to do with permissions. Thoughts on putting this in `grouper.permissions` or some such?

grouper/models/public_key.py
```diff
@@ -18,3 +75,35 @@ class PublicKey(Model):
...
+        Returns:
+            a list of all permissions this public key has
+        """
+        my_perms = self.user.my_permissions()
+        for tag in self.tags():
+            my_perms = _permission_intersection(my_perms, tag.my_permissions())
+
+        return list(my_perms)
+
+    def pretty_my_permissions(self):
```
herb: We do this in a few different places. primarily in the `pemission_panel`. It'd be nice if this was kept in templating/display code vs something in business logic.

grouper/models/public_key.py
```diff
@@ -18,3 +75,35 @@ class PublicKey(Model):
...
+        # type: () -> List[str]
+        """Returns a list of this public key's permissions formatted nicely as strings because
+        tyleromeara@ got tired of figuring out how to do this in Jinja."""
+        return ["{} ({})".format(perm.name, perm.argument if perm.argument else "unargumented")
+               for perm in self.my_permissions()]
```
herb: we're trying to avoid coupling the schema definition with business logic. this is mostly the reason `model_soup.py` is a soup. can you pull these guys into a `grouper` level module (e.g. `grouper/audit.py`')?

grouper/models/public_key_tag.py
```diff
@@ -0,0 +1,55 @@
+from sqlalchemy import Column, Integer, String, Text, Boolean
+from grouper.models.base.model_base import Model
+from grouper.constants import MAX_NAME_LENGTH
+from grouper.models.tag_permission_map import TagPermissionMap
+from grouper.models.permission import Permission
+from sqlalchemy.sql import label
+from grouper.models.audit_log import AuditLog
```
herb: nit: alphabetical imports, separating 3rd party and grouper modules

grouper/models/public_key_tag_map.py
```diff
@@ -0,0 +1,32 @@
+from datetime import datetime
+from sqlalchemy import Column, Integer, UniqueConstraint, ForeignKey, DateTime
+from sqlalchemy.orm import relationship
+from grouper.models.base.model_base import Model
+from grouper.models.public_key_tag import PublicKeyTag
+
```
herb: ditto on import ordering and separation

grouper/permissions.py
```diff
@@ -59,6 +60,31 @@ def grant_permission(session, group_id, permission_id, argument=''):
...
+        AssertError if argument does not match ARGUMENT_VALIDATION regex
+    """
+    assert re.match(ARGUMENT_VALIDATION, argument), 'Permission argument does not match regex.'
+
+    mapping = TagPermissionMap(permission_id=permission_id, tag_id=tag_id, argument=argument)
+    mapping.add(session)
+
+    Counter.incr(session, "updates")
+
+    session.commit()
```
herb: consider catching the `IntegrityError` and handling the rollback here instead of the handler. keeps to direct database code confined to this module.

grouper/public_key.py
```diff
@@ -79,8 +89,55 @@ def delete_public_key(session, user_id, key_id):
         KeyNotFound if specified key wasn't found
```
herb: nit: update docstring

tests/test_tags.py
```diff
@@ -0,0 +1,391 @@
...
+
+    assert tag.description == "Don't tag me bro", "The description should have been updated"
+
+
+@pytest.mark.gen_test
+def test_permissions(users, http_client, base_url, session):
+
+    user = session.query(User).filter_by(username="testuser@a.co").scalar()
+
+    perm = Permission(name=TAG_EDIT, description="Why is this not nullable?")
```
herb: prob cause we want to know what a permission's intended use case is. the description can be an empty string fwiw.

herb:
can you talk about how you intend to implement the API side of things?

grouper/models/public_key.py
Acidity: I talked with @lfaraone about this, and we weren't sure on the best way to approach this. The biggest problem is the lack of a clean way (that I'm familiar with at least) to do the equivalent functionality in Jinja. Additionally, because the use case for this (https://github.com/dropbox/grouper/pull/208/files#diff-64a1bd8aadee752b60baa65ad62066a8R31) requires that it be run for every key that a user has, if we wanted to move this to another part of the code and run it in say the handler to pass into the template, we'd have to have some way of coupling the output with the PermissionKey objects themselves. Do you have a particular solution in mind?

grouper/public_key.py
Acidity: What's wrong with this docstring? That function throws KeyNotFound by virtue of calling get_public_key and not catching KeyNotFound from it.

grouper/models/public_key.py
Acidity: Was this intended for this line? If so, see my comment above re: best place to put it.

Acidity:
For this change, I added a tags field to the public key data in grouper/graph, as well as the permission name and argument for each permission that public keys have. This gets reflected in (at least) the user api. If adding or modifying the other api endpoints is desirable, I'd prefer to do so in a later pull request because this one is getting rather long as it is.

Acidity:
Updated

grouper/models/public_key.py
herb: sorry, can you clarify for me how an equivalence function plays into the use case? iirc the use case you pointed out just display the results in a modal?

we do `perm.argument if perm.argument else "unargumented"` in the template with the `default` function, i.e. `perm.argument|default("unargumented", True)`. is that what you're trying to accomplish?

grouper/models/public_key.py
herb: no. it was meant for the entire block. sigh. i hate github diffs.

this applies to the functions `tags` (and transitively on `my_permissions`) which creates a dependency on the `PublicKeyTagMap` model.

grouper/models/public_key.py
Acidity: The issue is that we want to apply the formatting to a list of things (the key's permissions), and then provide the resultant formatted list back to another function. I.e. ["{} ({})".format(perm.name, perm.argument) for perm in key.permissions()].
Unfortunately, Jinja doesn't allow list comprehensions, and I was unable to find some way to unfurl the list comprehension into a traditional for loop in the template that would allow for appending to a list. One possible solution might be to try just appending to the HTML entity that we ultimately set to as the list, but I don't know that that'd be a readability/usability improvement.

grouper/models/public_key.py
Acidity: Ahh I see. I was mainly trying to keep it similar to the existing interfaces provided by the User and Group class, but if you think it's preferable to split it out to get_tags_for_pub_key and get_permissions_for_pub_key functions I can definitely do that.

grouper/public_key.py
herb: sorry, it would't let me common on the right line (81). just update to reflect that it removes the key as well as associated tags.

grouper/models/public_key.py
herb: ah, gotcha. it feels slightly cleaner to put this in the handler and post process the template context and decorate it with this string, i.e. monkey-patch an attribute with the result of this list comprehension. that way you're not passing it around to thing that don't care about presentation layer stuff and it's calculated close to where it's used.

Acidity:
Updated the code, should resolve all of the previously brought up issues

lfaraone:
can you push a rebased branch?

Acidity:
@lfaraone Done

grouper/permissions.py
```diff
@@ -61,6 +62,39 @@ def grant_permission(session, group_id, permission_id, argument=''):
     session.commit()
 
 
+def grant_permission_to_tag(session, tag_id, permission_id, argument=''):
+    # type: Session, int, int, str -> None
```
herb: are these `type:` declarations meant to conform to [PEP 484](https://www.python.org/dev/peps/pep-0484/#id44)? if i'm not mistaken these will all fail things like `mypy` because
1. they reference objects that are not imported, `Session` in this case and
2. the formatting is off, should be something like `# type: (Session, int, int, str) -> None`

I'm happy you're adding these but if can run these through `mypy` to make sure they work it'd help when we add `mypy` to CI.

herb:
> For this change, I added a tags field to the public key data in grouper/graph, as well as the permission name and argument for each permission that public keys have. This gets reflected in (at least) the user api. If adding or modifying the other api endpoints is desirable, I'd prefer to do so in a later pull request because this one is getting rather long as it is.

gotcha. do you mind adding a test that verifies what you want appears in the api results? or did i miss that implementation?

Acidity:
I fixed the format on the type annotations; do we want to import the annotated types even though they're not used just for the mypy support? We'd have to manually flag them as # noqa so flake doesn't alert on them, and that could lead to importing unused classes in the future.

herb:
thanks for adding that test. do you mind moving it to it's own module? in retrospect, `test_api_handlers` is a poor name that promotes monolithic tests piles. :/

agreed on `# noqa` not being great for imported stuff. ideally we can teach `flake8` and `mypy` to like each other. unclear how much's built in right. i'm ok not importing them for now as we don't have many cases where the references isn't obvious.

Acidity:
Test moved :)

PR #209, opened 2016-06-05
Add support for storing password hashes for users.

Acidity:
Allows users to add named passwords to their accounts. The passwords
are hashed using SHA512, and salted using a unique 160-bit salt that
is randomly generated for each password. The hashes and their salts
are exposed in the users API endpoint, so high risk passwords should
not be stored in Grouper due to the lack of authentication on the
user endpoint.

### DIFF
--- a/grouper/fe/forms.py
+++ b/grouper/fe/forms.py
@@ -3,14 +3,8 @@ import re
 
 import sshpubkey
 from wtforms import (
-        BooleanField,
-        HiddenField,
-        IntegerField,
-        PasswordField,
-        SelectField,
-        StringField,
-        TextAreaField,
-        validators,
+        BooleanField, HiddenField, IntegerField, PasswordField,
+        SelectField, StringField, TextAreaField, validators
         )
 from wtforms.validators import ValidationError
 from wtforms_tornado import Form
--- a/grouper/fe/forms.py
+++ b/grouper/fe/forms.py
@@ -6,6 +6,7 @@ from wtforms import (
         BooleanField,
         HiddenField,
         IntegerField,
+        PasswordField,
         SelectField,
         StringField,
         TextAreaField,
@@ -322,3 +323,12 @@ class PublicKeyAddTagForm(Form):
     tagname = SelectField("Tag", [
         validators.DataRequired(),
     ], choices=[["", "(select one)"]], default="")
+
+
+class UserPasswordForm(Form):
+    name = StringField("Password name", [
+        validators.DataRequired(),
+    ])
+    password = PasswordField("Password", [
+        validators.DataRequired(),
+    ])
--- a/grouper/graph.py
+++ b/grouper/graph.py
@@ -15,6 +15,7 @@ from grouper.models.permission_map import PermissionMap
 from grouper.models.public_key import PublicKey
 from grouper.models.user import User
 from grouper.models.user_metadata import UserMetadata
+from grouper.models.user_password import UserPassword
 from grouper.public_key import get_public_key_permissions, get_public_key_tags
 from grouper.util import singleton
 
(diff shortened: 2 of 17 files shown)


grouper/fe/forms.py
```diff
@@ -264,3 +265,12 @@ class UserTokenForm(Form):
     name = StringField("Token name", [
         validators.DataRequired(),
     ])
+
+
+class UserPasswordForm(Form):
+    name = StringField("Password name", [
```
lfaraone: Hmm. @rra should we make this user-defined, only have one, or constrain the set of possible password names?

(I assume this is intended to be used in the context of "service user is setting a password for")

grouper/user_password.py
```diff
@@ -0,0 +1,48 @@
...
+
+    Args:
+        session(grouper.models.base.session.Session): database session
+        password_name: the name of the password to delete
+        user_id: the user whose password is being deleted
+    """
+    p = session.query(UserPassword).filter_by(name=password_name, user_id=user_id).scalar()
+    if not p:
+        raise PasswordDoesNotExist()
+    p.delete(session)
```
lfaraone: In general, I try to avoid deleting objects from the database for audit/recoverability reasons. But @herb feel free to disagree. 

Perhaps instead mark the password as inactive and rename it?

grouper/models/user_password.py
```diff
@@ -0,0 +1,84 @@
...
+    @property
+    def password(self):
+        return self._hashed_secret
+
+    @password.setter
+    def password(self, new_password):
+        self.salt = _make_salt()
+        self._hashed_secret = hashlib.sha512(new_password + self.salt).hexdigest()
+
+    def check_password(self, password_to_check):
```
lfaraone: We could be fancy and override `__eq__`, but that's probably just needlessly magical.

tests/test_passwords.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_passwords(session, users):
+    user = users['zorkian@a.co']
+
+    add_new_user_password(session, "test", "test_password_please_ignore", user.id)
+    assert len(user.my_passwords()) == 1, "The user should only have a single password"
+    password = user.my_passwords()[0]
+    assert password.name == "test", "The password should have the name we gave it"
+    assert password.password != "test_password_please_ignore", "The password should not be what is passed in"
```
lfaraone: if you refactor this out to a constant, we don't have to worry about typos in your checks.

grouper/fe/forms.py
rra: We have historically had a single password hash for a given user.  It would be better from a security standpoint to let the user create multiple hashes, one for each service, but in practice the only service that still uses this is TACACS.  So I'm not sure it's worth spending too much time or effort on.

rra: Separate passwords per application are definitely better, but we don't use this much, so I'm not sure how much effort it's worth putting into this.

grouper/graph.py
```diff
@@ -141,25 +142,34 @@ def _get_user_metadata(session):
...
         '''
-        users = session.query(User)
 
-        public_keys = {}
-        for key in session.query(PublicKey):
-            if key.user_id not in public_keys:
-                public_keys[key.user_id] = []
-            public_keys[key.user_id].append(key)
+        def user_indexify(data):
+            ret = {}
```
herb: `ret = defaultdict(list)` will save you a few lines. you could also replace the entire block with a `reduce` which is arguably less readable.

grouper/models/user_password.py
```diff
@@ -0,0 +1,84 @@
+from datetime import datetime
+import hashlib
+import hmac
+import os
+
+from grouper.models.base.model_base import Model
```
herb: nit: this should be a 3rd, separate group of imports

tests/test_passwords.py
```diff
@@ -0,0 +1,133 @@
+import json
+from urllib import urlencode
+
+import pytest
+from tornado.httpclient import HTTPError
+
+from fixtures import standard_graph, graph, users, groups, session, permissions  # noqa
+from fixtures import fe_app as app  # noqa
```
herb: can you add a api test as well to confirm your hashes are making it out correctly on the other side?

herb:
one other note: we call things 'password' all over the place but they're technically hashes. might be worth naming things more explicitly to avoid that confusion.

otherwise, awesome. :)

grouper/user_password.py
Acidity: Well at the moment we delete public keys, so I was modelling after that. If you still want me to switch it to disabling passwords I can do that though.

grouper/fe/forms.py
Acidity: Is there a final decision on whether we want to restrict user input or just make them type the right name (if applications restrict what they use based on the password name)?

Acidity: Talked with Luke, he thinks we should stick with freeform entry

grouper/user_password.py
Acidity: Talked with Luke, he's cool with keeping it with deleting passwords

grouper/models/user_password.py
Acidity: I'm leaning towards that being needlessly magical

lfaraone:
Always be rebasing

Acidity:
Rebased

lfaraone:
one more rebase at your convenience :) (no rush)

Acidity:
Rebased ^_^

PR #211, opened 2016-06-06
Add the ability for users to select which shell they prefer.

Acidity:


grouper/constants.py
```diff
@@ -71,3 +71,6 @@
 
 # Maximum length a name can be. This applies to user names and permission arguments.
 MAX_NAME_LENGTH = 128
+
+# Grouper used UserMetadata data_keys
+SHELL_MD_KEY = "shell"
```
herb: Got confused when first reading this variable name. Can we be more explicit? Something like `USER_METADATA_SHELL_KEY`. other keys will keep the same pre- and suf- fix.

grouper/fe/settings.py
```diff
@@ -25,6 +25,7 @@ def override_timezone(self, value):
     "permission_request_text_help": None,
     "port": 8989,
     "refresh_interval": 60,
+    "shell": [["/bin/bash", "bash"], ["/bin/zsh", "zsh"]],
```
herb: can we make this default similar to the dev config default? so `[["/bin/false", ""]]` seems similar enough for me.

tests/test_shells.py
```diff
@@ -0,0 +1,52 @@
+import json
+from urllib import urlencode
+from mock import patch
+
+import pytest
+from tornado.httpclient import HTTPError
+
+from fixtures import standard_graph, graph, users, groups, session, permissions  # noqa
+from fixtures import fe_app as app  # noqa
```
herb: can we add a test to ensure this shows up in the API correctly?

grouper/fe/settings.py
Acidity: Originally it was, but I had to change this so we could have these set for the tests accompanying this change. I tried to patch the fe.settings object for the test, but the interaction between mock.patch and pytest.mark.gen_test led to that not working. If you know of another way of working around that issue, please let me know :)

Acidity:
Updated

grouper/fe/settings.py
herb: interesting. i want to say i've had `patch` work with these requests but haven't confirmed recently. can you recall how you specifically tried to patch settings?

Acidity:
```
@patch("grouper.fe.settings.settings")
@pytest.mark.gen_test
def test_shell(settings, session, users, http_client, base_url):
    settings.shell = [["/bin/bash", "bash"], ["/bin/zsh", "zsh"]]
    user = users['zorkian@a.co']
    assert not user.get_metadata(USER_METADATA_SHELL_KEY)
...
```

herb: i see. you can use patch in a context instead which works around this. also, fwiw, you should [patch at the point of import](http://www.voidspace.org.uk/python/mock/patch.html#where-to-patch) vs where it's defined.

so something like
```
def test_shell(...):
  with('grouper.fe.handlers.user_shell.settings') as mock_settings:
    mock_settings.settings.shell = [['blha', 'blah'], ['foo', 'foo']]
    [...]
```

tests/test_shells.py
```diff
@@ -0,0 +1,53 @@
...
+    user = User.get(session, name=user.username)
+    fe_url = url(base_url, '/users/{}/shell'.format(user.username))
+    resp = yield http_client.fetch(fe_url, method="POST",
+            body=urlencode({'shell': "/bin/bash"}),
+            headers={'X-Grouper-User': user.username})
+    assert resp.code == 200
+
+    user = User.get(session, name=user.username)
+
+    assert user.get_metadata(USER_METADATA_SHELL_KEY) is not None, "The user should have shell metadata"
```
herb: also i noticed these lines exceeded 100 characters. guess we should `flake8` the test directory as well. :)

mind wrapping these?

Acidity:
Fixed :)

PR #212, opened 2016-06-07
Change import statement ordering to match the Google Python Style Guide

Acidity:
Also adds flake8-import-ordering to requirements-dev.txt to prevent
imports from being improperly ordered in the future.


Acidity:
https://travis-ci.org/dropbox/grouper/builds/135744140
Travis now fails tests if you have incorrect import styling too ^_^

PR #213, opened 2016-06-09
Pull User out of model_soup

Acidity:


grouper/fe/handlers/group_edit_member.py
```diff
@@ -3,7 +3,8 @@
 from grouper.audit import assert_can_join, UserNotAuditor
 from grouper.fe.forms import GroupEditMemberForm
 from grouper.fe.util import Alert, GrouperHandler
-from grouper.model_soup import Group, GroupEdge, OBJ_TYPES, User
+from grouper.model_soup import Group, GroupEdge, User
```
lfaraone: This should import `User` from `grouper.models.user` instead.

lfaraone:
Thanks for this! I'll take a deeper look tomorrow, and you don't need to make any changes until then, but I'm wary of the number of inline imports this adds. Inline imports are generally a code smell, and can make it a lot harder to reason about the code.

We have them already in a few places, but they should basically always be considered to be prepended with an implicit `TODO: refactor me`

grouper/ctl/user.py
```diff
@@ -4,7 +4,8 @@
 from grouper.ctl.util import ensure_valid_username, make_session
 from grouper.model_soup import User
```
herb: is there a reason we can't renaming these existing imports to directly import the new model?

grouper/models/user.py
```diff
@@ -0,0 +1,82 @@
...
+from sqlalchemy.ext.hybrid import hybrid_property
+from sqlalchemy.orm import relationship
+
+from grouper.constants import GROUP_ADMIN, MAX_NAME_LENGTH, PERMISSION_ADMIN, USER_ADMIN
+from grouper.models.audit_log import AuditLog
+from grouper.models.base.model_base import Model
+from grouper.models.comment import CommentObjectMixin
+from grouper.models.counter import Counter
+from grouper.models.user_token import UserToken
+from grouper.plugin import get_plugins
```
herb: it'd be nice to avoid these cross model dependencies if possible. mostly that looks like pull the function into a grouper level module.

In the case of `UserToken` that might just be changing the reference to a string.
```
tokens = relationship('user_token', back_populates="user")
```

grouper/user.py
```diff
@@ -57,3 +69,232 @@ def get_all_enabled_users(session):
...
+
+def user_role(user, members):
+    role_index = user_role_index(user, members)
+    if not role_index:
+        return None
+    else:
+        return GROUP_EDGE_ROLES[role_index]
+
+
+def user_has_permission(session, user, permission, argument=None):
```
herb: these new user + permission methods seem to make a logical group. thoughts on moving then into their own module `user_permissions` or some such?

grouper/user_metadata.py
```diff
@@ -39,28 +52,25 @@ def set_user_metadata(session, user_id, data_key, data_value):
     """
     assert re.match(PERMISSION_VALIDATION, data_key), 'proposed metadata key is valid'
 
-    user_md = None
-    for try_user_md in get_user_metadata(session, user_id):
-        if try_user_md.data_key == data_key:
-            user_md = try_user_md
-            break
+    user_md = get_user_metadata_by_key(session, user_id, data_key)
+    print(user_md)
```
herb: stray print

grouper/models/user.py
Acidity: So at the moment I've left 3 of the models imports as is: Model, which must  be imported because it's the base model, CommentObjectMixin, because User must inherit from it, and Counter, because that's needed in User.add which is a method inherited from Model, so moving it out to another module would require changing that method for every other model in Grouper.

Acidity:
Updated to include @herb's suggestions

grouper/models/user.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @property
+    def group_admin(self):
+        from grouper.user_permissions import user_has_permission
+        return user_has_permission(self.session, self, GROUP_ADMIN)
+
+    @property
+    def permission_admin(self):
+        from grouper.user_permissions import user_has_permission
+        return user_has_permission(self.session, self, PERMISSION_ADMIN)
```
herb: sorry, i didn't see this earlier. this isn't great cause we're just masking the circular dependency. to break it we'd need to change all calls to these properties to call the function directly. is there a specific reason you didn't that?

grouper/models/user.py
herb: Makes sense. Yeah, I kind of ignore `Counter` through all my refactoring cause it was somewhat harmless. It doesn't really call out to anything. I suspect we'll want to move `CommentObjectMixin` to it's own thing at some point but let's leave it until it becomes a problem. :)

grouper/models/user.py
Acidity: The only reason I didn't was because I felt these would be awkward as their own functions; but thinking on it more it's probably better to replace User.user_admin with just a direct call to user_has_permission(..., USER_ADMIN) at the call site anyways.

herb: you can also create wrappers in `user_permissions` so call sites don't need to import constants unnecessarily.

Acidity:
Move the *_admin properties

grouper/models/user.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def is_member(self, members):
+        return ("User", self.name) in members
+
+    def set_metadata(self, key, value):
+        from grouper.user_metadata import set_user_metadata
+
+        logging.warning("User.set_metadata is deprecated."
+            "Please switch to using grouper.user_metadata.set_user_metadata")
+        set_user_metadata(self.session, self.id, key, value)
```
herb: i think this guy is the same deal?

herb:
thanks for doing this!

PR #214, opened 2016-06-10
Add support for service accounts

Acidity:
Note: This was branched off of #213, so it shouldn't be reviewed/pulled until 213 is merged into master.

### DIFF
--- a/grouper/constants.py
+++ b/grouper/constants.py
@@ -10,7 +10,7 @@ NAME_VALIDATION = _NAME_VALIDATION.format("name")
 NAME2_VALIDATION = _NAME_VALIDATION.format("name2")
 
 # This regex is specifically to validate usernames.
-USERNAME_VALIDATION = r"(?P<name>\w+@\w+\.\w+)"
+USERNAME_VALIDATION = r"(?P<name>\w+@\w+[\.\w]+)"
 
 # UserToken validators
 TOKEN_SECRET_VALIDATION = r"(?P<token_secret>[a-f0-9]{40})"
--- a/grouper/fe/forms.py
+++ b/grouper/fe/forms.py
@@ -16,7 +16,7 @@ from .. import model_soup
 GROUP_CANJOIN_CHOICES = [
     ("canjoin", "Anyone"), ("canask", "Must Ask"),
     ("nobody", "Nobody"),
-    ]
+]
 
 
 class ValidateRegex(object):
--- a/grouper/fe/forms.py
+++ b/grouper/fe/forms.py
@@ -326,3 +326,17 @@ class UserPasswordForm(Form):
     password = PasswordField("Password", [
         validators.DataRequired(),
     ])
+
+
+class ServiceAccountCreateForm(Form):
+    name = StringField("Name", [
+        validators.Length(min=3, max=32),
+        validators.DataRequired(),
+        ValidateRegex(constants.NAME_VALIDATION),
+        ValidateRegex(constants.USERNAME_VALIDATION),
+    ])
+    description = TextAreaField("Description")
+    canjoin = SelectField("Who Can Join?", choices=[
+        ("canjoin", "Anyone"), ("canask", "Must Ask"),
+        ("nobody", "Nobody"),
+    ], default="canask")
--- a/grouper/fe/forms.py
+++ b/grouper/fe/forms.py
@@ -13,6 +13,12 @@ from .. import constants
 from .. import model_soup
 
 
+GROUP_CANJOIN_CHOICES = [
+    ("canjoin", "Anyone"), ("canask", "Must Ask"),
+    ("nobody", "Nobody"),
+    ]
+
+
 class ValidateRegex(object):
     def __init__(self, regex):
         # We assume any regex passed to us does not have line anchors.
@@ -99,10 +105,8 @@ class GroupCreateForm(Form):
         ValidateRegex(constants.NAME_VALIDATION),
     ])
     description = TextAreaField("Description")
-    canjoin = SelectField("Who Can Join?", choices=[
-        ("canjoin", "Anyone"), ("canask", "Must Ask"),
-        ("nobody", "Nobody"),
-    ], default="canask")
+    canjoin = SelectField("Who Can Join?", choices=GROUP_CANJOIN_CHOICES,
+        default="canask")
     auto_expire = DaysTimeDeltaField("Default Expiration (Days)")
 
 
@@ -113,10 +117,8 @@ class GroupEditForm(Form):
         ValidateRegex(constants.NAME_VALIDATION),
     ])
     description = TextAreaField("Description")
-    canjoin = SelectField("Who Can Join?", choices=[
-        ("canjoin", "Anyone"), ("canask", "Must Ask"),
-        ("nobody", "Nobody"),
-    ], default="canask")
+    canjoin = SelectField("Who Can Join?", choices=GROUP_CANJOIN_CHOICES,
+        default="canask")
     auto_expire = DaysTimeDeltaField("Default Expiration (Days)")
 
 
@@ -336,7 +338,5 @@ class ServiceAccountCreateForm(Form):
         ValidateRegex(constants.USERNAME_VALIDATION),
     ])
     description = TextAreaField("Description")
-    canjoin = SelectField("Who Can Join?", choices=[
-        ("canjoin", "Anyone"), ("canask", "Must Ask"),
-        ("nobody", "Nobody"),
-    ], default="canask")
+    canjoin = SelectField("Who Can Join?", choices=GROUP_CANJOIN_CHOICES,
+    default="canask")
--- a/grouper/fe/handlers/groups_view.py
+++ b/grouper/fe/handlers/groups_view.py
@@ -26,6 +26,7 @@ class GroupsView(GrouperHandler):
         else:
             groups = self.graph.get_groups(audited=False)
             directly_audited_groups = set()
+        groups = [group for group in groups if not group.service_account]
         total = len(groups)
         groups = groups[offset:offset + limit]
 
--- a/grouper/fe/handlers/public_key_add.py
+++ b/grouper/fe/handlers/public_key_add.py
@@ -5,17 +5,28 @@ from grouper.fe.settings import settings
 from grouper.fe.util import GrouperHandler
 from grouper.models.audit_log import AuditLog
 from grouper.models.user import User
+from grouper.service_account import can_manage_service_account
 from grouper.user_permissions import user_is_user_admin
 
 
 class PublicKeyAdd(GrouperHandler):
+
+    @staticmethod
+    def check_access(session, actor, target):
+        if actor.name == target.name:
+            return True
+
+        if not target.role_user:
+            return False
+
+        return can_manage_service_account(session, actor, tuser=target)
+
     def get(self, user_id=None, name=None):
         user = User.get(self.session, user_id, name)
         if not user:
             return self.notfound()
 
-        if ((user.name != self.current_user.name) and
-                not user_is_user_admin(self.session, self.current_user)):
+        if not self.check_access(self.session, self.current_user, user):
             return self.forbidden()
 
         self.render("public-key-add.html", form=PublicKeyForm(), user=user)
@@ -25,8 +36,7 @@ class PublicKeyAdd(GrouperHandler):
         if not user:
             return self.notfound()
 
-        if ((user.name != self.current_user.name) and
-                not user_is_user_admin(self.session, self.current_user)):
+        if not self.check_access(self.session, self.current_user, user):
             return self.forbidden()
 
         form = PublicKeyForm(self.request.arguments)
--- a/grouper/fe/handlers/public_key_add.py
+++ b/grouper/fe/handlers/public_key_add.py
@@ -6,7 +6,6 @@ from grouper.fe.util import GrouperHandler
 from grouper.models.audit_log import AuditLog
 from grouper.models.user import User
 from grouper.service_account import can_manage_service_account
-from grouper.user_permissions import user_is_user_admin
 
 
 class PublicKeyAdd(GrouperHandler):
--- a/grouper/fe/handlers/public_key_add.py
+++ b/grouper/fe/handlers/public_key_add.py
@@ -12,13 +12,8 @@ class PublicKeyAdd(GrouperHandler):
 
     @staticmethod
     def check_access(session, actor, target):
-        if actor.name == target.name:
-            return True
-
-        if not target.role_user:
-            return False
-
-        return can_manage_service_account(session, actor, tuser=target)
+        return (actor.name == target.name or
+            (target.role_user and can_manage_service_account(session, actor, tuser=target)))
 
     def get(self, user_id=None, name=None):
         user = User.get(self.session, user_id, name)
--- a/grouper/fe/handlers/public_key_delete.py
+++ b/grouper/fe/handlers/public_key_delete.py
@@ -4,17 +4,28 @@ from grouper.fe.util import GrouperHandler
 from grouper.models.audit_log import AuditLog
 from grouper.models.user import User
 from grouper.public_key import delete_public_key, get_public_key, KeyNotFound
+from grouper.service_account import can_manage_service_account
 from grouper.user_permissions import user_is_user_admin
 
 
 class PublicKeyDelete(GrouperHandler):
+
+    @staticmethod
+    def check_access(session, actor, target):
+        if actor.name == target.name:
+            return True
+
+        if not target.role_user:
+            return False
+
+        return can_manage_service_account(session, actor, tuser=target)
+
     def get(self, user_id=None, name=None, key_id=None):
         user = User.get(self.session, user_id, name)
         if not user:
             return self.notfound()
 
-        if ((user.name != self.current_user.name) and
-                not user_is_user_admin(self.session, self.current_user)):
+        if not self.check_access(self.session, self.current_user, user):
             return self.forbidden()
 
         try:
@@ -29,8 +40,7 @@ class PublicKeyDelete(GrouperHandler):
         if not user:
             return self.notfound()
 
-        if ((user.name != self.current_user.name) and
-                not user_is_user_admin(self.session, self.current_user)):
+        if not self.check_access(self.session, self.current_user, user):
             return self.forbidden()
 
         try:
--- a/grouper/fe/handlers/public_key_delete.py
+++ b/grouper/fe/handlers/public_key_delete.py
@@ -5,7 +5,6 @@ from grouper.models.audit_log import AuditLog
 from grouper.models.user import User
 from grouper.public_key import delete_public_key, get_public_key, KeyNotFound
 from grouper.service_account import can_manage_service_account
-from grouper.user_permissions import user_is_user_admin
 
 
 class PublicKeyDelete(GrouperHandler):
--- a/grouper/fe/handlers/public_key_delete.py
+++ b/grouper/fe/handlers/public_key_delete.py
@@ -13,13 +13,6 @@ class PublicKeyDelete(GrouperHandler):
     def check_access(session, actor, target):
         return (actor.name == target.name or
             (target.role_user and can_manage_service_account(session, actor, tuser=target)))
-        if actor.name == target.name:
-            return True
-
-        if not target.role_user:
-            return False
-
-        return can_manage_service_account(session, actor, tuser=target)
 
     def get(self, user_id=None, name=None, key_id=None):
         user = User.get(self.session, user_id, name)
--- a/grouper/fe/handlers/public_key_delete.py
+++ b/grouper/fe/handlers/public_key_delete.py
@@ -11,6 +11,8 @@ class PublicKeyDelete(GrouperHandler):
 
     @staticmethod
     def check_access(session, actor, target):
+        return (actor.name == target.name or
+            (target.role_user and can_manage_service_account(session, actor, tuser=target)))
         if actor.name == target.name:
             return True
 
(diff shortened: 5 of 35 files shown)


lfaraone:
rebase when you get a chance @Acidity

grouper/constants.py
```diff
@@ -10,7 +10,7 @@
 NAME2_VALIDATION = _NAME_VALIDATION.format("name2")
 
 # This regex is specifically to validate usernames.
-USERNAME_VALIDATION = r"(?P<name>\w+@\w+\.\w+)"
+USERNAME_VALIDATION = r"(?P<name>\w+@\w+[\.\w]+)"
```
lfaraone: why was this changed btw?

grouper/fe/handlers/service_account_create.py
```diff
@@ -0,0 +1,112 @@
+from datetime import datetime
+
+from sqlalchemy.exc import IntegrityError
+
+from grouper import group as group_biz
```
lfaraone: What's the `_biz` suffix? Business logic?

grouper/constants.py
Acidity: Yes

grouper/fe/handlers/service_account_create.py
```diff
@@ -0,0 +1,112 @@
...
+from grouper.audit import assert_can_join, UserNotAuditor
+from grouper.email_util import send_email
+from grouper.fe.forms import ServiceAccountCreateForm
+from grouper.fe.settings import settings
+from grouper.fe.util import Alert, GrouperHandler
+from grouper.model_soup import (
+        APPROVER_ROLE_INDICIES,
+        Group,
+        GROUP_EDGE_ROLES,
+        GROUP_JOIN_CHOICES,
```
lfaraone: When sorted with vim's `sort` command, I get:
```
        APPROVER_ROLE_INDICIES,
        GROUP_EDGE_ROLES,
        GROUP_JOIN_CHOICES,
        Group,
```

grouper/fe/forms.py
```diff
@@ -291,3 +291,17 @@ class UserShellForm(Form):
...
+class ServiceAccountCreateForm(Form):
+    name = StringField("Name", [
+        validators.Length(min=3, max=32),
+        validators.DataRequired(),
+        ValidateRegex(constants.NAME_VALIDATION),
+        ValidateRegex(constants.USERNAME_VALIDATION),
+    ])
+    description = TextAreaField("Description")
+    canjoin = SelectField("Who Can Join?", choices=[
+        ("canjoin", "Anyone"), ("canask", "Must Ask"),
```
lfaraone: Can we factor these out into a constant / enum?

grouper/fe/handlers/service_account_create.py
Acidity: Within each grouping, imports should be sorted lexicographically, ignoring case, according to each module's full package path.

grouper/fe/templates/users.html
```diff
@@ -45,3 +48,4 @@
         </table>
     </div>
 {% endblock %}
+
```
lfaraone: Stray space

grouper/fe/handlers/service_account_create.py
Acidity: Looks like. Looks like I forgot to change that after copying group_join.py

Acidity: Ah, looking at it more, yeah it's used in one of the unnecessary methods that I later deleted from that class (which also leads to that import being reviewed).

grouper/fe/forms.py
```diff
@@ -18,6 +18,12 @@
 from .. import model_soup
 
 
+GROUP_CANJOIN_CHOICES = [
+    ("canjoin", "Anyone"), ("canask", "Must Ask"),
+    ("nobody", "Nobody"),
+    ]
```
lfaraone: style nit: the closing `]` should be unindented. there's some inconsistency in this file already, so I'll just clean it all up later.

grouper/fe/handlers/service_account_create.py
```diff
@@ -0,0 +1,112 @@
...
+                alerts=self.get_form_alerts(form.errors)
+            )
+
+        group.add_member(self.current_user, self.current_user, "GroupCreator",
+            "actioned", None, "np-owner")
+        group.add_member(self.current_user, user, "Service Account",
+            "actioned", None, "member")
+        self.session.commit()
+
+        AuditLog.log(self.session, self.current_user.id, 'create_group',
```
lfaraone: @herb thoughts on using a new log action here?

grouper/fe/handlers/service_account_create.py
```diff
@@ -0,0 +1,112 @@
...
+                     'Created new service account.', on_group_id=group.id)
+
+        return self.redirect("/groups/{}?refresh=yes".format(group.name))
+
+    def _get_member(self, member_choice):
```
lfaraone: Could we add some comments explaining what these methods are for and what they do?

grouper/fe/templates/service-account-create.html
```diff
@@ -0,0 +1,32 @@
...
+{% block heading %}
+    <a href="/users">Users</a>
+{% endblock %}
+
+{% block subheading %}
+    Create Service Account
+{% endblock %}
+
+{% block content %}
+<div class="col-md-6 col-md-offset-3">
```
lfaraone: Should be indented.

grouper/fe/handlers/service_account_create.py
Acidity: They don't do anything; they were mistakenly left in and then deleted in a later commit.

grouper/fe/handlers/service_account_view.py
```diff
@@ -0,0 +1,90 @@
...
+        OWNER_ROLE_INDICES)
+from grouper.permissions import get_requests_by_owner
+from grouper.public_key import get_public_keys_of_user
+from grouper.user import user_open_audits, user_requests_aggregate, user_grantable_permissions, user_role, user_role_index
+from grouper.user_metadata import get_user_metadata_by_key
+
+
+class ServiceAccountView(GrouperHandler):
+    def get(self, user_id=None, name=None):
+        print("HELOO")
```
lfaraone: HAI!

stray print?

grouper/fe/handlers/service_account_view.py
```diff
@@ -0,0 +1,90 @@
...
+        self.handle_refresh()
+        user = User.get(self.session, user_id, name)
+        if user_id is not None:
```
lfaraone: We get the `user` but never access its properties before rewriting it?

grouper/fe/handlers/service_account_view.py
```diff
@@ -0,0 +1,90 @@
...
+            user = self.session.query(User).filter_by(id=user_id).scalar()
+        else:
+            user = self.session.query(User).filter_by(username=name).scalar()
+
+        if not user:
+            return self.notfound()
+
+        if not user.role_user:
```
lfaraone: combine this check and the one directly above into one line

grouper/fe/handlers/service_account_view.py
```diff
@@ -0,0 +1,90 @@
...
+            return self.notfound()
+
+        can_control = user.name == self.current_user.name or self.current_user.user_admin
+        can_disable = UserDisable.check_access(self.session, self.current_user, user)
+        can_enable = UserEnable.check_access(self.session, self.current_user, user)
+
+        if user.id == self.current_user.id:
```
lfaraone: fwiw this should always be false, since a service account shouldn't ever hit `grouper-fe`.

grouper/fe/routes.py
```diff
@@ -71,6 +72,7 @@
     (r"/users", UsersView),
     (r"/users/public-keys", UsersPublicKey),
     (r"/users/tokens", UsersUserTokens),
+    (r"/users/service_create", ServiceAccountCreate),
```
lfaraone: consider making this consistent with the other service account paths.

also, I like `/service` as a prefix a bit more than the inscrutable `/svc`

grouper/fe/templates/service.html
```diff
@@ -0,0 +1,318 @@
...
+{% block heading %}
+    <a href="/svc">Service Accounts</a>
+{% endblock %}
+
+{% block subheading %}
+    {{ user.username }}
+    {% if user.role_user %}<small>(role)</small>{% endif %}
+    {% if not user.enabled %}<small>(disabled)</small>{% endif %}
+
+    {% if group.audit and not group.audit.complete %}
```
lfaraone: Hm, hadn't thought about how this'll interact with audits… is an audit of a SA really meaningful? I could imagine wanting to have a SA audit process, but I think for now we'll want to exclude SA groups from audits

grouper/fe/templates/service.html
```diff
@@ -0,0 +1,318 @@
...
+            {% if audited %}
+                <span class="label label-warning pull-right">
+                    <i class="fa fa-gavel"></i> <span {{ help_for("audited_group") }}>
+                        Auditing Enabled
+                    </span>
+                </span>
+            {% endif %}
+            {% if group.canjoin == "canjoin" %}
+                <span class="label label-success pull-right">
+                    <i class="fa fa-group"></i> Open Membership
```
lfaraone: there shouldn't be any SA groups that anyone can join without approval.

grouper/fe/templates/service.html
```diff
@@ -0,0 +1,318 @@
+{% extends "base.html" %}
```
lfaraone: In general, any case where we're blanketly-duplicating template logic used elsewhere with no changes, we should turn that into a macro.

grouper/fe/handlers/user_token_add.py
```diff
@@ -8,19 +8,29 @@
...
+
+    @staticmethod
+    def check_access(session, actor, target):
+        if actor.name == target.name:
+            return True
+
+        if not target.role_user:
+            return False
+
+        return can_manage_service_account(session, actor, tuser=target)
```
lfaraone: I'd consolidate these onto one chained conditional.

grouper/fe/handlers/user_token_disable.py
```diff
@@ -7,13 +8,26 @@
...
+        if actor.name == target.name:
+            return True
+
+        if user_has_permission(session, actor, USER_ADMIN):
+            return True
+
+        if not target.role_user:
+            return False
+
+        return can_manage_service_account(session, actor, tuser=target)
```
lfaraone: Same as for `UserTokenEnable.check_access`

grouper/fe/handlers/user_disable.py
```diff
@@ -25,7 +25,11 @@ def post(self, user_id=None, name=None):
         if not self.check_access(self.session, self.current_user, user):
             return self.forbidden()
 
-        disable_user(self.session, user)
+        if user.role_user:
+            disable_service_account(self.session, user=user)
```
lfaraone: it'd be great if the API for `disable_service_account` was the same as that for `disable_user`

lfaraone:
let's also make sure `grouper-ctl`'s user management commands DTRT here.

grouper/fe/handlers/service_account_create.py
```diff
@@ -33,6 +33,13 @@ def post(self):
                 alerts=self.get_form_alerts(form.errors)
             )
 
+        if form.data["name"].split("@")[-1] != settings.service_account_email_domain:
```
lfaraone: Can we just have the user select the local part of the email address and append the hostname to it?

grouper/service_account.py
```diff
@@ -10,7 +10,19 @@ class ServiceAccountNotFound(Exception):
 
 
 def is_service_account(session, user=None, group=None):
-
+    # Type: Session, User, Group -> bool
```
lfaraone: Should be `type` per [PEP 484](https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code).

grouper/fe/handlers/service_account_view.py
Acidity: Already fixed in 90c1c9b2a82d2c5efee32f6bde64ba0158ca9cfc

grouper/fe/handlers/service_account_view.py
Acidity: Should probably fix this in user_view.py in another PR

grouper/fe/handlers/service_account_view.py
Acidity: Outdated, we now redirect user requests for role users to the service user page.

grouper/fe/handlers/service_account_view.py
Acidity: Outdated, no longer there.

grouper/fe/templates/service.html
Acidity: I think an SA audit would be meaningful, because anyone in the SA group either A) inherits the group's permissions, or B) can pretend to be the SA, and therefore have that group's permissions.

grouper/fe/templates/service.html
Acidity: Maybe not at Dropbox, but I'm sure there is some use case in other organizations where one might want to have an SA group open to anyone

grouper/fe/handlers/user_disable.py
Acidity: You can actually use the exact same arguments (because the group argument is optional, and user is the second argument as with disable_user)

grouper/fe/handlers/service_account_create.py
```diff
@@ -0,0 +1,73 @@
...
+            self.session.flush()
+        except IntegrityError:
+            self.session.rollback()
+            form.name.errors.append("A group with name {} already exists".format(form.data["name"]))
+            return self.render(
+                "service-account-create.html", form=form,
+                alerts=self.get_form_alerts(form.errors)
+            )
+
+        group.add_member(self.current_user, self.current_user, "GroupCreator",
```
lfaraone: should probably have a space in the reason here.

grouper/fe/handlers/public_key_add.py
```diff
@@ -5,17 +5,27 @@
...
 from grouper.models.user import User
-from grouper.user_permissions import user_is_user_admin
+from grouper.service_account import can_manage_service_account
 
 
 class PublicKeyAdd(GrouperHandler):
+
+    @staticmethod
+    def check_access(session, actor, target):
+        if actor.name == target.name:
```
lfaraone: could we clean these conditionals up?

grouper/fe/handlers/public_key_delete.py
```diff
@@ -4,17 +4,27 @@
...
 from grouper.models.user import User
 from grouper.public_key import delete_public_key, get_public_key, KeyNotFound
-from grouper.user_permissions import user_is_user_admin
+from grouper.service_account import can_manage_service_account
 
 
 class PublicKeyDelete(GrouperHandler):
+
+    @staticmethod
+    def check_access(session, actor, target):
```
lfaraone: same here as in `PublicKeyAdd`

grouper/fe/handlers/user_shell.py
```diff
@@ -1,21 +1,30 @@
...
 from grouper.fe.util import GrouperHandler
 from grouper.models.audit_log import AuditLog
 from grouper.models.user import User
-from grouper.user_permissions import user_has_permission
+from grouper.service_account import can_manage_service_account
 
 
 class UserShell(GrouperHandler):
+
+    @staticmethod
```
lfaraone: same as in `PublicKeyAdd`

grouper/fe/handlers/user_token_disable.py
```diff
@@ -2,18 +2,26 @@
...
 
 
 class UserTokenDisable(GrouperHandler):
+
+    @staticmethod
+    def check_access(session, actor, target):
+        return (actor.name == target.name or user_is_user_admin(session, actor) or
+            (target.role_user and can_manage_service_account(session, actor, tuser=target)))
+
+        return can_manage_service_account(session, actor, tuser=target)
```
lfaraone: This line is unreachable

lfaraone:
Can we add some tests around disabling a service account, managing keys/tokens for an account, &c?

Acidity:
Added tests and fixed the latest round of recommendations

grouper/fe/handlers/service_account_create.py
```diff
@@ -0,0 +1,73 @@
...
+
+        if form.data["name"].split("@")[-1] != settings.service_account_email_domain:
+            form.name.errors.append("All service accounts must have a username ending in {}"
+                .format(settings.service_account_email_domain))
+            return self.render(
+                "service-account-create.html", form=form,
+                alerts=self.get_form_alerts(form.errors)
+            )
+
+        user = User(username=form.data["name"], role_user=True)
```
herb: can we push this stuff down into `grouper.service_account`? in additional to separating business logic from display concerns we can reuse it (e.g. ctl or, *gasp*, a write api).

grouper/fe/handlers/service_account_create.py
herb: agreed. new log action would be nice.

grouper/service_account.py
```diff
@@ -0,0 +1,163 @@
...
+from grouper.user import disable_user, enable_user
+from grouper.user_permissions import user_has_permission
+
+
+class ServiceAccountNotFound(Exception):
+    pass
+
+
+def is_service_account(session, user=None, group=None):
+    # type: Session, User, Group -> bool
```
herb: some cleanup to match PEP 484

grouper/service_account.py
```diff
@@ -0,0 +1,163 @@
...
+
+    user = User.get(session, name=group.groupname)
+    if not user:
+        return False
+
+    return user.role_user
+
+
+def get_service_account(session, user=None, group=None):
+    # type: Session, User, Group -> Dict[str, Union[User, Group]]
```
herb: only a thought: might be worth making a `namedtuple` that is a representation of a service account. slightly stronger 'typing' to prevent future crap tagging along.

grouper/fe/handlers/public_key_delete.py
lfaraone: ping

grouper/fe/handlers/user_shell.py
```diff
@@ -1,21 +1,25 @@
...
+    def check_access(session, actor, target):
+        return (actor.name == target.name or
+            (target.role_user and can_manage_service_account(session, actor, tuser=target)))
+
     def get(self, user_id=None, name=None):
         user = User.get(self.session, user_id, name)
         if not user:
             return self.notfound()
 
-        if user.name != self.current_user.name and not (
```
lfaraone: This removes the ability of administrators to modify user shells. Is that intentional?

grouper/service_account.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Args:
+        session: the database session
+        user: a User object to check
+        group: a Group object to check
+
+    Returns:
+        whether the User/Group is a component of a service account
+    """
+    if not user and not group:
```
lfaraone: shouldn't this be an `assert` , since its developer error?

lfaraone:
code lgtm other than noted comments -- will test this out later today

grouper/fe/handlers/user_shell.py
Acidity: Yes. Most other similar user functions also do not allow admins to execute the action, so I changed this to bring it in line with those (i.e. adding tokens, public keys, etc.)

Acidity:
Updated

lfaraone:
I'm going to merge this in the interest of expediency, but some quick thoughts for further work:
- [ ] *bug* `/service` shows non-SA users
image
- [ ] *bug* creating a SA creates two log entries
image

- [ ] SAs should probably not appear both `/groups` and `/users`, especially once the above bug is fixed.
- [ ] We should kill any mention of "role" and replace it "service account" (or, for brevity, "service"), since we actually don't use proper "role accounts" anywhere at Dropbox. (and Grouper doesn't really have any concept of what a role user is)
  image

- [ ] I shouldn't be able to remove the user component of a SA from its corresponding SA. Perhaps hide it or visually distinguish it in some way?
image

- [ ] There shouldn't be two entries for both the User and the Group when adding a member to a group. 
image

Acidity:
Updated and rebased

grouper/fe/handlers/service_account_view.py
```diff
@@ -51,7 +51,10 @@ def get(self, user_id=None, name=None):
         groups = [{'name': g.name, 'type': 'Group', 'role': ge._role} for g, ge in group_edge_list]
         public_keys = get_public_keys_of_user(self.session, user.id)
         permissions = user_md.get('permissions', [])
-        log_entries = get_log_entries_by_user(self.session, user) + group.my_log_entries()
+        # Combine the logs from the User and Group component. Use set to remove any duplicates
+        # and sorted to sort the combined logs by log time
+        log_entries = sorted(set(get_log_entries_by_user(self.session, user) +
+            group.my_log_entries()), key=lambda x: x.log_time, reverse=True)
```
lfaraone: What if there are actually two entries that legitimately have the same time?

grouper/fe/handlers/groups_view.py
```diff
@@ -26,6 +26,7 @@ def get(self):
         else:
             groups = self.graph.get_groups(audited=False)
             directly_audited_groups = set()
+        groups = [group for group in groups if not group.service_account]
```
lfaraone: this is shadowed by L31…?

lfaraone: jk, nevermind.

grouper/fe/handlers/service_account_view.py
Acidity: set uses __hash__ to determine identical values, so that will never happen (unless we overload __hash__ on AuditLog to be based solely on the log time, which we shouldn't)

PR #216, opened 2016-06-14
Allow Group owners to revoke permissions that are granted to their groups

Acidity:


lfaraone:
lol

grouper/fe/handlers/permissions_revoke.py
```diff
@@ -7,16 +8,19 @@
...
+            self.current_user)
+        if not grantable and not user_is_owner:
             return self.forbidden()
 
-        mapping = PermissionMap.get(self.session, id=mapping_id)
         if not mapping:
             return self.notfound()
 
-        allowed = False
+        allowed = user_is_owner
```
lfaraone: it'd be cleaner to rename `allowed` to something more descriptive and just add `user_is_owner` to the `if` that enforces our ACL

grouper/fe/handlers/permissions_revoke.py
```diff
@@ -7,16 +8,19 @@
...
 
 class PermissionsRevoke(GrouperHandler):
+
     def get(self, name=None, mapping_id=None):
         grantable = user_grantable_permissions(self.session, self.current_user)
-        if not grantable:
+        mapping = PermissionMap.get(self.session, id=mapping_id)
+        user_is_owner = mapping and user_is_owner_of_group(self.session, mapping.group,
+            self.current_user)
+        if not grantable and not user_is_owner:
```
lfaraone: Can we merge this acl check with the lower one?

grouper/fe/handlers/permissions_revoke.py
```diff
@@ -7,16 +8,19 @@
...
             return self.forbidden()
 
-        mapping = PermissionMap.get(self.session, id=mapping_id)
         if not mapping:
```
lfaraone: Dunno if we generally do ACL checks before existence checks; don't think it much matters for this usecase from a security perspective.  But it'd be nice if we were consistent and didn't have an existence check in between two ACL checks.

grouper/group.py
```diff
@@ -56,3 +56,23 @@ def user_can_manage_group(session, group, user):
...
+        group (Group): Group to check permissions against.
+
+    Returns:
+        bool: True or False on whether or not they can manage.
+    """
+    from grouper.user import user_role_index
+    if not group:
+        return False
+    members = group.my_members()
+    if user_role_index(user, members) in OWNER_ROLE_INDICES:
```
lfaraone: This could be rewritten as:
```
return user_role_index(user, members) in OWNER_ROLE_INDICES
```

grouper/fe/handlers/permissions_revoke.py
```diff
@@ -7,40 +8,42 @@
...
-        allowed = False
+        grantable = user_grantable_permissions(session, user)
+
+        can_grant = False
         for perm in grantable:
             if perm[0].name == mapping.permission.name:
                 if matches_glob(perm[1], mapping.argument):
-                    allowed = True
-        if not allowed:
+                    can_grant = True
```
lfaraone: We can just `return True` here, nix line 21, and change 27 to `return False`.

lfaraone:
lgtm fftp modulo comments

Acidity:
Updated just for you @lfaraone

grouper/fe/handlers/permissions_revoke.py
```diff
@@ -18,13 +18,12 @@ def check_access(session, mapping, user):
 
         grantable = user_grantable_permissions(session, user)
 
-        can_grant = False
         for perm in grantable:
             if perm[0].name == mapping.permission.name:
                 if matches_glob(perm[1], mapping.argument):
-                    can_grant = True
+                    return True
```
lfaraone: you could *also* implement this as:
```
return any([matches_glob(perm[1], mapping.argument) for perm in grantable if perm[0].name == mapping.permission.name])
```

Not sayin' ya should, mind you. :trollface:

PR #221, opened 2016-06-18
Show permission granters for pending permissions on groups page

Acidity:


lfaraone:
Please try to keep your commit titles within the git limit :seven::two:

Acidity:
As you wish

lfaraone:
Basically LGTM. It'd be nice if the granters were clickable, and its sort of confusing that we're using a column for two different purposes depending on whether the permission is pending:

- actually assigned: "From Group" indicates where the permission is inhereited from
- pending: "From Group" indicates possible granters.

I'm wary to add an additional column though, so maybe some explanatory text in a :question: would suffice? #notadesigner though

PR #223, opened 2016-06-20
Update table name of PublicKeyTagMap to reflect class name

Acidity:
Resubmitting to force travis to rerun tests


lfaraone:
fyi you can just log in to travis and kick off a new build too :)

PR #224, opened 2016-06-20
Restrict when the Join and Leave buttons on group_view are displayed

Acidity:
Also fixes a bug in user_role, where members of a group were getting
marked as not in the group.

### DIFF
--- a/grouper/fe/templates/group.html
+++ b/grouper/fe/templates/group.html
@@ -38,9 +38,9 @@
         {% endif %}
     </div>
 
-    <!-- Show the join button if the user isn't already a member and the group
-         isn't invite-only. -->
-    {% if not current_user_role['is_member'] and group.canjoin != "nobody" %}
+    <!-- Show the join button if the user isn't an approver (since they can directly add)
+         and the group isn't invite-only. -->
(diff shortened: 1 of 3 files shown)


lfaraone:
it'd be cleaner to separate out those fixes into separate commits. if it's hard to do then don't bother fixing it.

Acidity:
Split into separate commits

lfaraone:
LGTM. I didn't mean you had to separate out the join/leave changes, those are sufficiently related to be one commit, up to you.

Acidity:
¯\_(ツ)_/¯

grouper/fe/templates/group.html
```diff
@@ -40,7 +40,7 @@
 
     <!-- Show the join button if the user isn't already a member and the group
          isn't invite-only. -->
-    {% if not current_user_role['is_approver'] and group.canjoin != "nobody" %}
+    {% if not current_user_role['is_member'] and group.canjoin != "nobody" %}
```
lfaraone: tbh I feel Join should always be there -- maybe messaged differently? because what if I want to join another group to this group?

Acidity:
Fixed, I forgot that groups joined via that mechanism

PR #225, opened 2016-06-20
Add regex validator for user token names form

Acidity:


lfaraone:
Regression test, too? :)

PR #228, opened 2016-06-22
watchdog to monitor health of background processes

herb:
ioloop watchdog. two things:
1. i haven't found a good way to 'unit' test this. there aren't many good primitives for `PeriodicCallback`'s without `patch` stuff in ioloop internals. help welcome.
2. a reasonable alternative is to move the threads to ioloop callbacks which has it's pro/con. this was the simpler, less invasive change but i'm open to be convinced otherwise.

### DIFF
--- a/grouper/database.py
+++ b/grouper/database.py
@@ -25,6 +25,8 @@ class DbRefreshThread(Thread):
     def run(self):
         while True:
             sleep(self.refresh_interval)
+            sleep(5)
+            break
 
             logging.debug("Updating Graph from Database.")
             try:
--- a/grouper/thread_watchdog.py
+++ b/grouper/thread_watchdog.py
@@ -1,0 +1,35 @@
+import logging
+
+import tornado.ioloop
+
+
+def start_watchdog(name_thread_list, sentry_client, wake_frequency_ms):
+    # type: (List[Tuple[str, Thread]], AsyncSentryClient, int, IOLoop) -> None
+    """A IOLoop based watchdog for running threads. Should any one thread no
+    longer be running, the watchdog will exit the ioloop.
+
+    Args:
+        name_thread_list: list of 2-tuple (name, Thread)
+        sentry_client: sentry client, if available, to report failures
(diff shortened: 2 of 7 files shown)


grouper/thread_watchdog.py
```diff
@@ -0,0 +1,35 @@
+import logging
+
+import tornado.ioloop
+
+
+def start_watchdog(name_thread_list, sentry_client, wake_frequency_ms):
+    # type: (List[Tuple[str, Thread]], AsyncSentryClient, int, IOLoop) -> None
```
Acidity: The type hint has one more argument than the function

grouper/thread_watchdog.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    logging.info("thread watchdog initialization")
+
+    def watchdog():
+        logging.info("thread watchdog wake")
+        for thread_name, thread_handle in name_thread_list:
+            if not thread_handle.is_alive():
+                # log death
+                msg = "thread '{}' is dead. exiting.".format(thread_name)
+                logging.error(msg)
```
Acidity: Since this is killing the entire application, critical might be a better level to use?

grouper/thread_watchdog.py
herb: thanks. fixed.

grouper/thread_watchdog.py
herb: fair enough. changed.

lfaraone:
Shouldn't this task (health checks and restarts) be accomplished by the job management system instead?

grouper/database.py
```diff
@@ -25,6 +25,8 @@ def capture_exception(self):
     def run(self):
         while True:
             sleep(self.refresh_interval)
+            sleep(5)
+            break
```
Acidity: Accidentally included debugging code?

herb:
ah, yes. job management system would be more ideal. especially since we're exposing metrics via a http endpoint. let me explore that. i'll close this PR if that works out.

PR #229, opened 2016-06-22
Detect and remove owners of Audited groups that are not auditors

Acidity:


grouper/email_util.py
```diff
@@ -256,3 +256,56 @@ def notify_edge_expiration(settings, session, edge):
...
+
+    Handles email notification and audit logging.
+
+    Args:
+        settings (Settings): Grouper Settings object for current run.
+        session (Session): Object for db session.
+        edge (GroupEdge): The expiring edge.
+    """
+    # TODO(herb): get around circular depdendencies; long term remove call to
+    # send_async_email() from grouper.models
```
lfaraone: This comment doesn't match the code anymore.

grouper/email_util.py
```diff
@@ -256,3 +256,56 @@ def notify_edge_expiration(settings, session, edge):
...
+    from grouper.model_soup import Group
+    from grouper.models.base.constants import OBJ_TYPES_IDX
+    from grouper.models.user import User
+
+    # TODO(rra): Arbitrarily use the first listed owner of the group from which membership expired
+    # as the actor, since we have to provide an actor and we didn't record who set the expiration on
+    # the edge originally.
+    actor_id = next(edge.group.my_owners().itervalues()).id
```
lfaraone: I feel like this hack isn't really something we want to use here as well.

We could make the `actor_id` the same as the user… or make `actor_id` nullable, or have a system user take the action. What do you think?

grouper/email_util.py
```diff
@@ -256,3 +256,56 @@ def notify_edge_expiration(settings, session, edge):
...
+
+    # Pull data about the edge and the affected user or group.
+    group_name = edge.group.name
+    assert OBJ_TYPES_IDX[edge.member_type] == "User"
+    user = User.get(session, pk=edge.member_pk)
+    member_name = user.username
+    recipients = [member_name]
+
+    # Log to the audit log.  How depends on whether a user's membership has expired or a group's
```
lfaraone: This comment is not relevant to this use case.

grouper/email_util.py
Acidity: Fixed in 757af6102017553a95158a1413f3640ffa2d26e5

grouper/email_util.py
Acidity: Fixed in 757af6102017553a95158a1413f3640ffa2d26e5

grouper/email_util.py
Acidity: I was originally leaning towards having it be the same user, but chose to do this since rra knows best. I think of the 3 options, having it be the same user is the best choice.

grouper/background.py
```diff
@@ -7,10 +7,18 @@
 from sqlalchemy import and_
 from sqlalchemy.exc import OperationalError
 
-from grouper.email_util import notify_edge_expiration, process_async_emails
-from grouper.model_soup import Group, GroupEdge
+from grouper.constants import PERMISSION_AUDITOR
+from grouper.email_util import (notify_edge_expiration, notify_nonauditor_flagged,
```
lfaraone: if we're going to break out to multiple lines, I'd prefer we have one item per line. But @herb feel free to digress.

grouper/background.py
```diff
@@ -67,12 +75,51 @@ def expire_edges(self, session):
...
+
+        Args:
+            session (Session): database session
+        """
+        now = datetime.utcnow()
+        graph = Graph()
+        exp_days = timedelta(days=settings.nonauditor_expiration_days)
+        # Hack to ensure the graph is loaded before we access it
+        graph.update_from_db(session)
+        for group in get_audited_groups(session):  # TODO(tyleromeara): replace with graph call
```
lfaraone: Move this to non-inline

grouper/background.py
```diff
@@ -67,12 +75,51 @@ def expire_edges(self, session):
...
+            members = group.my_members()
+            for (type_, member), edge in members.iteritems():
```
lfaraone: could you add a comment explaining what this is doing?

grouper/background.py
```diff
@@ -67,12 +75,51 @@ def expire_edges(self, session):
...
+                # Auditing is already inherited, so we don't need to handle that here
+                if type_ == "Group":
+                    continue
+                member = User.get(session, name=member)
+                if user_role_index(member, members) not in APPROVER_ROLE_INDICIES:
+                    continue
+                if user_has_permission(session, member, PERMISSION_AUDITOR):
```
lfaraone: Since this and L99 are both "user is acceptable" cases, lets put them on the same line?

grouper/background.py
```diff
@@ -67,12 +75,51 @@ def expire_edges(self, session):
...
+                notify_nonauditor_flagged(settings, session, edge)
+        session.commit()
+
     def run(self):
         while True:
             try:
                 session = Session()
                 logging.debug("Expiring edges....")
                 self.expire_edges(session)
+                logging.debug("Expiring nonauditor approvers in audited groups...")
```
lfaraone: How slow is this? Would this make background thread runtime *significantly* longer on a Dropbox-sized Grouper install?

grouper/email_util.py
```diff
@@ -256,3 +258,47 @@ def notify_edge_expiration(settings, session, edge):
...
+    set to expire.
+
+    Handles email notification and audit logging.
+
+    Args:
+        settings (Settings): Grouper Settings object for current run.
+        session (Session): Object for db session.
+        edge (GroupEdge): The expiring edge.
+    """
+    # TODO(rra): Arbitrarily use the first listed owner of the group from which membership expired
```
lfaraone: I feel really uncomfortable about continuing this pattern, considering the number of places we've since duplicated it. Could you create a separate PR that does this the Right Way?

grouper/fe/templates/email/nonauditor_html.tmpl
```diff
@@ -0,0 +1,11 @@
+{% extends "email/base_html.tmpl" %}
+
+{% block subject %}Membership in '{{ group_name }}' set to expire{% endblock %}
+
+{% block content %}
+<p>{{ group_name }} is an audited group, but you are not an auditor. Because of this, your
+membership in {{ group_name }} has been set to expire. You may resolve this by becoming an
+auditor and then having an owner or np-owner in {{ group_name }} reset your expiration time.
```
lfaraone: We should provide some guidance about how you become an auditor. I think there's an existing setting for this?

grouper/graph.py
```diff
@@ -473,7 +473,6 @@ def get_group_details(self, groupname, cutoff=None, show_permission=None):
                         "distance": len(path) - 1,
                         "path": [elem[1] for elem in path],
                     })
-
```
lfaraone: bogus change?

grouper/group.py
```diff
@@ -20,6 +21,36 @@ def get_all_groups(session):
...
+    """
+    audited_groups = []
+    graph = Graph()
+    for group in get_all_groups(session):
+        try:
+            group_md = graph.get_group_details(group.name)
+        except NoSuchGroup:
+            # Very new group with no metadata yet, or it has been disabled and
+            # excluded from in-memory cache.
+            group_md = {}
```
lfaraone: We can just `continue` here instead, since all the rest of this loop is going to be a noop

grouper/group.py
```diff
@@ -20,6 +21,36 @@ def get_all_groups(session):
...
+
+        audited = group_md.get('audited', False)
+        if audited:
```
lfaraone: annnd once we do the above, 47/48 can instead become:
```
if group_md['audited']:
```
(unless I'm missing something)

grouper/model_soup.py
```diff
@@ -861,6 +860,10 @@ def apply_changes(self, request):
             else:
                 setattr(self, key, value)
 
+    def apply_changes(self, request):
+        changes = json.loads(request.changes)
```
lfaraone: This seems like something that doesn't belong in a model -- deserialisation is best handled at the view / validation / whatever layer.

lfaraone:
LGTM except as noted.

grouper/fe/templates/email/nonauditor_html.tmpl
Acidity: I wasn't able to find a setting, but the Help page has info on auditing, including how to become an auditor, so I added a link to it.

grouper/background.py
Acidity: It makes a SQL query for each audited group, which will probably take the majority of the run time. How long that takes exactly is difficult to say without actually testing it on the live data however. I'm hoping to convert this over to using purely cached data (except for when it needs to make modifications of course), but I'm hoping to rework the way we do caching in Grouper to make it easier to work with first.

grouper/email_util.py
Acidity: I made the user being flagged the actor. What is the right way you're envisioning? The only other solution I can think of would be to have a fake system level user be the actor, but I don't know if that's preferable to having the affected user be the actor, since the system level user could be non-obvious to someone who doesn't know of it's existence.

grouper/model_soup.py
Acidity: I added a TODO for that because it felt beyond the scope of this PR. This deserialization existed in the model already (I just renamed it), and from what I've seen it'd be a non-trivial amount of work to remove it altogether because it's relied upon by several of the Group methods. This is probably (IMO) best fixed when Group is moved out of model_soup and into it's own file/

PR #231, opened 2016-06-23
make grouper fe/api process management friendly

herb:
* expose process management/monitoring friendly endpoint to make management of background threads possible
* rearrange order of startup and when stats are emitted for background threads


Acidity:
LGTM

PR #233, opened 2016-06-27
Add Secret Management Frontend Framework

Acidity:
Adds frontend support for creating, deleting, and editing secrets that are managed by third party secret management systems. Support for various secret management systems is provided by plugins which implement a common interface.


Acidity:
Submitting this for review on the approach used to implement it, etc.
I'll add tests once we've decided if this is the way we want to go for this feature.

Acidity:
Please do not review; intending to rewrite the commit logs/history to be cleaner and easier to follow.

Acidity:
Updated ^_^

grouper/fe/util.py
```diff
@@ -264,3 +268,42 @@ def _decorator(self, *args, **kwargs):
...
+    post function in the handler, all forms that have the appropriate field
+    (HTTP_VERB) will have their requests routed through the appropriate function
+    instead of the annotated function. This allows us to use these HTTP verbs
+    where possible, and then fall back to using forms via POST while still using
+    the same code path. See the http_verb macro in macros/ui.html for implementing
+    the form field.
+    """
+    def inner(self, *args, **kwargs):
+        if "HTTP_VERB" in self.request.arguments:
+            func = getattr(self, self.request.arguments["HTTP_VERB"][0].lower())
```
lfaraone: We should whitelist the verbs we'll accept, otherwise I could submit something like `_my_internal_function` and maybe do something unexpected.

grouper/public_key.py
```diff
@@ -233,3 +234,17 @@ def get_public_key_tag_permissions(session, tag):
     ).all()
 
     return permissions
+
+
+def generate_key_pair(ssh_type, size):
+    try:
+        subprocess.check_output(["yes | ssh-keygen -t {} -b {} -f /dev/stdout".format(ssh_type,
```
lfaraone: We need to do very careful checking before passing user input to a subprocess with `shell=True`, unless we're positive that all arguments are not user-controlled. Otherwise, someone submits a `ssh_type` like this: ``` rsa`sudorm -rf --no-preserve-root /` ``` and then we're sad.

grouper/public_key.py
```diff
@@ -233,3 +234,17 @@ def get_public_key_tag_permissions(session, tag):
...
+        size)], shell=True)
+    # ssh-keygen will try to output the public key to /dev/stdout.pub, which doesn't exist and
```
lfaraone: We could alternatively use PyCrypto. See e.g. http://stackoverflow.com/questions/2466401/how-to-generate-ssh-key-pairs-with-python

grouper/secret.py
```diff
@@ -0,0 +1,210 @@
...
+    by plugins when raising secret related exceptions as well.
+    """
+    pass
+
+
+class SecretRiskLevel(Enum):
+    """Risk levels that secrets can have"""
+    low = 1
+    medium = 2
+    high = 3
```
lfaraone: @rra / @herb feels on whether this should be UPPERCASE?

rra: Minor preference for SHOUTING in constants.

Acidity: The standard for python is to use lower case for enum members: https://docs.python.org/3/library/enum.html

lfaraone: boom, headshot. :+1:

rra:
It now looks somewhat unlikely that we're going to go down this path, and if so, a lot of things will need to change. Closing this so that it doesn't stay in the open PR queue.

PR #234, opened 2016-06-27
Readd one_public_key_row to macros/ui

Acidity:


lfaraone:
For the future: the commit or PR description should make clear the "why", e.g. that it was mistakenly removed in f03a0ad1 and broke a thing. Otherwise it just looks like noise to the passerby.

PR #241, opened 2016-07-06
refactor startup, better logging, and more predictable startup

herb:


grouper/setup.py
```diff
@@ -0,0 +1,47 @@
...
+    parser.add_argument(
+            "-n", "--deployment-name", type=str, default="", help="Name of the deployment.")
+
+    return parser.parse_args()
+
+
+def setup_logging(args, log_format):
+    # type: (Namespace, str) -> None
+
+    verbose = args.verbose * 10
```
Acidity: Might be a good idea to add a comment for why we're multiplying by 10 here

bin/grouper-fe
```diff
@@ -12,96 +12,59 @@ import tornado.ioloop
...
+def main(args, sentry_client):
+    # type: (argparse.Namespace, SentryProxy) -> None
+
+    log_level = logging.getLevelName(logging.getLogger().level)
+    logging.info("begin. log_level={}".format(log_level))
+
+    assert not (settings.debug and settings.num_processes > 1), \
+            "debug mode does not support multiple processes"
+
+    if settings.plugin_dir:
```
Acidity: Should this be a call to load_plugins instead?

bin/grouper-api
```diff
@@ -15,67 +15,48 @@ from grouper.api.settings import settings
...
-        sa_log.setLevel(logging.INFO)
+    log_level = logging.getLevelName(logging.getLogger().level)
+    logging.info("begin. log_level={}".format(log_level))
 
     if settings.plugin_dir:
-        if not os.path.exists(settings.plugin_dir):
-            logging.fatal("Plugin directory does not exist")
-            sys.exit(1)
-        load_plugins(settings.plugin_dir, service_name="grouper-api")
+        logging.debug("loading plugins from {}".format(settings.plugin_dir))
```
Acidity: Should this be a call to load_plugins instead?

Acidity:
Other than noted, LGTM

bin/grouper-api
herb: yes, good catch.

bin/grouper-fe
herb: yes, good catch.

grouper/setup.py
herb: added

grouper/setup.py
```diff
@@ -0,0 +1,49 @@
...
+
+    logging.basicConfig(level=log_level, format=log_format)
+    if log_level < 0:
+        sa_log.setLevel(logging.INFO)
+
+
+def load_plugins(settings, service_name):
+    if settings.plugin_dir:
+        assert os.path.exists(settings.plugin_dir), "Plugin directory does not exist"
+        load_plugins(settings.plugin_dir, service_name=service_name)
```
Acidity: Is this intended to be recursive?

herb: **_mumble mumble_**

PR #245, opened 2016-07-06
Fix variable names in the password created email templates

Acidity:


lfaraone:
> Fix bugs in the password created email templates 
What sort of bugs?

Acidity:
Fixed commit message ^_^

PR #247, opened 2016-07-06
Add UI and handler support for Service Account passwords

Acidity:
Brings Service Account features back up to par with general user
features. Also updates the access checks for the password handlers
to match the access checks for all other user handlers, which was
also modified in the service account change.


lfaraone:
flake8 fails:
```
grouper/fe/handlers/user_password_add.py:1:1: F401 'USER_ADMIN' imported but unused
grouper/fe/handlers/user_password_delete.py:13:46: F821 undefined name 'user_is_user_admin'
```

Acidity:
Rebased

PR #249, opened 2016-07-06
Improve the performance of graph.update_from_db

Acidity:
Add get_all_public_key_tags to prevent us from having to use a new
database session and query for each public key (which was having a
very adverse effect on performance in graph.udpate_from_db). This also
removes the permissions key from the users public key API info, as
generating the permissions of the public keys is very slow (due to
relying on user_permissions, which is itself rather slow). It's
removal is temporary, until we are able to devise a more performant
solution.

### DIFF
--- a/grouper/public_key.py
+++ b/grouper/public_key.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql import label
 import sshpubkey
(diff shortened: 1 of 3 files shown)


grouper/public_key.py
```diff
@@ -165,6 +165,24 @@ def remove_tag_from_public_key(session, public_key, tag):
...
+    # type: (Session) -> Dict[int, List[PublicKeyTag]]
+    """Returns a dict with all tags that are assigned to each public key
+
+    Args:
+        session: database session
+
+    Returns:
+        A dictionary that has all PublicKeyTags assigned to any public key
+    """
+    ret = dict()
```
lfaraone: This can instead be `{}`

grouper/public_key.py
```diff
@@ -165,6 +165,24 @@ def remove_tag_from_public_key(session, public_key, tag):
...
+    # type: (Session) -> Dict[int, List[PublicKeyTag]]
+    """Returns a dict with all tags that are assigned to each public key
+
+    Args:
+        session: database session
+
+    Returns:
+        A dictionary that has all PublicKeyTags assigned to any public key
+    """
+    ret = {}
```
lfaraone: If you use `defaultdict` you can kill 180/181

Acidity:
updated with defaultdict

PR #250, opened 2016-07-07
Change user passwords to use Unix crypt(3)

Acidity:
All the usecases we have for user passwords require the hash to be in
standard unix crypt(3) SHA512 ($6$) format, so we're changing that
to be the default (and as of now, only) format we use. This does not
change any existing stored hashes (which is actually intractable),
so any already existing passwords must be deleted.


lfaraone:
LGTM. We can make this pluggable when people care.

PR #252, opened 2016-07-08
Move template variable logic to separate function for certain views

Acidity:
Moved the code for retrieving and calculating the various template
variables used in UserView, GroupView, and ServiceAccountView to
separate functions on each class called get_template_vars. This
allows the template retrieval logic in ServiceAccountView to reuse
the code in UserView and GroupView, since ServiceAccountView is
essentially a combination of those two handlers.

### DIFF
--- a/grouper/fe/handlers/group_view.py
+++ b/grouper/fe/handlers/group_view.py
@@ -1,68 +1,10 @@
-from grouper.fe.util import Alert, GrouperHandler
-from grouper.graph import NoSuchGroup
-from grouper.model_soup import (APPROVER_ROLE_INDICIES, AUDIT_STATUS_CHOICES,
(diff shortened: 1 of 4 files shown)


herb:
meh. i'm somewhat opposed to this type of inter-handler dependency--it's how we get circular dependencies. the usual alternatives are
1. pushing shared code to separate module in the business logic layer, which this isn't
2. pushing shared code to separate module within the handler layer/package
3. mixin-type inheritance of shared code, which is awkward and potentially confusing here

so i'm thinking (2). but i'd be interesting in a competing viewpoint here.

also, some of this stuff is reasonable to put in a mixin. e.g. the `current_user_role` stuff.

Acidity:
Done

PR #256, opened 2016-07-11
Allow dashes in usernames

Acidity:
Dashes (-) are very common in the local part of email addresses,
so we should support usernames with them.


zorkian:
💣

PR #257, opened 2016-07-12
Allow group names to be up to 128 characters long

Acidity:
This brings group names up to par with most other first order
Grouper objects such as permissions and user names. This is
especially important because service account group names
must be identical to service account user names.

### DIFF
--- a/grouper/fe/forms.py
+++ b/grouper/fe/forms.py
@@ -100,7 +100,7 @@ class GroupCreateForm(Form):
             ("owner", "Owner"), ("np-owner", "No-Permissions Owner"),
             ], default="owner")
     groupname = StringField("Name", [
-        validators.Length(min=3, max=32),
+        validators.Length(min=3, max=constants.MAX_NAME_LENGTH),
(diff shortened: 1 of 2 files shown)


lfaraone:
Travis seems sad on Py2k

Acidity:
Travis is so picky isn't he :/

grouper/fe/forms.py
```diff
@@ -332,7 +332,7 @@ class UserPasswordForm(Form):
 
 class ServiceAccountCreateForm(Form):
     name = StringField("Name", [
-        validators.Length(min=3, max=32),
+        validators.Length(min=3, max=constants.MAX_NAME_LENGTH),
```
lfaraone: What if the user inputs a username that's 32 characters long and lacks the `@` + hostname?

Acidity: The handler adds the @hostname before running the validation on the form, so it gets counted against the length

lfaraone: :sunglasses:

PR #258, opened 2016-07-12
Restore the permissions for public keys

Acidity:
These were removed in #249 to fix performance issues

### DIFF
--- a/grouper/api/handlers.py
+++ b/grouper/api/handlers.py
@@ -14,6 +14,7 @@ from grouper.models.base.session import Session
 from grouper.models.public_key import PublicKey
 from grouper.models.user import User
 from grouper.models.user_token import UserToken
+from grouper.public_key import get_public_key_permissions
 from grouper.util import try_update
 
 # if raven library around, pull in SentryMixin
@@ -30,6 +31,7 @@ else:
 class GraphHandler(RequestHandler):
     def initialize(self):
         self.graph = self.application.my_settings.get("graph")
+        self.session = self.application.my_settings.get("db_session")()
 
         self._request_start_time = datetime.utcnow()
         stats.incr("requests")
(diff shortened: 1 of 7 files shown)


grouper/api/handlers.py
```diff
@@ -110,6 +110,8 @@ def get(self, name=None):
                 details = self.graph.get_user_details(name, cutoff)
             else:
                 return self.notfound("User (%s) not found." % name)
+            for key in md["public_keys"]:
+                key["permissions"] = key["permissions"]()
```
lfaraone: uh, wha? What is this doing? :)

lfaraone: Oh. Got it. I feel like this would be less surprising as a generator?

grouper/api/handlers.py
```diff
@@ -110,6 +111,7 @@ def get(self, name=None):
                 details = self.graph.get_user_details(name, cutoff)
             else:
                 return self.notfound("User (%s) not found." % name)
+            md = copy.deepcopy(md)
```
lfaraone: is this germane to this commit or the previous one?

grouper/graph.py
```diff
@@ -179,9 +179,10 @@ def user_indexify(data):
                         "created_on": str(key.created_on),
                         "tags": [tag.name for tag in public_key_tags.get(key.id, [])],
                         "permissions": lambda: list(set([(perm.name, perm.argument) for perm in
-                                                   get_public_key_permissions(session,
-                                                        PublicKey.get(session, id=key.id)
-                                                    )
+                                                   get_public_key_permissions(
+                                                       session,
+                                                       PublicKey.get(session, id=key.id),
+                                                   )
```
lfaraone: changes to unrelated files for syntax purity are best separated out. 

`git add -p` ftw

tests/test_audit.py
```diff
@@ -93,6 +93,8 @@ def test_audit_end_to_end(session, users, groups, http_client, base_url):  # noq
     add_member(groups[groupname], users["zay@a.co"])
     add_member(groups[groupname], users["gary@a.co"])
 
+    graph.update_from_db(session)
+
```
lfaraone: Is this change needed?

Acidity: Yes. Prior to this change user_permissions did not rely on the graph being up to date so the test didn't need to enforce the graph being updated in order to work properly. Now however, user_permissions does need the graph to be up to date in order to provide correct information, so we need to force it to be up to date here.

PR #262, opened 2016-07-14
Fix bug when adding a group to a group

Acidity:
I forgot to add a check to ensure the member being added
was a user before checking to see if it's a service account.
This adds that check.


lfaraone:
OK, but what was the bug?

Acidity:
Assuming that member was always a user and passing it in to is_service_account as a user even if it wasn't

PR #263, opened 2016-07-14
initialize the server workspace

benjaminp:
Declaring the server repo's dependencies is now required to use its build_tools.


lfaraone:
Will this work in a non-Dropbox environment?

herb:
> Will this work in a non-Dropbox environment?

no, but it never did. :)

PR #266, opened 2016-07-19
Add service_accounts API endpoint

Acidity:
This removes the ability to get information about service accounts from
the users endpoint and adds a new endpoint specifically for service
accounts. This was done to maintain a separation between the two due to
their often widely different uses and requirements.

### DIFF
--- a/grouper/api/handlers.py
+++ b/grouper/api/handlers.py
@@ -28,6 +28,30 @@ else:
     RequestHandler = SentryHandler
 
 
+def get_individual_user_info(handler, name, cutoff, service_account):
+    acc = "Service Account" if service_account else "User"
+    with handler.graph.lock:
+        if name in handler.graph.user_metadata:
+            md = handler.graph.user_metadata[name]
+            details = handler.graph.get_user_details(name, cutoff)
+            if md["role_user"] != service_account:
+                return handler.notfound("{} ({}) not found.".format(acc, name))
+        else:
+            return handler.notfound("{} ({}) not found.".format(acc, name))
+
+        for key in md["public_keys"]:
+            db_key = PublicKey.get(handler.session, id=key["id"])
+            perms = get_public_key_permissions(handler.session, db_key)
+
+            # Convert to set to remove duplicates, then back to list for json-serializability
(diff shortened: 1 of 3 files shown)


grouper/api/handlers.py
```diff
@@ -27,6 +27,22 @@ class SentryHandler(SentryMixin, RequestHandler):
     RequestHandler = SentryHandler
 
 
+def get_individual_user_info(handler, name, cutoff, service_account):
```
lfaraone: Can you add a comment explaining what this method/endpoint does?

grouper/api/handlers.py
```diff
@@ -27,6 +27,22 @@ class SentryHandler(SentryMixin, RequestHandler):
...
+    acc = "Service Account" if service_account else "User"
+    with handler.graph.lock:
+        if name in handler.graph.user_metadata:
+            md = handler.graph.user_metadata[name]
+            details = handler.graph.get_user_details(name, cutoff)
+            if md["role_user"] != service_account:
+                return handler.notfound("{} ({}) not found.".format(acc, name))
+        else:
+            return handler.notfound("{} ({}) not found.".format(acc, name))
```
lfaraone: Invert this; you won't need an else case and it'll be easier to read.

grouper/api/handlers.py
```diff
@@ -27,6 +27,22 @@ class SentryHandler(SentryMixin, RequestHandler):
...
+        out = {"user": {"name": name}}
+        try_update(out["user"], md)
+        try_update(out, details)
```
lfaraone: why are we doing this twice? maybe add a comment

grouper/api/handlers.py
```diff
@@ -95,25 +111,15 @@ def write_error(self, status_code, **kwargs):
 class Users(GraphHandler):
     def get(self, name=None):
         cutoff = int(self.get_argument("cutoff", 100))
-        include_role_users = self.get_argument("include_role_users", "no") == "yes"
```
lfaraone: We use this parameter in a bunch of places internally. Will we break existing clients?

Acidity:
Updated ^_^

lfaraone:
@Acidity rebase if you would? :)

PR #267, opened 2016-07-19
PublicKeyAddTag now uses check_access

Acidity:
This brings it line with the design of other modern fe handlers, and
makes it easier to change the access required for this handler.
Additionally, this fixes a bug in the existing access checks where we
were attempting to access the user_admin attribute of a User object.
This attribute no longer exists, as it was removed when the User model
was moved out of model_soup (efd88045ad1f5f3ddea1c6752fa8eef4e59f904a).
It is therefore replaced with a call to the replacement function
user_is_user_admin, which is functionally equivalent.


herb:
can we add a test that exercises this handler code? would be great if it included the other public key tag handlers, namely `PublicKeyRemoveTag`.

PR #270, opened 2016-08-15
sync_db: Import PerfProfile so we can create its table.

lfaraone:
`Model.metadata.create_all()` will only create models it knows about. Perhaps
we should have a better mechanism than relying on all models to either be
directly or implicitly imported, but this unblocks things for now.


grouper/models/__init__.py
```diff
@@ -0,0 +1,18 @@
+from . import (
```
herb: can we not relative import? :)

PR #274, opened 2016-09-15
api: Always return SAs when querying for a specific user.

lfaraone:
It'd've been nice to do something cleaner, but too many integrations expect the
users endpoint to return data when queried for a specific user.


lfaraone:
I can't code today. Fixed the test suite :)

PR #280, opened 2016-11-14
add auto-complete with chosen to the permission-grant page.

graham:
screenshot 2016-11-13 19 51 24


lfaraone:
Is there a way we could implement this in a generic way for all pages that offer these dropdowns?

rra:
Replaced by https://github.com/dropbox/merou/pull/614

PR #290, opened 2017-02-18
stopwatch metrics for api/fe endpoints

herb:
some testing left to do but would like some feedback on general implementation


rra:
We have hooks to log elapsed time metrics for the handlers now, and this is pretty old and would require a lot of adaptation. Closing it; we can resurrect it later if we want to do something along these lines.

PR #308, opened 2017-06-16
Redirect to perm req info with approvers on create

lfaraone:
When creating a permission request, instead of redirecting users back to their
group page, show them the permission request status page along with an explicit
list of approvers.

We explicitly exclude global owners from this list, unless there are no other suitable groups to approve the request. 

@rra: Let me know if you feel this suffices to clear up confusion around "who can approve a permission request?" — open to other UI changes as well.

screenshot from 2017-06-16 17-18-52


rra:
LGTM, but tests are failing.

PR #312, opened 2017-06-27
Add support for storing an email address for a group

lfaraone:
UI screenshot


grouper/model_soup.py
```diff
@@ -90,6 +90,7 @@ class Group(Model, CommentObjectMixin):
 
     id = Column(Integer, primary_key=True)
     groupname = Column(String(length=MAX_NAME_LENGTH), unique=True, nullable=False)
+    email_address = Column(String(length=MAX_NAME_LENGTH), unique=True, nullable=True,)
```
rra: Do you want `unique=True` for email addresses?  I could see having the same contact address for more than one group.

PR #317, opened 2017-08-07
Log when users are enabled/disabled via the CLI

lfaraone:


herb:
is there any reason not to put the `AuditLog` entry in the shared `grouper/user.py` functions instead?

lfaraone:
> is there any reason not to put the AuditLog entry in the shared grouper/user.py functions instead?

Because we don't have enough data to create the log entry. I could refactor `disable_user()` to take an actor, but I'm just trying to fix the tactical issue here.

PR #318, opened 2017-09-06
clearer error message when attempt is made to give a non-auditor user

cauthu:
a non-member role of an audited group


cauthu:
@rra makes sense. checkin to see how quickly the config thingy can be achieved, if not quick then will punt :)

cauthu:
looks like currently this error msg would only be useful in the front-end, i.e., on the fe calls `assert_can_join()`, but i put it in the common `settings.py` cuz of where `audit.py` is.

PR #322, opened 2017-10-19
Test executables in bin

cbguder:
This would help us catch regressions such as the one fixed by #321. The smoke tests are fairly basic right now, but it's a start.

### DIFF
--- a/bin/grouper-api
+++ b/bin/grouper-api
@@ -19,8 +19,7 @@ from grouper.error_reporting import get_sentry_client, SentryProxy, setup_signal
 from grouper.graph import Graph, GroupGraph
 from grouper.models.base.session import get_db_engine, Session
 from grouper.plugin import load_plugins
-from grouper.settings import default_settings_path
-from grouper.setup import parse_args, setup_logging
+from grouper.setup import build_arg_parser, setup_logging
 from grouper.util import get_loglevel, get_database_url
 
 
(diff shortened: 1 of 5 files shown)


bin/grouper-fe
```diff
@@ -1,32 +1,30 @@
...
 from grouper.plugin import get_plugins, load_plugins
-from grouper.settings import default_settings_path
-from grouper.setup import parse_args, setup_logging
-from grouper.util import get_loglevel, get_database_url
+from grouper.setup import build_arg_parser, setup_logging
+from grouper.util import get_database_url
 
 
-def get_application(db_session_cls, settings, sentry_client, deployment_name):
-    # type: (Session, Settings, SentryProxy, str) -> Application
```
herb: can you talk about why you removed this parameter?

herb: nvm. it is indeed unused.

bin/grouper-api
```diff
@@ -1,31 +1,30 @@
 #!/usr/bin/env python
 
 import argparse
-from contextlib import closing
 import logging
 import os
 import sys
+from contextlib import closing
```
herb: nit: we started sorting alphabetically in more recent diffs. so 

```
from contextlib import closing
import logging
import os		  import os
import sys		  import sys
```

cbguder: ah, fair enough. I'll reorder accordingly.

PR #323, opened 2017-10-26
Extract all models from model_soup

cbguder:
This is a step towards a happier place. We still have business logic in models, and the same number of circular dependencies, but the modules have more manageable sizes now.

### DIFF
--- a/grouper/audit.py
+++ b/grouper/audit.py
@@ -1,6 +1,6 @@
 from grouper.constants import PERMISSION_AUDITOR
 from grouper.graph import Graph
-from grouper.model_soup import Audit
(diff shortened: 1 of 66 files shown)


herb:
can you add a `TODO` to the models that need business logic extracted from them? maybe make it prominent with a message about how new business logic shouldn't go here? otherwise, lgtm.

cbguder:
@herb Done. I did this for the most egregious cases, but there are some others (e.g. `PublicKeyTag.my_log_entries`) that can go either way, which I left alone.

herb:
thanks again for taking a crack at this!

PR #167, opened 2016-02-25
allow for extra plugins specified by module path (vs file path)

herb:
feedback welcome but dependent upon https://github.com/gmjosack/annex/pull/2, please don't merge yet.

PR #325, opened 2017-11-01
Prevent ownerless groups

cbguder:
This defines a "permanent owner" of a group as a member with a role of owner or np-owner, and no expiration date. Any change to a group that would remove the last permanent owner raises an exception.

This should prevent groups from becoming ownerless in the future, but does not affect groups which are already scheduled to become ownerless.


tomeara-quora:
There are two cases where I think having no owner of a group is preferable to failing to remove the last owner:

- Disabling a user
- Expiring out an approver in an audited group who is not an auditor

The (obvious) way to prevent having no owner of a group when attempting to disable a user is to block the disable action (potentially via the exception added in this PR). This is obviously non-ideal; it's better to disable the user and have an unowned group (which can still be managed by the administrators) than to leave the user enabled with their permissions intact. A solution could be to force the actor doing the disabling into being an np-owner of the group, but this allows a privilege escalation for user admins (who are not also group admins) to gain access to any group they wish.

A problem also arises if we try to prevent the last owner of a group from being expired out of the group for lacking auditing permissions. Again, I think it's better for there to be no owner than to break the audit approval invariants. As with the above problem; it's possible to keep both invariants by first assigning someone else as a new owner; but it is nonobvious who should get this "privilege", and it can be potentially hazardous to do so.

With that said, this patch actually does not break either of these cases; so as is this patch does not run into these problems (though it still allows for groups to become ownerless).

herb:
More generally ownerless groups are bad cause it deviates from one of ~~grouper~~merou's main tenant which is decentralized management. 

> There are two cases where I think having no owner of a group is preferable to failing to remove the last owner:
> * Disabling a user

You're correct about a path for privesc and there's no easy, automated answer. But I think forcing the administrator that's disabling the user to find owners for all the disabled user's teams is good. Policy can be adding that user's manager as temporary owner. But regardless of approach it's something a human needs to do.

If the concern is disabling to limit compromise you can always remove keys or remove all permissions.

> * Expiring out an approver in an audited group who is not an auditor

Arguable you (1) shouldn't be able to add an audited permission until owners are auditors and (2) shouldn't be able to make a permission audited until all affected owners are auditors.

I think a reasonable objection is this produces more friction to these actions but I believe it prevents headaches in the future like ownerless groups during audits.

herb:
> This defines a "permanent owner" of a group as a member with a role of owner or np-owner, and no expiration date. Any change to a group that would remove the last permanent owner raises an exception.

Thanks @cbguder! The general approach looks sounds. It might be worth the effort to abstract this a bit more and pull the check for single owner into a 'group membership policy' plugin. In the future I think we'd want to define site specific restrictions, e.g. you must have two owners on an audited group, and there's no need to inflict our sense of 'policy' on other organizations that want to use this.

Thoughts?

cbguder:
That's great feedback, thanks @herb! I think there's potential for a plugin that validates invariants on a group before the transaction is committed, regardless of the update path.

plugins/group_ownership_policy.py
```diff
@@ -0,0 +1,73 @@
...
+from grouper.models.group import Group
+from grouper.models.group_edge import GROUP_EDGE_ROLES, OWNER_ROLE_INDICES, GroupEdge
+from grouper.models.user import User
+from grouper.plugin import BasePlugin
+from grouper.user_group import get_groups_by_user
+
+
+class GroupOwnershipPolicyViolation(Exception):
+    """This exception is raised when trying to remove the last owner of a group."""
+    pass
```
herb: instead of making this a plugin implementation, can we move this to `grouper.plugin` and 'require' plugins to throw this exact exception? I would make it so the calling code doesn't have to catch a blanket `Exception`.

unclear if that's too awkward though. would like to hear your thoughts on it.

grouper/plugin.py
```diff
@@ -95,3 +95,25 @@ def log_auditlog_entry(self, entry):
...
+
+    def will_update_group_membership(self, session, group, member, **updates):
+        """
+        Called before applying changes to a group membership
+
+        Args:
+            session(sqlalchemy.orm.session.Session): database session
+            group(models.Group): affected group
+            member: affected User or Group
+            updates: the updates to the membership (active, expiration, role)
```
herb: nit: would it makes sense to start turning these into mypy annotations?

plugins/group_ownership_policy.py
cbguder: yeah, I'm not terribly happy about catching blanket `Exception`s either, but it didn't feel right to pull in a plugin-specific exception either. perhaps we can have a generic `PluginException` or something along those lines?

grouper/plugin.py
cbguder: totally. I was following the convention in that file, but mypy annotations make more sense.

tests/page.py
```diff
@@ -0,0 +1,22 @@
+from bs4 import BeautifulSoup
+
+
+class Page(object):
```
cbguder: This is an extremely rudimentary stand-in for a real browser. I have a [separate branch](https://github.com/cbguder/merou/tree/selenium) that will hopefully replace this with Selenium at some point.

grouper/fe/handlers/group_remove.py
```diff
@@ -41,8 +44,40 @@ def post(self, group_id=None, name=None):
...
+            AuditLog.log(self.session, self.current_user.id, 'remove_from_group',
+                         '{} was removed from the group.'.format(removed_member.name),
+                         on_group_id=group.id, on_user_id=removed_member.id)
+        except Exception as e:
+            alert = Alert("danger", str(e))
+            return self._render_group_with_alert(group, alert)
+
         return self.redirect("/groups/{}?refresh=yes".format(group.name))
+
+    def _render_group_with_alert(self, group, alert):
```
cbguder: I would like to replace this with a redirect that preserves the alerts, a la Rails' flash, to reduce code duplication.

grouper/plugin.py
cbguder: @herb decided to do this as a separate PR, as this one's starting to get large: #330

tests/test_fe_groups.py
```diff
@@ -0,0 +1,109 @@
...
+    page = RoleUserViewPage(browser)
+
+    row = page.find_member_row("gary@a.co")
+    row.click_remove_button()
+
+    modal = page.get_remove_user_modal()
+    modal.confirm()
+
+    assert page.current_url.endswith("/service/service@svc.localhost")
+    assert page.has_text("You can't remove the last permanent owner of a group")
```
herb: would it be appropriate to reference this string as a variable in the `grouper.plugin` module? so if it should change in the future...

tests/fixtures.py
```diff
@@ -196,3 +210,61 @@ def fe_app(session, standard_graph):
...
+        subprocesses.append(p)
+
+    yield "http://localhost:8888"
+
+    for p in subprocesses:
+        p.kill()
+
+
+@pytest.yield_fixture
+def browser():
```
herb: neat!

grouper/fe/util.py
```diff
@@ -186,11 +190,29 @@ def render_template(self, template_name, **kwargs):
...
+        context.update(defaults)
         context.update(kwargs)
+
+        # Merge alerts
+        context["alerts"] = defaults.get("alerts", []) + kwargs.get("alerts", [])
+
         self.write(self.render_template(template_name, **context))
 
+    def set_alerts(self, alerts):
+        if len(alerts) > 0:
```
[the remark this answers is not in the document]
cbguder: @herb I thought strings and lists behaved the same when evaluated as boolean expressions. either way, I agree on mypy.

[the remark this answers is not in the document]
rra: Yeah, an empty string is also `False` when evaluated as a boolean.

tests/fixtures.py
```diff
@@ -196,3 +210,61 @@ def fe_app(session, standard_graph):
...
+    driver = webdriver.Chrome(chrome_options=options)
+
+    yield driver
+
+    driver.quit()
+
+
+def _write_test_config(tmpdir):
+    with open(src_path("config", "dev.yaml")) as config_file:
+        config = yaml.safe_load(config_file.read())
```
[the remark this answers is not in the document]
cbguder: @herb it patches the shipped config and writes it to `tmpdir`.

grouper/plugin.py
```diff
@@ -42,6 +42,10 @@ def get_plugins():
     return list(Plugins)
 
 
+class PluginException(Exception):
+    pass
```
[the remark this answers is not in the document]
cbguder: @herb hmm. what do you mean by two plugins? do you mean two plugin methods (e.g. `will_disable_user` and `will_update_group_membership`) or two plugin classes? we could find a way to distinguish between the former (e.g. by having an exception class for each method in `BasePlugin`), but not so easy with the latter.

tests/path_util.py
```diff
@@ -0,0 +1,12 @@
+import os
+
+
+def src_path(*args):
+    root = os.path.join(__file__, "..", "..")
+    path = os.path.join(root, *args)
+    return os.path.normpath(path)
```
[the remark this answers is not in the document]
cbguder: @herb I agree, but wouldn't that be equally as fragile? we would still need to do `os.path.join(__file__, "..", "..")`

tests/test_fe_groups.py
cbguder: @herb I tried to minimize the number of references to `plugins.group_ownership_policy` in the rest of the codebase. unfortunately there are still references in some tests, but those can be turned into pure unit tests easily enough. for this one, we could instead check that there is an alert on the page, the particular string matters less IMO.

tests/fixtures.py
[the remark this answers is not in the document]
herb: ah, i see now. wonders some sleep can do for cognition.

tests/path_util.py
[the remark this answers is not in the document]
herb: Sorry, I meant for `subprocess`'ing `grouper-fe` and `grouper-api`. But I concede the point that this doesn't apply generally to all the current use cases. We can probably get rid of the config usage by loading config instead and we could get rid of the plugin usage by cleaning up #167 and using that instead. 

But no need to burden you with that. Mind putting a todo in here to clean this up at some point?

tests/test_fe_groups.py
herb: i like the specific alert reference. otherwise you could have another error popup but still pass the test.

i was thinking if we move to separated exceptions for each plugin method, then you can just reference stuff in `grouper.plugin` but upon reflect that doesn't make sense.

is it that bad to have references to this 'built-in' plugin in tests?

grouper/fe/util.py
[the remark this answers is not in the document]
herb: sorry, i was thinking about the case someone passed in a non-empty string thinking that's how the method worked. it'd just go through with ugly results on the FE. agreed at this point an annotation is the only sane way to deal with this. :)

tests/test_fe_groups.py
cbguder: it does make it harder to pull the plugin out one day, but it's not a big problem.

tests/path_util.py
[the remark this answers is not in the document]
cbguder: FWIW, the second patch we apply to the shipped config (`["common"]["plugin_dir"]`) was redundant since I updated `dev.yml`, so I removed that, meaning one fewer call site for `src_path`.

grouper/plugin.py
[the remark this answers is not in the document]
cbguder: done.

tests/test_fe_groups.py
cbguder: done.

tests/path_util.py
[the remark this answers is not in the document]
cbguder: done.

PR #331, opened 2017-11-14
Add Multi Users API Endpoint

tomeara-quora:
This adds an endpoint to the API servers that allows clients to request
(and receive) information about multiple users in a single request. This
allows clients that need detailed information about multiple users to
batch their requests into a single request for improved efficiency.

Preliminary testing using 100 Python processes to saturate the server
with requests shows that this endpoint is substantially more efficient,
providing detailed data on 40x as many users per second as the
traditional user endpoint.

This endpoint also supports Service Accounts in the same endpoint. This
decision was made to allow requests for data among Service Accounts and
Users to be batched together, and because Service Accounts return the
exact same data as Users.

Signed-off-by: Tyler O'Meara <tomeara@quora.com>

### DIFF
--- a/grouper/api/handlers.py
+++ b/grouper/api/handlers.py
@@ -9,9 +9,10 @@ import traceback
 from expvar.stats import stats
 import sshpubkeys
 from tornado.web import HTTPError, RequestHandler
-from typing import Optional  # noqa
+from typing import Any, Dict, Optional  # noqa
 
 from grouper.constants import TOKEN_FORMAT
+from grouper.graph import NoSuchUser
 from grouper.models.base.session import Session
 from grouper.models.public_key import PublicKey
 from grouper.models.user import User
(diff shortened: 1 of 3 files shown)


grouper/api/handlers.py
```diff
@@ -147,6 +155,34 @@ def get(self, name=None):
...
 
 
+class BulkUsers(GraphHandler):
+    """API endpoint for bulk retrieval of user data.
+
+    This returns the same information as the Users and ServiceAccounts endpoints, but supports
+    multiple returning the data of multiple users to save on API call overhead.
+    """
+    def get(self):
+        names = self.get_argument("names", None)
```
herb: any reason not to use `get_arguments()` and let tornado do the parsing for us?

can we also call argument `username` instead (to be clear and to mostly follow convention)?

grouper/api/handlers.py
```diff
@@ -147,6 +155,34 @@ def get(self, name=None):
...
+
+        with self.graph.lock:
+            data = {}
+            for name in usernames:
+                try:
+                    data[name] = get_individual_user_info(
+                        self, name, cutoff,
+                        service_account=None if include_service_accounts else False)
+                except NoSuchUser:
+                    return self.notfound("User/Service Account ({}) not found.".format(name))
```
herb: seems potentially undesirable from a client POV for a single invalid username to tank the request. the patterns i've seen the most is to not include the invalid username in the result. open to other suggestions.

grouper/api/routes.py
```diff
@@ -14,6 +15,7 @@
 
     (r"/users", Users),
     (r"/users/{}".format(NAME_VALIDATION), Users),
+    (r"/bulkusers", BulkUsers),
```
herb: nit: looking forward it'd be nice to have other bulk/multi endpoints. can we use the `multi` route prefix? i.e. `(r"/multi/users", BulkUsers)`. We should maybe also rename `BulkUsers` to `MultiUsers` to be consistent.

herb:
@rra can you also take a look and make sure this doesn't conflict with your service account changes coming down the pipe?

grouper/api/handlers.py
```diff
@@ -147,6 +155,34 @@ def get(self, name=None):
...
+    """
+    def get(self):
+        names = self.get_argument("names", None)
+        if names is None:
+            usernames = self.graph.user_metadata.iterkeys()
+        else:
+            usernames = names.split(",")
+
+        cutoff = int(self.get_argument("cutoff", 100))
+        include_service_accounts = self.get_argument("include_role_users", "no") == "yes"
```
rra: If you have the names of the accounts, I think you should get the information regardless of whether it was a service account or not.  If you then want to filter out service accounts, the client can easily do that by looking at the `service_account` key of the user data.  So I would drop this argument entirely and just always include service accounts in the result, similar to what `/users/<username>` currently does.

grouper/api/handlers.py
```diff
@@ -147,6 +155,34 @@ def get(self, name=None):
...
+
+        with self.graph.lock:
+            data = {}
+            for name in usernames:
+                try:
+                    data[name] = get_individual_user_info(
+                        self, name, cutoff,
+                        service_account=None if include_service_accounts else False)
```
rra: You can then set `service_account=None` here unconditionally.

tomeara-quora:
Updated to fix code review concerns, and added a test.

PR #332, opened 2017-11-15
Add support for SMTP servers that require logins

tomeara-quora:
Adds 3 new configuration options:
* smtp_use_ssl to force SSL on the SMTP connection
* smtp_username for the username to use when logging in
* smtp_password for the password to use when logging in

Support for servers that do not require authentication is supported by
leaving the smtp_username field blank.

### DIFF
--- a/grouper/settings.py
+++ b/grouper/settings.py
@@ -101,6 +101,9 @@ settings = Settings({
     "restricted_ownership_permissions": None,
     "send_emails": True,
     "sentry_dsn": None,
+    "smtp_password": "",
     "smtp_server": "localhost",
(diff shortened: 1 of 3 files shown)


grouper/settings.py
```diff
@@ -101,6 +101,9 @@ def default_settings_path():
     "restricted_ownership_permissions": None,
     "send_emails": True,
     "sentry_dsn": None,
+    "smtp_password": "",
     "smtp_server": "localhost",
+    "smtp_use_ssl": False,
+    "smtp_username": "root",
```
herb: can we make the default `""` as well to be consistent with the configs?

PR #333, opened 2017-11-15
Suppress exceptions in the background thread

tomeara-quora:
Currently, non-database related exceptions in the background thread are
reraised, terminating the background thread. This is nonideal as
temporal errors can result in the background thread ceasing to run,
requiring a server restart. This change adds logging for all unexpected
exceptions in the background thread and then suppresses the exception,
continuing to attempt to perform the background tasks.

### DIFF
--- a/grouper/background.py
+++ b/grouper/background.py
@@ -147,6 +147,8 @@ class BackgroundThread(Thread):
                 stats.set_gauge("successful-background-update", 0)
                 stats.set_gauge("failed-background-update", 1)
                 self.capture_exception()
-                raise
+                self.logger.exception(
+                    "Unexpected exception occurred while performing background tasks."
+                )
 
             sleep(60)


herb:
i'm somewhat opposed. if there's stuff that can fail temporarily we should be catching it, logging it, and continuing. otherwise an unexpected exception should tank the thread because the thrown exception could have consequences for the health of underlying libraries.

tomeara-quora:
Yeah I'm not a huge fan either. Ultimately it comes down to whether it's better to (potentially) be making bad calls in the background thread, or have the background thread not running and the main thread continue to serve requests even though the application as a whole is no longer in a good state.

One option could be to have the entire application die when the background thread exits, though I don't know if that's the most ideal solution.

herb:
> One option could be to have the entire application die when the background thread exits, though I don't know if that's the most ideal solution.

I'm more onboard with this at the end of the day. You'd rely on process management to restart and/or page. But probably not great in the interim. Any reason you can't monitor the `failed-background-update` and alert/page appropriately?

Also, would adding the temporary exceptions you're seeing be an option?

tomeara-quora:
Monitoring `failed-background-update` is certainly possible, but because the background thread has died the responding engineer has to restart the application anyways in order to restart the background thread.

I'm not concerned with any particular exceptions for this PR, and ideally all potentially transient exceptions would be able to be properly handled, but in practice I've found that's usually not done. 

That said, it is definitely an acceptable option in lieu of this change.

herb:
> because the background thread has died the responding engineer has to restart the application anyways in order to restart the background thread.

yeah, good point. i withdraw my previous misgivings in the interim and i would entertain a PR that just tanked the parent process if the thread dies.

and any temporary failures you can identify would also be a welcome addition should you want to pursue that.

tomeara-quora:
Closing in favor of https://github.com/dropbox/merou/pull/335

PR #334, opened 2017-11-16
Add new service account implementation

rra:
This adds a new type of service account, distinct from the role user
implementation, represented by a ServiceAccount object owned by a
Group.  These accounts use a different permission and management model:
all members of the owning group can manage the service account, and
the service account only has permissions directly granted to it.  Those
permissions must be held by the group that owns the service account.

Users now have a new service_account key in their metadata in the graph
that contains the service account metadata (description, machine set,
and owner).

Removal of permissions from the service account when they're removed
from the owning group is not yet implemented.

Deployment will require a database migration to add several new tables
and the new is_service_account column on the users table.

### DIFF
--- a/grouper/models/group_service_accounts.py
+++ b/grouper/models/group_service_accounts.py
@@ -15,7 +15,7 @@ class GroupServiceAccount(Model):
     __table_args__ = (
         Index(
             "group_service_account_idx",
-            "group_id", "service_account_pk",
+            "group_id", "service_account_id",
             unique=True
         ),
     )
@@ -25,11 +25,11 @@ class GroupServiceAccount(Model):
     group_id = Column(Integer, ForeignKey("groups.id"), nullable=False)
     group = relationship("Group", backref="service_accounts", foreign_keys=[group_id])
 
-    service_account_pk = Column(Integer, ForeignKey("service_accounts.id"), nullable=False)
+    service_account_id = Column(Integer, ForeignKey("service_accounts.id"), nullable=False)
     service_account = relationship("ServiceAccount", backref=backref("owner", uselist=False),
-        foreign_keys=[service_account_pk])
+        foreign_keys=[service_account_id])
 
     def __repr__(self):
         # type: () -> str
-        return "%s(group_id=%s, service_account_pk=%s)" % (
-            type(self).__name__, self.group_id, self.service_account_pk)
+        return "%s(group_id=%s, service_account_id=%s)" % (
+            type(self).__name__, self.group_id, self.service_account_id)
--- a/grouper/models/group_service_accounts.py
+++ b/grouper/models/group_service_accounts.py
@@ -1,0 +1,35 @@
+from sqlalchemy import Column, ForeignKey, Index, Integer
+from sqlalchemy.orm import backref, relationship
+
+from grouper.models.base.model_base import Model
+
+
+class GroupServiceAccount(Model):
+    """Service accounts owned by a group.
+
+    A group may own zero or more service accounts. This table holds the mapping between a Group and
+    the ServiceAccount objects it owns.
+    """
+
+    __tablename__ = "group_service_accounts"
+    __table_args__ = (
+        Index(
+            "group_service_account_idx",
+            "group_id", "service_account_pk",
+            unique=True
+        ),
+    )
+
+    id = Column(Integer, primary_key=True)
+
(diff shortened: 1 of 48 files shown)


rra:
In addition to the included tests, I also ran this locally and poked around a bunch in the UI, manipulating all of the pieces of a service account and adding and removing them.

grouper/models/service_account_permission_map.py
```diff
@@ -0,0 +1,93 @@
...
+            User.enabled == True,
+        )
+        for permission in permissions:
+            out[permission[1].service_account.user.username].append(ServiceAccountPermission(
+                permission=permission[0].name,
+                argument=permission[1].argument,
+                granted_on=permission[1].granted_on,
+                mapping_id=permission[1].id,
+            ))
+        return out
```
herb: similar comment about moving this to a biz logic module

grouper/models/group_service_accounts.py
```diff
@@ -0,0 +1,35 @@
...
+            unique=True
+        ),
+    )
+
+    id = Column(Integer, primary_key=True)
+
+    group_id = Column(Integer, ForeignKey("groups.id"), nullable=False)
+    group = relationship("Group", backref="service_accounts", foreign_keys=[group_id])
+
+    service_account_pk = Column(Integer, ForeignKey("service_accounts.id"), nullable=False)
```
herb: do you have a philosophy in your mind in regards to naming columns `_id` vs `_pk`? it seems we aren't entirely consistent as it is or at least i can't discern the pattern.

grouper/models/service_account_permission_map.py
```diff
@@ -0,0 +1,93 @@
...
+        )
+        out = []
+        for permission in permissions:
+            out.append(ServiceAccountPermission(
+                permission=permission[0].name,
+                argument=permission[1].argument,
+                granted_on=permission[1].granted_on,
+                mapping_id=permission[1].id
+            ))
+        return out
```
herb: similar request to move to business logic module

grouper/models/group_service_accounts.py
rra: I was just trying to copy our existing inconsistency. I'd be totally okay with just using `_id` everywhere; I think that would be better than whatever we're currently doing.

grouper/models/service_account_permission_map.py
rra: This, I don't think is business logic, but maybe I have a different definition? I think of business logic as anything that is higher-level than a direct implication of the schema. This code is a direct implication of the database schema: given the tables we're creating, this is the one and only one way that you would pull out a list of granted permissions from the database. So it's effectively a storage-layer API with no logic other than bundling together the data as a tuple. If we were to refactor the database schema without changing the business logic, this method would have to change but its output wouldn't.

Does that make sense?

grouper/models/service_account_permission_map.py
rra: Similar point here about this being a direct reflection of the schema and therefore a storage-level API.

grouper/models/service_account_permission_map.py
herb: i usually draw the line when code interacts with more than one db table. i think the real reason in this case is to avoid tangling the models further.

grouper/models/group_service_accounts.py
herb: yeah, i'd agree with using `_id` for new fields.

grouper/models/service_account.py
```diff
@@ -0,0 +1,59 @@
...
+
+    id = Column(Integer, primary_key=True)
+    description = Column(Text)
+    machine_set = Column(Text)
+    user_pk = Column(Integer, ForeignKey("users.id"), nullable=False)
+    user = relationship("User", uselist=False, backref=backref("service_account", uselist=False))
+
+    def __repr__(self):
+        # type: () -> str
+        return "<%s: id=%s user_pk=%s>" % (
```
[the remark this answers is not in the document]
rra: User and Group also use angle brackets, so I was trying to be consistent by using them.

rra:
I think I've addressed all review comments except for the angle bracket thing. Should be ready for another review.

grouper/models/service_account.py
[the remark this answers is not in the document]
herb: (╯°□°）╯︵ ┻━┻)

herb:
thanks for addressing all those things so quickly!

PR #339, opened 2017-11-23
Call user_created plugins for role users

rra:
user_created() plugin methods were only called for regular user
creation, not role user creation, essentially by accident.  The
just_created() method was only plugged into get_or_create() model
methods but not add().  Move it to add() so that it's always
called.

Provide an additional argument to the user_created() plugin method
flagging whether or not this is a service account (role user).


cbguder:
👍

PR #340, opened 2017-11-23
Remove obsolete User.set_metadata method

rra:
Fix the remaining caller and remove the obsolete method.

### DIFF
--- a/grouper/fe/handlers/user_shell.py
+++ b/grouper/fe/handlers/user_shell.py
@@ -6,6 +6,7 @@ from grouper.models.audit_log import AuditLog
 from grouper.models.user import User
 from grouper.role_user import can_manage_role_user
 from grouper.service_account import can_manage_service_account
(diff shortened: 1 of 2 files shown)


grouper/fe/handlers/user_shell.py
```diff
@@ -45,9 +45,7 @@ def post(self, user_id=None, name=None):
                 alerts=self.get_form_alerts(form.errors),
             )
 
-        user.set_metadata(USER_METADATA_SHELL_KEY, form.data["shell"])
-        user.add(self.session)
-        self.session.commit()
+        set_user_metadata(self.session, user.id, USER_METADATA_SHELL_KEY, form.data["shell"])
```
cbguder: Looks like you're missing an import here

rra:
Yeah, resolved a merge conflict incorrectly. This diff keeps conflicting with every other diff I land. Will be fixed in just a second.

PR #341, opened 2017-11-29
Add plugin support for verifying machine_set

rra:
New service accounts and edited service accounts will now run their
machine set through a plugin function, if defined, and give it the
opportunity to reject that machine set with an exception.

### DIFF
--- a/grouper/service_account.py
+++ b/grouper/service_account.py
@@ -24,6 +24,7 @@ from grouper.models.permission import Permission
 from grouper.models.service_account import ServiceAccount
 from grouper.models.service_account_permission_map import ServiceAccountPermissionMap
 from grouper.models.user import User
+from grouper.plugin import get_plugins, PluginRejectedMachineSet
 from grouper.user import disable_user, enable_user
 
 # A single service account permission.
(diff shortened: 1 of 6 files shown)


grouper/service_account.py
```diff
@@ -24,30 +24,54 @@
...
+    """Verify a service account machine set with plugins.
+
+    Raises:
+        BadMachineSet: if some plugin rejected the machine set
+    """
+    try:
+        for plugin in get_plugins():
+            plugin.check_machine_set(service_account.user.username, machine_set)
+    except PluginRejectedMachineSet as e:
+        raise BadMachineSet(str(e))
```
cbguder: for `PluginRejectedGroupMembershipUpdate` and `PluginRejectedDisablingUser` we bubbled the exception all the way up. this is a little more overhead but I like the separation. perhaps we should do the same for the other plugin exceptions.

tests/test_service_accounts.py
```diff
@@ -262,3 +265,70 @@ def test_service_account_fe_perms(session, standard_graph, http_client, base_url
...
+    resp = yield http_client.fetch(fe_url, method="POST", headers={"X-Grouper-User": admin},
+            body=urlencode(data))
+    assert resp.code == 200
+    graph.update_from_db(session)
+    metadata = graph.user_metadata["other@svc.localhost"]
+    assert metadata["service_account"]["description"] == "some other service account"
+    assert metadata["service_account"]["machine_set"] == "is okay"
+
+    # Reset for any subsequent tests.
+    grouper.plugin.Plugins.pop()
```
cbguder: In [`test_group_ownership_policy.py`](https://github.com/dropbox/merou/blob/master/tests/test_group_ownership_policy.py) and other places we patch `get_plugins` which avoids this at the expense of a different kind of implementation detail leaking into the test. I would like to clean this up and get rid of the global variable at some point.

grouper/service_account.py
rra: Yeah, >>TODO.

tests/test_service_accounts.py
rra: Indeed, I have on my list already to abstract out a better way of testing plugins so that we can just reuse a single pattern with more shared machinery.

PR #345, opened 2017-12-05
Skip Selenium tests if not available

rra:
If either the Python selenium module or the necessary webdriver
is not available, skip the Selenium tests.  This allows running
the test suite in integration test environments where Selenium
is not available.

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -1,7 +1,4 @@
 import pytest
-from selenium import webdriver
-import subprocess
-import yaml
 
 from grouper.api.routes import HANDLERS as API_HANDLERS
 from grouper.app import Application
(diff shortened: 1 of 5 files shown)


cbguder:
I'm worried this might lead to a situation where the Selenium tests quietly stop running on CI. Would it be too messy to separate them into a different folder?

rra:
Hm, yeah, we could do that.

My goal is to get the tests running inside Changes (I know we run Selenium tests there, but the setup for that is very inobvious to me and just getting the non-Selenium tests running in server would be very useful). In that environment, the selenium module won't be available at all, so I could remove the guards around the web driver setup. Then the tests would only be skipped if the selenium module were missing entirely, which seems like an unlikely problem to have (without also causing `pip install` to fail). That would save on the complexity of breaking everything out.

rra:
Merging this after further in-person discussion.

PR #349, opened 2017-12-07
Add grouper-ctl dump_sql command

rra:
Just walks the model and dumps the SQL to create the tables in
the configured engine.  This is enough to manually generate a
database migration with some confidence that the table SQL will be
correct, without the overhead of setting up a proper migration
management system.

### DIFF
--- a/grouper/ctl/dump_sql.py
+++ b/grouper/ctl/dump_sql.py
@@ -1,0 +1,24 @@
+from typing import TYPE_CHECKING
+
+from sqlalchemy.schema import CreateTable
+
+from grouper.models.base.model_base import Model
+from grouper.models.base.session import get_db_engine
+from grouper.settings import settings
+from grouper.util import get_database_url
+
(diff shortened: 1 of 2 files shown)


grouper/ctl/dump_sql.py
```diff
@@ -0,0 +1,25 @@
...
+def dump_sql_command(args):
+    # type: (argparse.Namespace) -> None
+    db_engine = get_db_engine(get_database_url(settings))
+    for table in Model.metadata.sorted_tables:
+        print CreateTable(table).compile(db_engine)
+
+
+def add_parser(subparsers):
+    # type: (argparse._SubParsersAction) -> None
+    print type(subparsers)
```
cbguder: Looks like this got left behind, causing the test failure

rra: Whoops, thank you!  Finished this relatively late last night and didn't check.

PR #359, opened 2018-01-04
Move all code into libraries

rra:
Best practice for testable code is to put as little as possible
in the command wrapper, since one has to go to special effort to
load and run that code.  Following the pattern for grouper-ctl,
move the main() routines of grouper-api and grouper-fe into the
library.

No tests currently take advantage of this (the problem I thought
I had turned out to be something else), but it lays the groundwork
for later test work.


cbguder:
Ugh, looks like `git grep` exits with `1` when it finds no results. We can probably get rid of everything related to `bin` in `mypy.sh` now that those files contain very little code.

PR #363, opened 2018-01-11
Skip test_bin tests if bin/* was not found

rra:
test_bin.py assumes the tests are being run from the root of a
Marou source distribution, but when run under Bazel, the current
working directory can be somewhere else entirely.  Skip these
tests if the files can't be found in `bin/` relative to the current
working directory.  (These tests essentially just check the imports
in the `bin/*` scripts, so it's no great loss if they're skipped.)

### DIFF
--- a/tests/test_bin.py
+++ b/tests/test_bin.py
@@ -1,16 +1,21 @@
 import subprocess
 
+from path_util import src_path
+
 
 def test_api():
-    out = subprocess.check_output(["bin/grouper-api", "--help"])
+    bin_path = src_path("bin", "grouper-api")
+    out = subprocess.check_output([bin_path, "--help"])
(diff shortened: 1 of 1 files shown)


cbguder:
Can we not find the absolute path to these files? If there's no clean way to do that, we can also change these tests to just call `main()` since there's no other code in `bin/*` (provided we agree to keep it that way, of course).

rra:
Oh, good call, I forgot about `path_util` and that approach.  How does this look?  (Unfortunately, it's rather hard to test whether this works in Bazel without merging and then updating.  I wish we had a better solution for that.)

cbguder:
Yeah, that is rather unfortunate, but it looks like it should work. This looks great to me, thanks!

PR #367, opened 2018-01-31
Increase default grouper-ctl log level to INFO

rra:
`grouper-ctl oneoff list` logs the list of modules at INFO level, so
if the log level isn't at least INFO, that command silently does
nothing.  Change the default log level for grouper-ctl to INFO
since it's interactive without changing the default for other
programs, by allowing `get_log_level` to take an optional base log
level.


rra:
Yeah, agreed. I'll create a task for that.

PR #368, opened 2018-01-31
Fix permission argument checking

rra:
Permission argument checking was a bit of a mess.  We validated
arguments when granting permissions directly, but not when creating
a request for a permission.  The backup regex checking the argument
inside the API only checked that the argument started with a valid
string, not consisted solely of a valid string.

Add validators when requesting a permission, anchor the regex in
the assertions in the API, and add tests to ensure we're properly
validating permission arguments.

Allow spaces in permission arguments, since we were already using
them (via requested permissions).

### DIFF
--- a/grouper/permissions.py
+++ b/grouper/permissions.py
@@ -63,7 +63,8 @@ def grant_permission(session, group_id, permission_id, argument=''):
     Throws:
         AssertError if argument does not match ARGUMENT_VALIDATION regex
     """
-    assert re.match(ARGUMENT_VALIDATION, argument), 'Permission argument does not match regex.'
+    assert re.match(ARGUMENT_VALIDATION + r"$", argument), \
+        'Permission argument does not match regex.'
 
     mapping = PermissionMap(permission_id=permission_id, group_id=group_id, argument=argument)
     mapping.add(session)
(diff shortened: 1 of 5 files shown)


grouper/permissions.py
```diff
@@ -63,7 +63,8 @@ def grant_permission(session, group_id, permission_id, argument=''):
     Throws:
         AssertError if argument does not match ARGUMENT_VALIDATION regex
     """
-    assert re.match(ARGUMENT_VALIDATION, argument), 'Permission argument does not match regex.'
+    assert re.match(ARGUMENT_VALIDATION + r"$", argument), \
```
cbguder: These should probably have `r"^"` prepended as well, right?

tests/test_permissions.py
```diff
@@ -233,15 +246,26 @@ def test_permission_request_flow(session, standard_graph, groups, grantable_perm
...
+    emails = _get_unsent_and_mark_as_sent_emails(session)
+    assert len(emails) == 0, "no emails queued"
+
+    # REQUEST: 'grantable.one', 'some argument' for 'serving-team'
     resp = yield http_client.fetch(fe_url, method="POST",
             body=urlencode({"permission_name": "grantable.one", "argument": "some argument",
                 "reason": "blah blah black sheep", "argument_type": "text"}),
             headers={'X-Grouper-User': username})
     assert resp.code == 200
+    print resp.body
```
cbguder: Leftover `print`? #noprintleftbehind

grouper/permissions.py
rra: `re.match` only matches at the start of the string (as opposed to `re.search`). It's one of the weird things about Python regexes.

tests/test_permissions.py
rra: Good catch.  Fixed!

PR #374, opened 2018-02-09
api: Don't query for User object when validating tokens

lfaraone:


lfaraone:
We could also just keep a local cache, but I sort of want the database roundtrip so revocation is quick.

grouper/models/user_token.py
```diff
@@ -34,6 +35,11 @@ class UserToken(Model):
         UniqueConstraint("user_id", "name"),
     )
 
+    @staticmethod
+    def get_by_value(session, username, name):
+        print username, name
```
cbguder: Stray print here

PR #376, opened 2018-02-09
List instance ID in API stats as well

cbguder:
D'oh!


grouper/api/main.py
```diff
@@ -3,8 +3,10 @@
 import sys
 from typing import TYPE_CHECKING
 
+from expvar.stats import stats
```
lfaraone: or we could be even more opaque and do `from expvar.stats.stats import set_label` jk

PR #381, opened 2018-03-26
Bump requirement versions to match Dropbox

rra:
We're testing Merou internally with a specific pinned set of
versions, so sync all the dependency versions with requirements
so that both test environments test the same thing as much as
possible.  Only do this when it pushes dependencies forward.

We only test on Python 2.7, so remove argparse from requirements.txt
since it's only required on Python 2.6.

Bump mypy to the latest release, which is closer to what we do
internally at Dropbox.


rra:
We do that too now!  But this way our public requirements match what we use internally, which makes it more likely everything will work properly. (I probably should have said that in the diff...)

PR #382, opened 2018-03-26
Add missing frontend length validators

rra:
The permission argument is currently restricted to 64 characters,
and the user token name is restricted to 16 characters.  Add those
restrictions directly to the form validation, since MySQL will
otherwise accept invalid data and silently truncate it.

Note that the permission *request* table limits arguments to 64
characters, but the permission *map* table allows 128 characters,
so the limits are inconsistent between grant and request.  This will
be fixed with a later commit but will require a database migration.

Also comment all of the unconstrained choice fields that I manually
verified did provide choices at runtime.

### DIFF
--- a/grouper/fe/forms.py
+++ b/grouper/fe/forms.py
@@ -125,6 +125,7 @@ class AuditCreateForm(Form):
 
 
 class GroupRequestModifyForm(Form):
+    # Caller of the form will add choices based on current status.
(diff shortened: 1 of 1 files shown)


rra:
Follow-on work to true up the length limits is at https://github.com/rra/merou/tree/args-truncate. I'll rebase and open a pull request when this lands.

PR #383, opened 2018-03-27
Increase permission argument length in request

rra:
Permission arguments were allowed to be MAX_NAME_LENGTH (128
characters) in the permission_map table, but the request was still
capped at 64 characters.  Make these consistent.

Add a test, although since SQLite doesn't check any column
constraints the test will never fail unless you run the test against
a MySQL database.

### DIFF
--- a/grouper/fe/forms.py
+++ b/grouper/fe/forms.py
@@ -214,7 +214,7 @@ class GroupPermissionRequestDropdownForm(Form):
         "Argument",
         [
             validators.DataRequired(),
-            validators.Length(min=0, max=64),
+            validators.Length(min=0, max=constants.MAX_NAME_LENGTH),
             ValidateRegex(constants.ARGUMENT_VALIDATION),
         ],
     )
@@ -231,7 +231,7 @@ class GroupPermissionRequestTextForm(Form):
         "Argument",
         [
             validators.DataRequired(),
-            validators.Length(min=0, max=64),
+            validators.Length(min=0, max=constants.MAX_NAME_LENGTH),
             ValidateRegex(constants.ARGUMENT_VALIDATION),
         ],
     )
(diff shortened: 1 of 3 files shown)


rra:
I'm not super-excited about increasing the length of all names, since I think people are doing themselves a disservice by giving a group or a user that long of name. I have a hard time thinking of cases where you'd really want that, and you might even increase the chances of names that downstream systems can't handle.

If we want to make it larger, I'd be more inclined to introduce a new constant, `MAX_ARGUMENT_LENGTH`, and change all the code that manipulates permission arguments to use that instead.

cbguder:
Fair enough. I would also prefer to enforce length constraints in the application and not in the database, but that's even more work. Let's merge this and worry about that if we run into length issues again.

rra:
+100 on enforcing length constraints in the application instead of the database, particularly since SQLite ignores all database schemas and therefore our tests won't catch any of these failures. (A friend of mine recommended using `pg_virtualenv` for any tests of database applications instead since PostgreSQL actually implements the SQL standard, unlike SQLite and MySQL. Maybe worth considering in the future.)

I'm not sure there's any point in having any of the database fields be smaller than 255 characters. Maybe performance? I'm not sure whether the size of columns actually matters for modern databases, although I assume you don't want to make fields that you actually want to index into `Text` types.

cbguder:
Yep. For `VARCHAR` columns there should be no difference between 128 and 255, but I agree we probably don't want `TEXT`, and I definitely like PostgreSQL =)

rra:
FYI, this is stalled pending us figuring out a database migration strategy.

akorzo-dbx:
Closing as it's outdated and code changed since. The following PR takes care of this:

https://github.com/dropbox/merou/pull/676

PR #396, opened 2018-05-21
Subresource integrity & CSP

rickmark:


grouper/fe/template_util.py
```diff
@@ -7,6 +10,29 @@
 from grouper.fe.settings import settings
 
 
+RESOURCE_INTEGRITY_VALUES = {}
+CHUNK_SIZE = 1024 * 16
+
+
+def get_integrity_of_static(path):
```
lfaraone: Please cache this.

grouper/fe/templates/service.html
```diff
@@ -291,34 +298,3 @@ <h4 class="modal-title">Audit Group</h4>
 {% endif %}
 
 {% endblock %}
-
-{% block script %}
-{# The removeUserModal is generated once per page but could be used for any
```
lfaraone: Preserve this comment mayhaps.

grouper/fe/templates/service.html
```diff
@@ -238,8 +238,15 @@ <h4 class="modal-title">Remove User</h4>
     {% endif %}
 {% endmacro %}
 
-<div class="modal fade" id="auditModal" tabindex="-1" role="dialog"
-      aria-labelledby="auditModal" aria-hidden="true">
+<div class="modal fade"
```
lfaraone: Unneeded churn?

grouper/fe/template_util.py
```diff
@@ -1,6 +1,6 @@
+from base64 import b64encode
 from datetime import datetime
 from hashlib import sha384
-from base64 import b64encode
```
lfaraone: rebase and fixup this into the commit that introduced it

grouper/fe/templates/service.html
rickmark: The tag had to change due to the introduction of the UJS data-* fields anyway

grouper/fe/template_util.py
rickmark: It is cached in the `RESOURCE_INTEGRITY_VALUES` hash

grouper/fe/templates/service.html
rickmark: Moved to JS comment

lfaraone:
Please rebase rather than creating a merge commit as in 2c7b111381af3d66fc1b930d59ede5067b6517bb

grouper/fe/template_util.py
```diff
@@ -11,17 +11,24 @@
 
 
 RESOURCE_INTEGRITY_VALUES = {}
+CHUNK_SIZE = 1024 * 16
```
lfaraone: nit, but consider instead:
``` py
try:
    CHUNK_SIZE = os.statvfs(os.curdir).f_frsize
except AttributeError:
    # statvfs is unix only, fall back to sensible default
    CHUNK_SIZE = 4096 
```

fundamentally these files are tiny so it doesn't really matter.

config/dev.yaml
```diff
@@ -114,7 +114,7 @@ fe:
     # This is the prefix url in template to cdnjs. This allows you to point at
     # and internal mirror for production deployments.
     # Type: str
-    cdnjs_prefix: "//cdnjs.cloudflare.com"
+    cdnjs_prefix: "http://cdnjs.cloudflare.com"
```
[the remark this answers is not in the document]
rickmark: The problem is that CSP doesn't allow for scheme agnostic URLs.  So when we craft the CSP it needs to know the correct scheme (which of course we never really know because it may be TLS terminated before...)

What do you suggest?

grouper/fe/util.py
```diff
@@ -72,6 +72,54 @@ def initialize(self):
...
+        ]
+
+        policy = "default-src 'none'; img-src 'self'; style-src 'self' 'unsafe-inline'"
+
+        for item in style_items:
+            policy += " {cdn_url}{item}".format(cdn_url=settings["cdnjs_prefix"], item=item)
+
+        policy += "; script-src 'self'"
+
+        for item in script_items:
```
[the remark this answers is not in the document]
rickmark: If we introduce a CSP nonce, it may not work well if there's a reverse proxy before it that generates such nonces.

config/dev.yaml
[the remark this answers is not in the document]
lfaraone: Always serve the resource over HTTPS. It is fine to load HTTPS JS from an insecure origin.

grouper/fe/util.py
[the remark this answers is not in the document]
lfaraone: The RP in question exposes that nonce to you as an HTTP header :)

rra:
Resolved by https://github.com/dropbox/merou/pull/520, https://github.com/dropbox/merou/pull/523, https://github.com/dropbox/merou/pull/531, and https://github.com/dropbox/merou/pull/533.

PR #400, opened 2018-07-09
Add integration tests for API endpoints

cbguder:
Tests will pass once #399 is merged.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,10 +1,11 @@
+coverage==3.7.1
+flake8-import-order==0.8
+flake8==2.5.4
+groupy
+mock==2.0.0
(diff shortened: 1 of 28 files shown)


requirements-dev.txt
```diff
@@ -8,3 +8,4 @@ mock==2.0.0
 flake8==2.5.4
 flake8-import-order==0.8
 selenium==3.6.0
+groupy
```
herb: shouldn't we pin the version? and sort alphabetically?

cbguder: Yeah, we should sort alphabetically. As for the version, I intentionally left that unpinned so tests always run against the latest stable release.

PR #401, opened 2018-07-11
Expiration date update for group request pages

bonniech3n:
Updating display of expiration date to be more aligned with way we represent other date fields.
- Updated template_util for date format handler
- Update the FE for displaying on group requests and individual group request action page

Tested:
- created user, request to join group
- logged in as admin, view group, view request pages, view individual request page, action
- Auto-expiry shows up when date is passed (log at the bottom of page)

### DIFF
--- a/grouper/fe/template_util.py
+++ b/grouper/fe/template_util.py
@@ -23,7 +23,9 @@ def _make_date_obj(input_date_obj):
     elif isinstance(input_date_obj, datetime):
         date_obj = input_date_obj
     else:
-        assert False, "input_date_obj {} is of unsupported type {}".format(input_date_obj, type(input_date_obj))
+        assert False, "input_date_obj {} is of unsupported type {}".format(
+            input_date_obj, type(input_date_obj)
+        )
 
     assert isinstance(date_obj, datetime)
 
--- a/grouper/fe/template_util.py
+++ b/grouper/fe/template_util.py
@@ -15,7 +15,7 @@ def _make_date_obj(input_date_obj):
     object."""
     if isinstance(input_date_obj, float):
         date_obj = datetime.fromtimestamp(input_date_obj, UTC)
-    elif isinstance(input_date_obj, basestring):
+    elif isinstance(input_date_obj, string_types):
         try:
             date_obj = datetime.strptime(input_date_obj, "%m/%d/%Y")
         except ValueError:
--- a/grouper/fe/template_util.py
+++ b/grouper/fe/template_util.py
@@ -23,7 +23,7 @@ def _make_date_obj(input_date_obj):
     elif isinstance(input_date_obj, datetime):
         date_obj = input_date_obj
     else:
-        assert False, "{}>>>{}".format(input_date_obj, type(input_date_obj))
+        assert False, "input_date_obj {} is of unsupported type {}".format(input_date_obj, type(input_date_obj))
 
     assert isinstance(date_obj, datetime)
 
--- a/grouper/fe/template_util.py
+++ b/grouper/fe/template_util.py
@@ -51,7 +51,7 @@ def print_date(input_date):
     if input_date is None or input_date == "" or isinstance(input_date, Undefined):
         return ""
 
-    date_obj = _make_date_obj(date_obj)
+    date_obj = _make_date_obj(input_date)
 
     return date_obj.strftime(settings["date_format"])
 
--- a/grouper/fe/template_util.py
+++ b/grouper/fe/template_util.py
@@ -2,6 +2,7 @@ from datetime import datetime
 
 from dateutil.relativedelta import relativedelta
 from jinja2 import Environment, PackageLoader
+from jinja2.runtime import Undefined
 from pytz import UTC
 from six import string_types
 
@@ -22,7 +23,7 @@ def _make_date_obj(input_date_obj):
     elif isinstance(input_date_obj, datetime):
         date_obj = input_date_obj
     else:
-        assert False, '{}>>>{}'.format(input_date_obj, type(input_date_obj))
+        assert False, "{}>>>{}".format(input_date_obj, type(input_date_obj))
 
     assert isinstance(date_obj, datetime)
 
@@ -47,7 +48,7 @@ def print_date(input_date):
 
     Returns human readable date/time string.
     """
-    if input_date is None or input_date == '' or isinstance(input_date, Undefined):
+    if input_date is None or input_date == "" or isinstance(input_date, Undefined):
         return ""
 
     date_obj = _make_date_obj(date_obj)
--- a/grouper/fe/template_util.py
+++ b/grouper/fe/template_util.py
@@ -8,16 +8,21 @@ from six import string_types
 from grouper.fe.settings import settings
 
 
-def _make_date_obj(date_obj):
+def _make_date_obj(input_date_obj):
     """Given either a datetime object, float date/time in UTC unix epoch, or
     string date/time in the form '%Y-%m-%d %H:%M:%S.%f' return a datetime
     object."""
-    if isinstance(date_obj, float):
-        date_obj = datetime.fromtimestamp(date_obj, UTC)
-
-    if isinstance(date_obj, string_types):
-        date_obj = datetime.strptime(date_obj, "%Y-%m-%d %H:%M:%S.%f")
-        date_obj = date_obj.replace(tzinfo=UTC)
+    if isinstance(input_date_obj, float):
+        date_obj = datetime.fromtimestamp(input_date_obj, UTC)
+    elif isinstance(input_date_obj, basestring):
+        try:
+            date_obj = datetime.strptime(input_date_obj, "%m/%d/%Y")
+        except ValueError:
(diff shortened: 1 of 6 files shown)


grouper/fe/template_util.py
```diff
@@ -11,22 +11,27 @@ def _make_date_obj(date_obj):
...
+    if date_obj == '' or date_obj is None:
+        return ''
+
     if isinstance(date_obj, float):
         date_obj = datetime.fromtimestamp(date_obj, UTC)
 
-    if isinstance(date_obj, basestring):
-        date_obj = datetime.strptime(date_obj, "%Y-%m-%d %H:%M:%S.%f")
-        date_obj = date_obj.replace(tzinfo=UTC)
+    if isinstance(date_obj, unicode) or isinstance(date_obj, basestring):
```
cbguder: `unicode` is a subclass of `basestring`, so `isinstance(date_obj, basestring)` should be enough

grouper/fe/template_util.py
```diff
@@ -11,22 +11,27 @@ def _make_date_obj(date_obj):
     """Given either a datetime object, float date/time in UTC unix epoch, or
     string date/time in the form '%Y-%m-%d %H:%M:%S.%f' return a datetime
     object."""
+    if date_obj == '' or date_obj is None:
```
cbguder: I think this check should happen somewhere else, probably `print_date`? Otherwise this method returns a datetime *or* a string.

grouper/fe/template_util.py
```diff
@@ -11,22 +11,27 @@ def _make_date_obj(date_obj):
...
+        return ''
+
     if isinstance(date_obj, float):
         date_obj = datetime.fromtimestamp(date_obj, UTC)
 
-    if isinstance(date_obj, basestring):
-        date_obj = datetime.strptime(date_obj, "%Y-%m-%d %H:%M:%S.%f")
-        date_obj = date_obj.replace(tzinfo=UTC)
+    if isinstance(date_obj, unicode) or isinstance(date_obj, basestring):
+        date_obj = str(date_obj)
```
cbguder: `strptime` can handle unicode, so no need to cast here

grouper/fe/template_util.py
```diff
@@ -11,22 +11,27 @@ def _make_date_obj(date_obj):
...
+        try:
+            date_obj = datetime.strptime(date_obj, "%m/%d/%Y")
+        except:
```
cbguder: Let's avoid a generic `except` here, and catch only what `strptime` raises, which I believe is `ValueError`.

grouper/fe/template_util.py
```diff
@@ -11,22 +11,27 @@ def _make_date_obj(date_obj):
...
+            date_obj = datetime.strptime(date_obj, "%Y-%m-%d %H:%M:%S.%f")
 
-    assert isinstance(date_obj, datetime)
+    if date_obj:
```
cbguder: By the time we get here `date_obj` should never be None, so we can probably keep the assertion the same

grouper/fe/template_util.py
```diff
@@ -11,22 +11,27 @@ def _make_date_obj(date_obj):
...
+        assert isinstance(date_obj, datetime)
 
     if date_obj.tzinfo is None:
-        # naive dates are assumed UTC
```
cbguder: Let's keep this comment

grouper/fe/template_util.py
```diff
@@ -41,14 +46,17 @@ def print_date(date_obj):
...
 
-    date_obj = date_obj.astimezone(settings["timezone"])
-    return date_obj.strftime(settings["date_format"])
-
+    if date_obj:
+        assert isinstance(date_obj, datetime)
+        date_obj = date_obj.astimezone(settings["timezone"])
+        return date_obj.strftime(settings["date_format"])
+    else:
+        return ""
```
cbguder: This check would be unnecessary if `_make_date_obj` always returns a `datetime`.

grouper/fe/templates/group-request-update.html
```diff
@@ -28,6 +29,9 @@
                     <td>{{ update.from_status|default("", True) }}</td>
                     <td>{{ update.to_status }}</td>
                     <td>{{ update.change_at|print_date }}</td>
+                    {% if request.changes.expiration is defined %}
```
cbguder: I think we always want the `<td>` to be there, even if there's no text in it. Also, can we add/update an itest for this?

grouper/fe/templates/group-requests.html
```diff
@@ -48,7 +48,12 @@
                         <td class="request-requester">{{ request.requester }}</td>
                         <td class="request-status">{{ request.status }}</td>
                         <td class="request-requested-at">{{ request.requested_at|print_date }}</td>
-                        <td class="request-expiration">{{ request.changes['expiration'] }}</td>
+                        {% if request.changes.expiration == '' or request.changes.expiration is none or request.changes.expiration is not defined%}
```
cbguder: Can we simplify this, maybe to something like `request.changes.get('expiration')` and let `print_date` handle the `None`?

grouper/models/group_edge.py
```diff
@@ -128,10 +129,13 @@ def apply_changes(self, changes):
                                       group_name,
                                       member_name)
                 if value:
-                    expiration = datetime.strptime(value, "%m/%d/%Y")
```
cbguder: Let's not touch this logic in this diff.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 annex==0.4.0
 Jinja2==2.7.3
-MySQL-python==1.2.5
+# MySQL-python==1.2.5
```
cbguder: We probably don't want to commit this

grouper/models/request.py
```diff
@@ -65,6 +65,7 @@ def my_status_updates(self):
             RequestStatusChange.to_status,
             label("changed_by", User.username),
             label("reason", Comment.comment)
+
```
cbguder: Accidental whitespace?

grouper/models/group_edge.py
```diff
@@ -115,10 +116,12 @@ def apply_changes(self, changes):
...
                     cancel_expiration(self.session, group_name, member_name)
                 if value:
-                    expiration = datetime.strptime(value, "%m/%d/%Y")
+                    expiration = _make_date_obj(value)
+                    assert isinstance(expiration, datetime)
                     setattr(self, key, expiration)
                     # Avoid sending notifications for expired edges.
-                    if expiration > datetime.utcnow():
+                    if expiration > _utcnow():
+                        assert expiration > _utcnow()
```
karthik-shanmugam: this assert looks pointless

grouper/fe/template_util.py
```diff
@@ -2,22 +2,28 @@
...
+        date_obj = datetime.fromtimestamp(input_date_obj, UTC)
+    elif isinstance(input_date_obj, string_types):
+        try:
+            date_obj = datetime.strptime(input_date_obj, "%m/%d/%Y")
+        except ValueError:
+            date_obj = datetime.strptime(input_date_obj, "%Y-%m-%d %H:%M:%S.%f")
+    elif isinstance(input_date_obj, datetime):
+        date_obj = input_date_obj
+    else:
+        assert False, "{}>>>{}".format(input_date_obj, type(input_date_obj))
```
karthik-shanmugam: I would add some text to this message that makes is more descriptive, maybe something like "date {} is of unsupported type {}"

grouper/fe/templates/group-requests.html
```diff
@@ -48,7 +48,9 @@
                         <td class="request-requester">{{ request.requester }}</td>
                         <td class="request-status">{{ request.status }}</td>
                         <td class="request-requested-at">{{ request.requested_at|print_date }}</td>
-                        <td class="request-expiration">{{ request.changes['expiration'] }}</td>
+                        {% if request.changes.get('expiration')%}
+                        <td class="request-expiration">{{ request.changes.expiration|print_date }}</td>
```
karthik-shanmugam: nit: indentation

grouper/models/group_edge.py
```diff
@@ -115,10 +116,12 @@ def apply_changes(self, changes):
                     # Check for and remove pending expiration notification.
                     cancel_expiration(self.session, group_name, member_name)
                 if value:
-                    expiration = datetime.strptime(value, "%m/%d/%Y")
+                    expiration = _make_date_obj(value)
+                    assert isinstance(expiration, datetime)
```
karthik-shanmugam: I don't think this assert is necessary; if `_make_date_obj` promises to always return a datetime we don't need to double check it.

karthik-shanmugam (changes requested):
Looks mostly good, couple small nits

grouper/models/group_edge.py
```diff
@@ -4,6 +4,7 @@
 from sqlalchemy.orm import relationship
 
 from grouper.expiration import add_expiration, cancel_expiration
+from grouper.fe.template_util import _make_date_obj, _utcnow
```
cbguder: If we're using these methods here, let's move them elsewhere. Importing front-end code in a model is backwards.

PR #407, opened 2018-08-28
Move all JavaScript into UJS to prevent needing unsafe-script

rickmark:
This supersedes #396 and breaks apart the changes for readability.


itests/fixtures.py
```diff
@@ -120,4 +126,5 @@ def wait_until_accept(port, timeout=3.0):
                 raise
         else:
             s.close()
+            time.sleep(0.25)
```
karthik-shanmugam: maybe add a comment explaining why we did this

grouper/fe/static/js/grouper.js
```diff
@@ -19,4 +19,171 @@ $(document).ready(function(){
...
+            // update view of the world
+            update_args();
+        });
+
+        $reason_fields.keyup(function() {
+            // keep reason values synced
+            $reason_fields.val($(this).val());
+        });
+    }
+});
```
karthik-shanmugam: is no newline at the end of js standard?

PR #408, opened 2018-08-31
Inline style removal for stricter CSP

rickmark:


itests/fixtures.py
```diff
@@ -12,15 +11,35 @@
 
 
 def _get_unused_port():
-    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
-        s.bind(('', 0))
-        return s.getsockname()[1]
+    return _get_unused_ports(1)[0]
+
+
+def _get_unused_ports(count):
```
rra: I think Karthik fixed the itest failures, and this looks wrong anyway (it allocates the sockets and then closes them, which means that something else could still take the socket before it's opened again). Could you remove this code, or at least put it into a different PR instead of mixing it into this one?

itests/fixtures.py
```diff
@@ -107,7 +126,7 @@ def api_client(async_api_server):
     return Groupy(async_api_server)
 
 
-def wait_until_accept(port, timeout=3.0):
+def wait_until_accept(port, timeout=10.0):
```
rra: This should be a separate PR, since it's unrelated to the rest of the change.

ci/setup.sh
```diff
@@ -3,7 +3,7 @@
 set -eux
 
 if [[ "$TRAVIS_PYTHON_VERSION" == 2* ]]; then
-  wget "https://chromedriver.storage.googleapis.com/2.37/chromedriver_linux64.zip"
+  wget "https://chromedriver.storage.googleapis.com/2.41/chromedriver_linux64.zip"
```
rra: Could you move this to a separate PR?

PR #410, opened 2018-09-04
Add audits to menu if user has permissions

rickmark:


grouper/fe/util.py
```diff
@@ -189,6 +192,7 @@ def get_template_namespace(self):
             "perf_trace_uuid": self.perf_trace_uuid,
             "xsrf_form": self.xsrf_form_html,
             "alerts": self.get_alerts(),
+            "current_user_has_audit": self.current_user_has_audit,
```
karthik-shanmugam: Is there a difference between calling these functions here and calling them from the template? I see `get_alerts` is the only one being called here.

rickmark: Passing it allows the code to only run if the result is of interest.  If I evaluate it, then every page will take a perf hit.

karthik-shanmugam: It looks like this is going to be run on every page regardless. Also, each call to `user_has_permissions` does a database query. Should we concerned about this?

rra: Yeah, I think this is going to be pretty slow. It wouldn't be as bad if `user_permissions` was changed per the `FIXME` comment to use the graph instead, so I think that's a prerequisite.

rra:
There are performance concerns with this. We can look at this later when we go back to investigate how to handle audits.

PR #412, opened 2018-09-21
Inline style for stricter CSP

rickmark:


grouper/fe/static/css/grouper.css
```diff
@@ -287,6 +287,48 @@ span.disabled {
...
+}
+
+.fingerprint-padding {
+    padding: 15px 0px;
+}
+
+.standard-panel {
+    max-height: 390px;
+    overflow-y: auto;
+}
```
karthik-shanmugam: nit: newline

PR #414, opened 2018-10-15
crash grouper on db_url change

karthik-shanmugam:
- Useful for forcing grouper processes to restart when a database promotion occurs

### DIFF
--- a/grouper/background/background_processor.py
+++ b/grouper/background/background_processor.py
@@ -1,11 +1,11 @@
 from contextlib import closing
 from datetime import datetime, timedelta
 import logging
+import os
 from time import sleep
 from typing import TYPE_CHECKING
 
 from sqlalchemy import and_
-from sqlalchemy.exc import OperationalError
 
 from grouper import stats
 from grouper.constants import PERMISSION_AUDITOR
@@ -16,7 +16,7 @@ from grouper.email_util import (
 )
 from grouper.graph import Graph
 from grouper.group import get_audited_groups
-from grouper.models.base.session import get_db_engine, Session
+from grouper.models.base.session import Session
 from grouper.models.group import Group
 from grouper.models.group_edge import APPROVER_ROLE_INDICES, GroupEdge
 from grouper.models.user import User
@@ -48,6 +48,9 @@ class BackgroundProcessor(object):
     def _capture_exception(self):
         if self.sentry_client:
             self.sentry_client.captureException()
+
+    def crash(self):
+        os._exit(1)
 
     def expire_edges(self, session):
         # type: (Session) -> None
@@ -120,8 +123,11 @@ class BackgroundProcessor(object):
 
     def run(self):
         # type: () -> None
+        initial_url = get_database_url(self.settings)
         while True:
             try:
+                if get_database_url(self.settings) != initial_url:
+                    self.crash()
                 with closing(Session()) as session:
                     self.logger.info("Expiring edges....")
                     self.expire_edges(session)
@@ -139,18 +145,12 @@ class BackgroundProcessor(object):
 
                 stats.log_gauge("successful-background-update", 1)
                 stats.log_gauge("failed-background-update", 0)
-            except OperationalError:
-                Session.configure(bind=get_db_engine(get_database_url(self.settings)))
-                self.logger.critical("Failed to connect to database.")
-                stats.log_gauge("successful-background-update", 0)
-                stats.log_gauge("failed-background-update", 1)
-                self._capture_exception()
             except:
                 stats.log_gauge("successful-background-update", 0)
                 stats.log_gauge("failed-background-update", 1)
                 self._capture_exception()
                 self.logger.exception("Unexpected exception occurred in background thread.")
-                raise
+                self.crash()
 
             self.logger.debug("Sleeping for {} seconds...".format(self.settings.sleep_interval))
             sleep(self.settings.sleep_interval)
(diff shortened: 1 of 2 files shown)


config/dev.yaml
```diff
@@ -12,6 +12,10 @@ common:
     # Type: str
     database_source: ""
 
+    # If set to false, anytime the url output by database_source changes the program will crash
```
rra: Could you mention here that `database_source` is run every `sleep_interval` seconds? That wasn't immediately obvious to me. Also might be worth documenting that this option does nothing if `database_source` is not set.

Nit: "any time".

grouper/background/background_processor.py
```diff
@@ -49,6 +50,9 @@ def _capture_exception(self):
         if self.sentry_client:
             self.sentry_client.captureException()
 
+    def crash(self):
+        os._exit(1)
```
rra: This seems aggressive. It will die in the middle of open transactions and not run any exception `finally` blocks or rollbacks. Does it not work to call `sys.exit(1)` like one normally would and let that exception propagate?

grouper/database.py
```diff
@@ -24,10 +25,19 @@ def capture_exception(self):
         if self.sentry_client:
             self.sentry_client.captureException()
 
+    def crash(self):
+        os._exit(1)
```
rra: Same comment here.

config/dev.yaml
herb: i'm kind of dubious having this option is meaningful. using the option means grouper is RO at best and unusable at worse. and it's cognitive load at configuration/setup time that isn't buying much.i vote to not have this flag.

grouper/background/background_processor.py
```diff
@@ -120,8 +124,14 @@ def expire_nonauditors(self, session):
...
     def run(self):
         # type: () -> None
+        initial_url = get_database_url(self.settings)
         while True:
             try:
+                if (
+                    not self.settings.allow_db_change and
+                    get_database_url(self.settings) != initial_url
+                ):
+                    self.crash()
```
herb: nit: formatting:
```
if (not self.settings.allow_db_change 
        and get_database_url(self.settings) != initial_url):
```

grouper/database.py
```diff
@@ -24,10 +25,19 @@ def capture_exception(self):
...
+
     def run(self):
+        initial_url = get_database_url(self.settings)
         while True:
             self.logger.debug("Updating Graph from Database.")
             try:
+                if (
+                    not self.settings.allow_db_change and
+                    get_database_url(self.settings) != initial_url
+                ):
```
herb: same formatting nit as above

config/dev.yaml
herb: if we get rid of this flag we should also get rid of the code that [tries to reconnect to the db in background_processor](https://github.com/dropbox/merou/pull/414/files#diff-9e79234266f5443175806b67277a40cdR153).

grouper/background/background_processor.py
karthik-shanmugam: In python, `sys.exit()` will raise will raise a `SystemExit` exception, which will cause this thread to exit. However, the process will not terminate unless `sys.exit()` is called from the main thread. The only other implementation I can think of is something causing the main thread to exit, which will be tricky because the main thread will usually be in an endless `app.run()` call or something.

https://stackoverflow.com/questions/905189/why-does-sys-exit-not-exit-when-called-inside-a-thread-in-python

rra: Oh, ugh. Yeah, in that case `_exit` is probably the best we can do.

config/dev.yaml
karthik-shanmugam: Is that necessarily true? What if some error other than a URL change occurs and we want to reconnect to the db?

herb: > Is that necessarily true?

is that in reference to removing the reconnection code? my suspicion is recovery from sqlachemy `OperationalError` is fraught and we should restart the process anyways to be sure we recovery correctly.

karthik-shanmugam: Alright I decided to handle ALL exceptions in both threads by crashing (the current behavior was to raise it, which I'm pretty sure kills the background thread while leaving the rest of the process running)

karthik-shanmugam:
I believe grouper currently has no tests for either `background_processor.py` or for the database thread

grouper/settings.py
```diff
@@ -91,6 +91,7 @@ def default_settings_path():
     return os.environ.get("GROUPER_SETTINGS", "/etc/grouper.yaml")
 
 settings = Settings({
+    "allow_db_change": True,
```
rra: Don't forget to remove this too.

rra:
> I believe grouper currently has no tests for either `background_processor.py` or for the database thread

After thinking about this further, I think that if you're already mostly done with adding a test, it would be great to have one, but if you haven't started yet, you should just skip that and land this. I'm going to have to redeploy Grouper soon anyway to fix removing users from role user groups, and it would be great to have this in. This is also a nice bit of KTLO reduction and progress towards our OKRs.

karthik-shanmugam:
I'm TRYING to write a test but my best idea for how to test it is pretty ugly. I could write an itest that runs grouper as a subprocess pointing at a temp file as its config file, edit the db url in the config file, and poll the process to make sure it exits within a timeout. I think just landing this would be a better idea.

rra:
Yeah, just land it. I think forward progress here is more important than closing our testing holes, and this is a tricky thing to test.

PR #425, opened 2019-01-03
Add a settings parameter for auditors group

cauthu:
The auditors group will have the auditing permission, and this group will soon be used in a future change.

### DIFF
--- a/grouper/ctl/sync_db.py
+++ b/grouper/ctl/sync_db.py
@@ -3,6 +3,7 @@ from sqlalchemy.exc import IntegrityError
 from grouper.constants import (
         GROUP_ADMIN,
         PERMISSION_ADMIN,
+        PERMISSION_AUDITOR,
         SYSTEM_PERMISSIONS,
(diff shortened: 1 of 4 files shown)


grouper/ctl/sync_db.py
```diff
@@ -63,6 +67,27 @@ def sync_db_command(args):
...
+
+        try:
+            auditors_group.add(session)
+            session.flush()
+        except IntegrityError:
+            session.rollback()
+            raise Exception('Failed to create group: {}'.format(auditors_group_name))
+
+        permission = Permission.get(session, PERMISSION_AUDITOR)
+        assert permission, "Permission should have been created earlier!"
```
wrodriguez-dbx: Nit: move this check before the group creation just in case it fails I guess

cauthu: nah it's fine. following the pattern above with the admin group. also this is to init a db for dev, so whatevs :P

PR #427, opened 2019-01-04
background processor automatically promotes non-auditor approvers to auditors

cauthu:
instead of scheduling the non-auditor approvers to expire out of the audited groups, we are going to convert them into auditors (and send them a notification email about it).


tests/test_audit.py
```diff
@@ -79,38 +81,39 @@ def test_assert_controllers_are_auditors(groups):  # noqa
...
-    resp = yield http_client.fetch(disable_url, method="POST", headers=priv_headers, body="")
-    assert resp.code == 200
-    # Perm no longer audited
-    resp = yield http_client.fetch(disable_url, method="POST", headers=priv_headers, body="")
-    assert resp.code == 200
-    # Perm still not audited
-    resp = yield http_client.fetch(enable_url, method="POST", headers=priv_headers, body="")
-    assert resp.code == 200
-    # Perm audited again
+    with patch('grouper.permissions.get_auditors_group_name', return_value='auditors'):
```
cauthu: this is just adding the `with patch` and indenting the rest, i.e., the body of the test was unchanged.

tests/test_permissions.py
```diff
@@ -241,6 +250,125 @@ def test_permission_grant_to_owners(session, standard_graph, groups, grantable_p
...
+def test_auditor_promition_when_enabling_permission_auditing(
+        session, graph, permissions, users):
+    """Test automatic promotion of non-auditor approvers
+
+    Two cases for this promotion: 1) when a permission becomes
+    audited, it can cause groups to be come audited, and we promote
+    non-auditor approvers of those groups; 2) when a non-auditor
+    approver is added to an audited group.
+
+    We use the standard_graph fixture here only for the `auditors`
```
cauthu: will remove this stale comment about standard_graph fixture.

tests/test_audit.py
cauthu: stale

cauthu:
previously the plan was to perform the automatic promotion at the following cases:
1. when enabling auditing on a permission
2. when a non-auditor user is added as an approver to an audited group
3. when a non-auditor user is edited to become an approver of an audited group
4. when an audited permission is granted to a group
5. maybe more?

it was starting to get hairy, so we decided to drop that approach, and go with the background processing doing the check-and-promote periodically.

grouper/audit.py
```diff
@@ -130,3 +133,15 @@ def get_audits(session, only_open):
...
+
+
+def get_auditors_group(session):
+    group_name = get_auditors_group_name(settings)
+    if not group_name:
+        raise NoSuchGroup('Please ask your admin to configure the default auditors group name')
+    auditors_group = Group.get(session, name=group_name)
+    if not auditors_group:
+        raise NoSuchGroup('Please ask your admin to configure the default group for auditors')
+    # maybe we can insist that the auditors group actually has the
```
cauthu: we probably *should* crash here or the background processor if the auditors group doesn't actually have to audit permission. this is so that if the group is misconfigured, the background processor crashes so we know to take a look; otherwise it just keeps adding non-auditor approvers to the group but that effectively means nothing because the next time around, the users will again have no audit permission.

essentially this would be failing silently. or, instead of crashing, we send an email to grouper admin. but i think crashing is the easy solution and should only happen once or very rarely, and so the disruption should be minimal. and also emailing doesn't guarantee that emails will be read :)

wrodriguez-dbx: Yea, logic makes sense to me. Maybe on principle add a quick comment on the function saying that it'll crash? Or to be really aggressive add a flag to control behavior but don't think that's necessary at this point

grouper/background/background_processor.py
```diff
@@ -132,8 +138,8 @@ def run(self):
                     self.logger.info("Expiring edges....")
                     self.expire_edges(session)
 
-                    self.logger.info("Expiring nonauditor approvers in audited groups...")
-                    self.expire_nonauditors(session)
+                    self.logger.info("Promote nonauditor approvers in audited groups...")
```
wrodriguez-dbx: Nit: Promote -> Promoting

grouper/background/background_processor.py
```diff
@@ -81,26 +84,27 @@ def expire_edges(self, session):
...
-        settings.nonauditor_expiration_days days in the future.
+        the PERMISSION_AUDITOR permission. All non-auditor approvers of audited groups will be
+        promoted to be auditors, i.e., added to the auditors group.
 
         Args:
             session (Session): database session
         """
-        now = datetime.utcnow()
         graph = Graph()
-        exp_days = timedelta(days=self.settings.nonauditor_expiration_days)
```
wrodriguez-dbx: Safe to remove `nonauditor_expiration_days` from settings file loading code?

grouper/fe/templates/email/nonauditor_promoted_html.tmpl
```diff
@@ -0,0 +1,11 @@
+{% extends "email/base_html.tmpl" %}
+
+{% block subject %}Added as member to '{{ auditors_group_name }}'{% endblock %}
+
+{% block content %}
+<p>You are an approver (owner or manager) of one or more audited groups, but you were previously not an auditor.
+We have automatically made you an auditor by adding you to group '{{ auditors_group_name }}'.
+For more information, please see <a href="{{ url }}/help">Grouper Help</a>.
```
wrodriguez-dbx: Can we point them instead to the description on the auditor group page? The Grouper Help page doesn't seem like it'd be helpful here.
(Or maybe that means we should beef up the Grouper Help page...)

tests/test_audit.py
```diff
@@ -200,3 +210,181 @@ def test_audit_end_to_end(session, users, groups, http_client, base_url, graph):
...
+    assert get_users(graph, AUDITORS_GROUP) == set(["user14@a.co"])
+    assert get_users(graph, "group-3") == set(["user12@a.co", "user22@a.co"])
+
+    #
+    # run the promotion logic -> nothing should happen because the
+    # test-permission is not yet audited
+    #
+    background = BackgroundProcessor(settings, None)
+    background.promote_nonauditors(session)
+    session.commit()
```
wrodriguez-dbx: Should be redundant (called at the end of promote_nonauditors) I believe? Also for the other call

cauthu: done

grouper/background/background_processor.py
cauthu: done

grouper/audit.py
cauthu: have made this function do the check and raise exception if the permission is not in the group

grouper/fe/templates/email/nonauditor_promoted_html.tmpl
cauthu: eh later

grouper/background/background_processor.py
cauthu: later

PR #428, opened 2019-01-12
call `graph.get_groups(audited=True)` instead of `get_audited_groups()`

cauthu:
`get_audited_groups()` actually calls `get_all_groups()` :(
we already that data cached in the graph, so let's try to use it.

### DIFF
--- a/grouper/background/background_processor.py
+++ b/grouper/background/background_processor.py
@@ -100,7 +100,8 @@ class BackgroundProcessor(object):
         # user is a nonauditor approver
         nonauditor_approver_to_groups = defaultdict(set)  # type: Dict[User, Set[str]]
         # TODO(tyleromeara): replace with graph call
-        for group in get_audited_groups(session):
+        for group_tuple in graph.get_groups(audited=True, directly_audited=False):
+            group = Group.get(session, pk=group_tuple.id)
             members = group.my_members()
             # Go through every member of the group and add them to
             # auditors group if they are an approver but not an
--- a/grouper/background/background_processor.py
+++ b/grouper/background/background_processor.py
@@ -99,29 +99,32 @@ class BackgroundProcessor(object):
         # map from user object to names of audited groups in which
         # user is a nonauditor approver
         nonauditor_approver_to_groups = defaultdict(set)  # type: Dict[User, Set[str]]
-        # TODO(tyleromeara): replace with graph call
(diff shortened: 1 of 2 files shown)


grouper/background/background_processor.py
```diff
@@ -99,28 +99,32 @@ def promote_nonauditors(self, session):
...
-            for (type_, member), edge in members.iteritems():
-                # Auditing is already inherited, so we don't need to handle that here
-                if type_ == "Group":
+        user_is_auditor = {}  # type: Dict[str, bool]
+        for group_tuple in graph.get_groups(audited=True, directly_audited=False):
+            group_md = graph.get_group_details(group_tuple.groupname, expose_aliases=False)
+            # members = group.my_members()
+            # # Go through every member of the group and add them to
+            # # auditors group if they are an approver but not an
+            # # auditor
```
wrodriguez-dbx: Nit: comment format & commented out code

grouper/background/background_processor.py
```diff
@@ -99,28 +99,32 @@ def promote_nonauditors(self, session):
...
+                if username not in user_is_auditor:
+                    user_perms = graph.get_user_details(username)['permissions']
+                    user_is_auditor[username] = any(
+                        [p['permission'] == PERMISSION_AUDITOR for p in user_perms])
+                if user_is_auditor[username]:
+                    # user is already auditor so can skip
                     continue
-                member = User.get(session, name=member)
-                member_is_approver = user_role_index(member, members) in APPROVER_ROLE_INDICES
-                member_is_auditor = user_has_permission(session, member, PERMISSION_AUDITOR)
```
wrodriguez-dbx: Can we not still use this function? Would simplify code, and you end up calling User.get later on anyway so doesn't hurt to pull that call up here.

grouper/graph.py
```diff
@@ -467,7 +467,16 @@ def get_disabled_groups(self):
             return sorted(self.disabled_group_tuples.values(), key=lambda g: g.groupname)
 
     def get_groups(self, audited=False, directly_audited=False):
-        """ Get the list of groups as GroupTuple instances sorted by groupname. """
+        """Get the list of groups as GroupTuple instances sorted by groupname.
+
+        Arg(s):
+            audited (bool): true to get only audited groups
+            directly_audited (bool): true to get only directly audited
```
wrodriguez-dbx: Yay comments so I don't have to read this function too :D

PR #433, opened 2019-02-07
Refactor permission disabling, add grouper-ctl command

rra:
Add a permission disable command to grouper-ctl with the same
functionality as the disable permission button in the UI.

Refactor the code to introduce UI objects, service objects, and
a use case object, with interfaces between them implemented with
abstract base classes.  Plumb this particular use case from the
UI down to nearly the storage layer, except for permission
checking and a storage service with graph integration (which will
be addressed in an upcoming change).

Add a group with permissions admin to the standard graph fixture
and update the permission tests accordingly.

### DIFF
--- a/grouper/ctl/permission.py
+++ b/grouper/ctl/permission.py
@@ -37,28 +37,23 @@ class PermissionCommand(CtlCommand, DisablePermissionUI):
     def disabled_permission(self, name):
         # type: (str) -> None
         logging.info("disabled permission %s", name)
-        self.exit_status = 0
 
     def disable_permission_failed_because_not_found(self, name):
         # type: (str) -> None
         logging.critical("permission %s not found", name)
-        self.exit_status = 1
 
     def disable_permission_failed_because_permission_denied(self, name):
         # type: (str) -> None
         logging.critical("not permitted to disable permission %s", name)
-        self.exit_status = 1
 
     def disable_permission_failed_because_system_permission(self, name):
         # type: (str) -> None
         logging.critical("cannot disable system permission %s", name)
-        self.exit_status = 1
 
     def run(self, args):
-        # type: (Namespace) -> int
+        # type: (Namespace) -> None
         """Run a permission command."""
         audit_log = AuditLogService(self.session)
         service = PermissionService(self.session, audit_log)
         usecase = DisablePermission(self.session, args.actor_name, self, service)
         usecase.disable_permission(args.name)
-        return self.exit_status
--- a/grouper/ctl/permission.py
+++ b/grouper/ctl/permission.py
@@ -1,0 +1,64 @@
+import logging
+
+from typing import TYPE_CHECKING
+
+from grouper.ctl.base import CtlCommand
+from grouper.services.audit_log import AuditLogService
+from grouper.services.permission import PermissionService
+from grouper.usecases.disable_permission import DisablePermission, DisablePermissionUI
+
+if TYPE_CHECKING:
+    from argparse import _SubParsersAction, Namespace  # noqa: F401
+    from grouper.models.base.session import Session  # noqa: F401
+
+
+class PermissionCommand(CtlCommand, DisablePermissionUI):
+    """Commands to modify permissions."""
+
+    def add_parser(self, subparsers):
+        # type: (_SubParsersAction) -> None
+        """Add the command-line options."""
+        parser = subparsers.add_parser("permission", help="Manipulate permissions")
+        parser.set_defaults(func=self.run)
+
+        parser.add_argument(
+            "-a",
+            "--actor",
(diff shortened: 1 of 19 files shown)


grouper/services/audit_log.py
```diff
@@ -0,0 +1,64 @@
...
+    ):
+        # type: (...) -> None
+        """Internal method to log an action to the audit log.
+
+        All uses of AuditLog.log should be replaced with more specific entry points like the above,
+        which in turn dispatch to this private method to make the database change.  Arguments don't
+        cover all use cases yet.  This method will be expanded as further use cases are ported to
+        this service.
+        """
+        actor = User.get(self.session, name=authorization.actor)
```
cbguder: Similarly here we can get the user from UserRepository when we have that

grouper/usecases/interfaces.py
```diff
@@ -0,0 +1,47 @@
...
+class PermissionNotFoundException(Exception):
+    """Attempt to operate on a permission not found in the storage layer."""
+
+    def __init__(self, name):
+        # type: (str) -> None
+        msg = "Permission {} not found".format(name)
+        super(PermissionNotFoundException, self).__init__(msg)
+
+
+class PermissionInterface(object):
```
cbguder: It looks like we use this in one place. Do we need an interface here? If we do, I'd argue for calling it `PermissionService` and naming the concrete implementations after what differentiates them (e.g. `DatabasePermissionService`)

grouper/usecases/disable_permission.py
```diff
@@ -0,0 +1,60 @@
...
+    """Disable a permission."""
+
+    def __init__(self, session, actor, ui, service):
+        # type: (Session, str, DisablePermissionUI, PermissionInterface) -> None
+        self.session = session
+        self.actor = actor
+        self.ui = ui
+        self.service = service
+
+    def disable_permission(self, name):
```
cbguder: It's almost like you can see the actual business logic at a glance. Magical =)

grouper/ctl/permission.py
```diff
@@ -0,0 +1,64 @@
...
+        self.exit_status = 1
+
+    def run(self, args):
+        # type: (Namespace) -> int
+        """Run a permission command."""
+        audit_log = AuditLogService(self.session)
+        service = PermissionService(self.session, audit_log)
+        usecase = DisablePermission(self.session, args.actor_name, self, service)
+        usecase.disable_permission(args.name)
+        return self.exit_status
```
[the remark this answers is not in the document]
rra: It's the return value for the command. Without this code, `grouper-ctl` exits with status 0, even though it failed. Definitely open to a better way to do this. I just couldn't come up with a better one.

grouper/services/audit_log.py
rra: Right, this refactoring stops short of the storage layer.

grouper/usecases/interfaces.py
rra: Without the interface, we have a type-checking dependency on the service in the use case, which potentially creates a circular dependency. This is why the hexagonal design pattern always defines an interface in the domain logic for every port between the domain and the application.

Since there's only one instantiation, calling the interface `PermissionService` leads to awkward naming for the actual service (`DefaultPermissionService` or something). I had it that way at one point and thought it was too ugly. I like the clarity of `Interface`, honestly, because it makes clear that this is an abstract base class (which otherwise isn't typical in Python).

grouper/ctl/permission.py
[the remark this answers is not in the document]
rra: An alternative would be to call `sys.exit` directly from the failure handler. I'm generally not a big fan of calling `sys.exit` anywhere down inside objects, but maybe that's better here?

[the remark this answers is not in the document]
rra: Ah, I see this is an existing bug everywhere in `grouper-ctl` and currently the return value isn't used, although at some point we have to fix that bug. I can omit this for now and make a note to come back to it later.

The problem with calling `sys.exit` directly is that if one later wraps that object in one that disables a bunch of permissions from an input file, now you have to do horrible things like catch `SystemExitException` if you want to process as many lines as possible, and bleh.

grouper/usecases/interfaces.py
cbguder: ah, I see. it feels like this can also be resolved by having the use case interface in a separate package that has no dependencies, but I might be missing something.

grouper/ctl/permission.py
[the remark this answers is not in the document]
cbguder: the UI in this case is the Command object right? we should probably avoid reusing those, so I don't think we'd run into this particular scenario

PR #445, opened 2019-02-13
Refactor list permissions

rra:
Add a new ListPermissions use case, backed by the graph, and replace
the frontend handler code with code that calls it.
    
This also refactors the PermissionRepository to have two
implementations, one based on the graph and one based on SQL.  The
grouper-ctl permission command now instantiates the graph so it is
(hopefully temporarily) rather slow.
    
Refactor interfaces that ask about the permissions held by a user
into a separate UserService instead of the PermissionService.

### DIFF
--- a/grouper/fe/handlers/permissions_view.py
+++ b/grouper/fe/handlers/permissions_view.py
@@ -1,60 +1,48 @@
-from datetime import timedelta
-
+from grouper.entities.pagination import PaginatedList, Pagination
+from grouper.entities.permission import Permission
 from grouper.fe.util import GrouperHandler
-from grouper.user_permissions import user_creatable_permissions
+from grouper.repositories.factory import RepositoryFactory
+from grouper.services.factory import ServiceFactory
+from grouper.usecases.factory import UseCaseFactory
+from grouper.usecases.list_permissions import ListPermissionsSortKey, ListPermissionsUI
 
 
-def _round_timestamp(timestamp):
-    return timestamp - timedelta(seconds=timestamp.second, microseconds=timestamp.microsecond)
+class PermissionsView(GrouperHandler, ListPermissionsUI):
+    """Controller for viewing the major permissions list.
 
-
-class PermissionsView(GrouperHandler):
+    There is no privacy here; the existence of a permission is public.
     """
-    Controller for viewing the major permissions list. There is no privacy here; the existence of
-    a permission is public.
-    """
 
-    def get(self, audited_only=False):
-        offset = int(self.get_argument("offset", 0))
-        limit = int(self.get_argument("limit", 100))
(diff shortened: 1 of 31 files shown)


rra:
I plan on adding some more tests before merging this, rather than only relying on the existing tests.

grouper/fe/handlers/permissions_view.py
```diff
@@ -1,60 +1,49 @@
...
-            key=sort_keys[sort_key],
-            reverse=(sort_dir == "desc"),
-        )
-
-        total = len(permissions)
-        permissions = permissions[offset : offset + limit]
-
-        can_create = user_creatable_permissions(self.session, self.current_user)
-
+        sort_key = self.get_argument("sort_by", "name")
```
cbguder: I tend to pass these through the use case as well (so they would be passed back to the UI in the callback), but it seems like Tornado holds onto the state here?

tests/usecases/list_permissions_test.py
```diff
@@ -0,0 +1,173 @@
...
+    # type: (Session) -> List[Permission]
+    """Sets up a very basic test graph and returns the permission objects."""
+    early_date = datetime.utcfromtimestamp(1)
+    permissions = [
+        Permission(name="first-permission", description="first", created_on=datetime.utcnow()),
+        Permission(name="audited-permission", description="", created_on=datetime.utcnow()),
+        Permission(name="early-permission", description="is early", created_on=early_date),
+    ]
+    for permission in permissions:
+        sql_permission = SQLPermission(
```
cbguder: We should be able to replace this with calls to PermissionRepository when we refactor creating permissions, but I think it's already better than fixtures.

grouper/fe/handlers/permissions_view.py
rra: Yeah, I thought about doing that but since Tornado stashes all of the state in its handler class, it wasn't necessary. Although on second thought this is a property of the returned list, so I'm going to change that and pass it through since it feels a bit more correct.

tests/usecases/list_permissions_test.py
rra: Yup, agreed, and making it much nicer is what I'm going to work on today.

grouper/fe/handlers/permissions_view.py
rra: This unfortunately makes some of the types more complicated, which in turn increases the length of the type signature, which in turn runs afoul of a bug in black. Plumbing the `audited_only` parameter through also requires either a much more complex data structure or (better) splitting the use case. I've therefore left this for now and will take a look in a later diff after there has been a new black release (the bug was fixed on tip but there's been no release to PyPI since).

PR #450, opened 2019-02-15
Refactor test setup for usecase tests

rra:
Provide a new TestSetup class that handles creating the database
session and provides utility functions for setting up test data.
This is the replacement for fixtures that set up a complex data
model that most tests don't need and that somewhat obscure what
the test is doing.

This is the first iteration, which messes with the database
directly via SQLAlchemy objects.  A later revision will use the
same service and repository path that non-test code uses.

### DIFF
--- a/itests/fe/permissions_test.py
+++ b/itests/fe/permissions_test.py
@@ -1,4 +1,3 @@
-import logging
 from datetime import datetime
 from time import time
 from typing import TYPE_CHECKING
@@ -55,7 +54,6 @@ def test_list(tmpdir, setup, browser):
 
         # Check the basic permission list.
         page = PermissionsPage(browser)
-        logging.warning("%s", page.root.page_source)
         seen_permissions = [(r.name, r.description, r.created_on) for r in page.permission_rows]
         assert seen_permissions == sorted(expected_permissions)
         assert page.heading == "Permissions"
@@ -97,7 +95,6 @@ def test_list_pagination(tmpdir, setup, browser):
     with frontend_server(tmpdir, "gary@a.co") as frontend_url:
         browser.get(url(frontend_url, "/permissions?limit=1&offset=1"))
         page = PermissionsPage(browser)
-        logging.warning("%s", page.root.page_source)
         seen_permissions = [(r.name, r.description, r.created_on) for r in page.permission_rows]
         assert seen_permissions == sorted(expected_permissions)[1:2]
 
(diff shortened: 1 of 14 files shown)


itests/fe/permissions_test.py
```diff
@@ -1,106 +1,120 @@
...
+        Permission(name="audited-permission", description="", created_on=now),
+        Permission(name="early-permission", description="is early", created_on=early_date),
     ]
-    assert seen_permissions == expected_permissions
+    for permission in permissions:
+        setup.create_permission(
+            name=permission.name,
+            description=permission.description,
+            created_on=permission.created_on,
+            audited=(permission.name == "audited-permission"),
```
cbguder: `audited` should be part of the value object as well, but I guess we haven't written any production code yet to drive that?

itests/fe/permissions_test.py
```diff
@@ -1,106 +1,120 @@
...
+    # type: (LocalPath, SetupTest, Chrome) -> None
+    permissions = create_test_data(setup)
+    expected_permissions = [(p.name, p.description, print_date(p.created_on)) for p in permissions]
+
+    with frontend_server(tmpdir, "gary@a.co") as frontend_url:
+        browser.get(url(frontend_url, "/permissions"))
+
+        # Check the basic permission list.
+        page = PermissionsPage(browser)
+        logging.warning("%s", page.root.page_source)
```
cbguder: Stray logging?

tests/setup.py
```diff
@@ -0,0 +1,149 @@
...
+        self.graph = GroupGraph()
+        self.repository_factory = RepositoryFactory(self.session, self.graph)
+        self.service_factory = ServiceFactory(self.session, self.repository_factory)
+        self.usecase_factory = UseCaseFactory(self.service_factory)
+
+    def create_session(self, tmpdir):
+        # type: (LocalPath) -> Session
+        db_engine = get_db_engine(db_url(tmpdir))
+
+        # If using a persistant database, clear the database first.
```
cbguder: Typo in persistent

itests/fe/permissions_test.py
rra: Right, I left that out because we don't use it yet and hence we don't test it yet. When we refactor `get_permission` to implement the permission view page, that will all happen naturally, and we will need to go back and retrofit this (it will fail mypy typing so it will be forced).

PR #453, opened 2019-02-15
Refactor grouper-ctl initialization

rra:
Move argument setup to a static method so that it can be called
without a session, and change the way that the right class is
determined to look at args.command if args.func is not set.
Automatically initialize all discovered subclasses, so all that's
required to add a new command to ctl/main.py is to import the
relevant class.

Move the factory setup into the CtlCommand base class.

Add sys.exit calls in the failure handlers in PermissionCommand
so that the exit status is correct.

### DIFF
--- a/grouper/ctl/base.py
+++ b/grouper/ctl/base.py
@@ -1,17 +1,8 @@
 from abc import ABCMeta, abstractmethod
 from typing import TYPE_CHECKING
 
-from grouper.graph import Graph
-from grouper.models.base.session import get_db_engine, Session
-from grouper.repositories.factory import RepositoryFactory
-from grouper.services.factory import ServiceFactory
-from grouper.settings import settings
-from grouper.usecases.factory import UseCaseFactory
-from grouper.util import get_database_url
-
 if TYPE_CHECKING:
-    from argparse import _SubParsersAction, Namespace
-    from typing import Optional
+    from argparse import ArgumentParser, Namespace
 
 
 class CtlCommand(object):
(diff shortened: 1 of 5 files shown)


grouper/ctl/base.py
```diff
@@ -1,29 +1,43 @@
...
     @abstractmethod
-    def add_parser(self, subparsers):
-        # type: (_SubParsersAction) -> None
-        """Add the parser for this subcommand."""
+    def add_parser(subparsers):
+        # type: (_SubParsersAction) -> str
+        """Add the parser for this subcommand and return the subcommand name."""
         pass
 
+    def __init__(self, session):
```
cbguder: It feels like we should either always inject the session and the graph, or go one level higher and only inject the use case factory here. That being said, we can wait until we move over other commands to the new style to see if something emerges organically.

grouper/ctl/base.py
```diff
@@ -1,29 +1,43 @@
...
     __metaclass__ = ABCMeta
 
+    @staticmethod
     @abstractmethod
-    def add_parser(self, subparsers):
-        # type: (_SubParsersAction) -> None
-        """Add the parser for this subcommand."""
+    def add_parser(subparsers):
+        # type: (_SubParsersAction) -> str
+        """Add the parser for this subcommand and return the subcommand name."""
```
cbguder: This is kind of a weird contract between CtlCommand subclasses and main, but I don't see a clear way to make it better after a quick glance at the argparse API. It would be nice if the CtlCommand subclass could return a parser that main could just add to the main parser, for instance.

grouper/ctl/base.py
rra: I looked at that, but there are two problems. One is that graph construction is really slow (so we would be slowing down every command a **lot** until we fix that). The other is that some commands explicitly want to **not** instantiate a session or the graph before running the command (`sync_db`, for instance).

I think the right fix for this is a factory. I came up with a design on the way into work and will put up a new diff in a moment for you to look at.

grouper/ctl/base.py
rra: I think it's okay for it to instead have to take a stub parser that it attaches to, but returning the command name is indeed odd. My factory approach would fix that by moving the knowledge of what command name is attached to which command class into the factory.

grouper/ctl/base.py
rra: Take a look at the latest diff and see if that construction looks better to you.

grouper/ctl/factory.py
```diff
@@ -0,0 +1,79 @@
...
+    from typing import Optional
+
+
+class UnknownCommand(Exception):
+    """Attempted to run a command with no known class."""
+
+    pass
+
+
+class CtlCommandFactory(object):
```
cbguder: It feels like this class has two responsibilities: building CtlCommand subclasses, and building top-level dependencies like Session, Graph, and UseCaseFactory. Not a big problem, but we should consider separating these responsibilities. The latter can be shared between API, FE, and CTL too.

rra: I'm going to merge this now so that I can merge some stuff blocked behind it, but will keep thinking about this and propose a fix in a follow-up PR.

PR #459, opened 2019-02-19
Add an audit log repository

rra:
Refactor the AuditLogService to use an AuditLogRepository.  Add a
repository method (not yet plumbed through to a service or use
case) to search for log entries about a specific permission, and
add a test that disabling a permission makes an audit log entry.

### DIFF
--- a/grouper/repositories/audit_log.py
+++ b/grouper/repositories/audit_log.py
@@ -1,0 +1,128 @@
+from datetime import datetime
+from typing import TYPE_CHECKING
+
+from sqlalchemy import desc
+
+from grouper.entities.audit_log_entry import AuditLogEntry
+from grouper.models.audit_log import AuditLog, AuditLogCategory
+from grouper.models.group import Group
+from grouper.models.permission import Permission
+from grouper.models.user import User
+from grouper.plugin import get_plugin_proxy
+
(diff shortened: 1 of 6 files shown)


grouper/repositories/audit_log.py
```diff
@@ -0,0 +1,100 @@
...
+        are ported to this service.
+        """
+        actor = User.get(self.session, name=authorization.actor)
+        if not actor:
+            raise UnknownUserException("unknown actor {}".format(authorization.actor))
+
+        # We currently have no way to log audit log entries for objects that no longer exist.  This
+        # should eventually be fixed via a schema change to use strings for all fields of the audit
+        # log.  For now, we'll die with an exception.
+        user_obj = User.get(self.session, name=on_user)
```
cbguder: This should probably have `if on_user else None` at the end, right? And we probably just want to enforce one of `on_user`, `on_group` or `on_permission` is set?

rra: I'm going to refactor this to use some helper methods, since my attempts to hack around it just made it worse. New diff coming.

There are some audit actions that don't act on any of a user, group, or permission. For instance, starting a global audit has only an actor, action, and description, and none of the other fields.

PR #460, opened 2019-02-20
Move checkpoint updates to the transaction service

rra:
Rather than independently updating the checkpoint counter in every
change, make the transaction service update it on every transaction
commit.  Add a checkpoint service to manage the checkpoint counter.

Use SELECT FOR UPDATE to lock the counter when updating it so that
we won't potentially race with other transactions.

Add a detailed comment about why updating the last_modified column
has been disabled.

### DIFF
--- a/grouper/entities/checkpoint.py
+++ b/grouper/entities/checkpoint.py
@@ -1,0 +1,3 @@
+from typing import NamedTuple
+
+Checkpoint = NamedTuple("Checkpoint", [("checkpoint", int), ("time", int)])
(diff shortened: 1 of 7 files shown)


rra:
BTW, on this, I think there's some further refactoring that might be worthwhile just for consistency: right now, this is the only service that knows about the database session, and it really shouldn't. I think there should be a `TransactionRepository` that handles the actual commit process, and the `TransactionService` should glue together the `CheckpointRepository` and the `TransactionRepository`. It's a minor change, but it will mean the `ServiceFactory` won't need to know about `Session`. I'll probably make that change in a follow-up PR.

PR #463, opened 2019-02-22
Refactor TransactionService to use a repository

rra:
Move the bits that call into SQLAlchemy into a TransactionRepository,
eliminating the last dependency of a ServiceFactory on a Session.
The TransactionService is now only responsible for ensuring a
checkpoint update, so it uses a simpler @contextmanager.  Adjust the
type of the expected interface in the use case to just any context
manager.

Add type annotations to every place a ServiceFactory was being
created so that mypy would catch the argument changes, and update
callers.  This includes removing now-duplicate work from the
PermissionDisable handler and reusing the UseCaseFactory stored in
the handler object.

Add a type signature to the post() method in the PermissionDisable
handler, which requires changing the method signature because mypy
isn't very happy with the flexibility of Tornado's handler methods.

### DIFF
--- a/grouper/fe/handlers/permission_disable.py
+++ b/grouper/fe/handlers/permission_disable.py
@@ -1,8 +1,5 @@
 from grouper.fe.util import GrouperHandler
-from grouper.repositories.factory import RepositoryFactory
-from grouper.services.factory import ServiceFactory
 from grouper.usecases.disable_permission import DisablePermissionUI
-from grouper.usecases.factory import UseCaseFactory
 
 
 class PermissionDisable(GrouperHandler, DisablePermissionUI):
@@ -24,11 +21,10 @@ class PermissionDisable(GrouperHandler, DisablePermissionUI):
         # type: (str) -> None
         return self.forbidden()
 
-    def post(self, user_id=None, name=None):
-        repository_factory = RepositoryFactory(self.session, self.graph)
-        service_factory = ServiceFactory(self.session, repository_factory)
-        usecase_factory = UseCaseFactory(service_factory)
-        usecase = usecase_factory.create_disable_permission_usecase(
+    def post(self, *args, **kwargs):
+        # type: (*str, **str) -> None
+        name = kwargs["name"]
+        usecase = self.usecase_factory.create_disable_permission_usecase(
             self.current_user.username, self
         )
         usecase.disable_permission(name)
--- a/grouper/services/transaction.py
+++ b/grouper/services/transaction.py
@@ -1,54 +1,31 @@
+from contextlib import contextmanager
 from typing import TYPE_CHECKING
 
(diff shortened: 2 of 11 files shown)


grouper/fe/handlers/permission_disable.py
```diff
@@ -24,11 +21,10 @@ def disable_permission_failed_because_system_permission(self, name):
...
         return self.forbidden()
 
-    def post(self, user_id=None, name=None):
-        repository_factory = RepositoryFactory(self.session, self.graph)
-        service_factory = ServiceFactory(self.session, repository_factory)
-        usecase_factory = UseCaseFactory(service_factory)
-        usecase = usecase_factory.create_disable_permission_usecase(
+    def post(self, *args, **kwargs):
+        # type: (*str, **str) -> None
+        name = kwargs["name"]
```
cbguder: This is interesting. It doesn't look like we use `kwargs` for other handlers. Is this one special, or are we moving in this direction?

grouper/services/transaction.py
```diff
@@ -1,54 +1,31 @@
...
-        # type: (Session, CheckpointRepository) -> None
-        self.session = session
+    def __init__(self, transaction_repository, checkpoint_repository):
+        # type: (TransactionRepository, CheckpointRepository) -> None
+        self.transaction_repository = transaction_repository
         self.checkpoint_repository = checkpoint_repository
 
     def commit(self):
         # type: () -> None
         """Provided for tests, do not use in use cases."""
```
cbguder: Would it be too hard to get rid of this?

grouper/fe/handlers/permission_disable.py
rra: Yeah, this is because of mypy. It's not a great situation; it's essentially a workaround.

Tornado allows you to define handlers that take arguments, and those arguments are then filled in based on named regex match groups in the route. However, mypy doesn't allow method signatures to vary in method overrides in subclasses, and from its perspective the `post` and `get` methods are method overrides of `RequestHandler`. The method signatures have to match the typeshed entry for `RequestHandler`, which is `(self, *args, **kwargs)`.

This is the workaround that allows these methods to still be typed. Otherwise, we have to either use `# type: ignore` or leave the handler method untyped, both of which are a bit ugly. (Admittedly, this is also ugly.)

grouper/services/transaction.py
rra: I thought it would be, but taking a closer look, it's easier than I expected. I'll do this in a separate follow-up diff.

grouper/fe/handlers/permission_disable.py
rra: https://github.com/python/typeshed/issues/1214 is a related issue and recommends `# type: ignore`. This felt marginally less bad than `# type: ignore` to me.

PR #464, opened 2019-02-22
 Support ed25519 public keys.

benjaminp:
ed25519 keys get exempt from size checks; all well-formed keys should be sufficiently large.

I modified one fe integration test to use the ed25519 fixture key instead of a rsa one for variety.


rra:
Thank you!

rra:
BTW, I think `dropbox.grouper.plugins.roca` may still reject these keys because `cryptography` doesn't support them, although I'm not sure.

cbguder:
I _think_ it should be safe since that one returns early for non-RSA keys, but there's only one way to be 100% sure.

rra:
Oh, you're right, I somehow missed the first two lines. Yup, we should be good.

PR #466, opened 2019-03-01
Refactor use case factory creation

rra:
Stop requiring a session and graph be injected into RepositoryFactory
and let it create them lazily when needed, since they're both
artifacts of the repository layer.  This requires passing Settings
to the RepositoryFactory (to get the database URL information).

Rather than repeating the same three lines to create a UseCaseFactory,
factor that out into a create_usecase_factory() function in a new
grouper.initialization module, with optional injection of a Session
and GroupGraph (which is somewhat widely used right now, until the API
and FE servers are further refactored).

Refactor Tornado application creation to avoid overriding the
constructor and creating new object attributes, and instead injecting
the global application settings as additional arguments to every
handler's initialize method.  Use this to simplify application creation
and plumb the new usecase_factory through to the handlers.  Stop
plumbing the SentryProxy through, since we aren't using it anywhere.

### DIFF
--- a/grouper/initialization.py
+++ b/grouper/initialization.py
@@ -1,0 +1,25 @@
+"""Setup functions common to all Grouper UIs."""
+
+from typing import TYPE_CHECKING
+
+from grouper.repositories.factory import RepositoryFactory
+from grouper.services.factory import ServiceFactory
+from grouper.usecases.factory import UseCaseFactory
+
+if TYPE_CHECKING:
+    from grouper.graph import GroupGraph
+    from grouper.models.base.session import Session
+    from grouper.settings import Settings
+    from typing import Optional
+
+
+def create_usecase_factory(settings, session=None, graph=None):
+    # type: (Settings, Optional[Session], Optional[GroupGraph]) -> UseCaseFactory
+    """Create a UseCaseFactory, with optional injection of a Session and GroupGraph.
+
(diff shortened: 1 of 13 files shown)


grouper/initialization.py
```diff
@@ -0,0 +1,25 @@
...
+    from grouper.models.base.session import Session
+    from grouper.settings import Settings
+    from typing import Optional
+
+
+def create_usecase_factory(settings, session=None, graph=None):
+    # type: (Settings, Optional[Session], Optional[GroupGraph]) -> UseCaseFactory
+    """Create a UseCaseFactory, with optional injection of a Session and GroupGraph.
+
+    Session and GroupGraph injection are supported primarily for tests.  If not injected, they will
```
cbguder: Looks like we still inject these in production code. When that changes, I'd suggest removing the optional args from this stack (i.e. this method and the constructor). One option worth considering is to pull the session and graph into their own factory/provider and to have two implementations for this, e.g. a DynamicSessionFactory and StaticSessionFactory with the latter being only used in tests. Yay dependency injection! =)

rra: Yup, that sounds like a great idea, and yeah, I only didn't push this farther because right now we're very inconsistent about when we create sessions and graphs and I didn't want to mess too much with the existing code.

PR #469, opened 2019-03-18
Add support for converting a user to a service account

rra:
Add a use case and a grouper-ctl command that converts a user to
a service account by first disabling it and then wrapping it in a
service account and enabling the service account.

Move the user subcommand of grouper-ctl to the new framework so
that it can run use cases, although the existing subcommands are
still implemented the old way.

Move the exceptions for unknown database entities to the
grouper.entities.* modules so that they can be shared across
repositories, services, and use cases (the latter two may need to
catch the exceptions in some situations).

### DIFF
--- a/grouper/ctl/user.py
+++ b/grouper/ctl/user.py
@@ -1,30 +1,26 @@
 import logging
+import sys
+from typing import TYPE_CHECKING
 
 from grouper import public_key
+from grouper.ctl.base import CtlCommand
 from grouper.ctl.util import ensure_valid_username, make_session
 from grouper.models.audit_log import AuditLog
 from grouper.models.user import User
 from grouper.plugin.exceptions import PluginRejectedDisablingUser
 from grouper.role_user import disable_role_user, enable_role_user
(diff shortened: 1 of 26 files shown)


grouper/ctl/user.py
```diff
@@ -127,41 +123,100 @@ def user_command(args):
...
+                "Name of the entity performing this action."
+                " Must be a valid Grouper human or service account."
+            ),
+        )
+
+        subparser = parser.add_subparsers(dest="subcommand")
+
+        user_key_parser = subparser.add_parser("add_public_key", help="Add public key to user")
+        user_key_parser.add_argument("username")
+        user_key_parser.add_argument("public_key")
```
wrodriguez-dbx: Do we ideas for how to keep/make this configuration readable in a world where there are a ton more plugins? At the least we have backlog tasks to add a few more, and this is already a bit painful to visually scan

rra: Yeah, I think we should add another layer that looks like the subcommand layer and configures and runs each sub-subcommand. I can put up another diff to look at that implements that.

PR #472, opened 2019-03-26
Make the permission audit toggling button visible to those with the AUDIT_MANAGER perm

wrodriguez-dbx:
Tested by navigating to the UI and confirming it works.

Note about style: Black wouldn't let me format the assignment of `can_change_audit_status` any other way. Anyone got ideas to make it not ugly? Having an explicit if statement felt even more terrible...

### DIFF
--- a/grouper/fe/handlers/permission_view.py
+++ b/grouper/fe/handlers/permission_view.py
@@ -1,10 +1,11 @@
+from grouper.constants import AUDIT_MANAGER
 from grouper.fe.util import GrouperHandler
 from grouper.permissions import (
     get_groups_by_permission,
     get_log_entries_by_permission,
     get_permission,
 )
-from grouper.user_permissions import user_is_permission_admin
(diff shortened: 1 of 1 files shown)


cbguder:
Oh, that being said, it would be good if we could add a test for this.

wrodriguez-dbx:
> I think that condition (`user_is_permission_admin(...) or user_has_permission(...)`) is a good candidate for extraction into a service method (e.g. `user_can_change_audit_status_for_permission`) that can be called from the relevant use cases to present this view, and to perform the actual toggle. Since we don't have use cases here yet, and since this is intended to be a quick fix, this may not be the best time to undertake that refactor though, so I'll leave that up to you.

Yea, I agree that this will be better after the refactor, but that it doesn't feel worth doing now.

> That being said, it would be good if we could add a test for this.

Eh, there's already test coverage that actually toggling auditing of the perm works. The only thing not tested is whether the button actually shows up, and it didn't feel worth spinning up a new itest setup for the page just to test that one thing.

PR #473, opened 2019-03-27
[WIP] Fix creation of UserToken table at startup

wrodriguez-dbx:
On a clean pull of Grouper, if you run `grouper-ctl sync_db` and then try to view any user on the fe, it errors out with the UserToken table not existing. Best I can tell, it's since the file defining that table isn't imported anywhere prior to the `create_all` call.

The best fix I could think of was to define the User<->UserToken relationship within User, which then must import UserToken. Note that per the semantics of backref ([doc](https://docs.sqlalchemy.org/en/latest/orm/backref.html)), the change should be a logic no-op.

Anyone know if there's a more correct solution I'm missing? (@rra since you were the last to muck with this table relationship)

### DIFF
--- a/grouper/models/user.py
+++ b/grouper/models/user.py
@@ -2,11 +2,13 @@ from typing import TYPE_CHECKING
 
 from sqlalchemy import Boolean, Column, Integer, String
 from sqlalchemy.ext.hybrid import hybrid_property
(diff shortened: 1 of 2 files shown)


rra:
Fixed in #478

PR #475, opened 2019-04-03
README: Use the SVG Travis badge

kevinji:
It's higher-res than the PNG image.


karthik-shanmugam:
Hi Kevin! I did a double take when I saw this was a PR on merou and not on PieCentral 😂

PR #479, opened 2019-04-09
Vary Grouper HTML page titles

brandon-rhodes:
Browser history (as well as any list of open tabs) is fairly useless if
every page ever visited is simply named “Grouper”, so add real titles.


brandon-rhodes:
All right, I went through and invented titles for the rest of the pages! Let's see if the tests all still pass.

PR #484, opened 2019-04-09
Rewrite grouper-ctl user_proxy

rra:
mrproxy doesn't support Python 3.  It is small enough, the changes
for porting are significant enough, and we were already overriding
enough of its logic that it made sense to replace it entirely.

Rewrite using http.server and urllib.request directly, in a way
that is compatible with both Python 2 and Python 3.  Covert it to
the new grouper-ctl class structure.

### DIFF
--- a/grouper/ctl/user_proxy.py
+++ b/grouper/ctl/user_proxy.py
@@ -1,79 +1,157 @@
 from __future__ import print_function
 
-import BaseHTTPServer
 import getpass
 import logging
-import urllib2
+from typing import cast, TYPE_CHECKING
 
-from mrproxy import UserProxyHandler
+from six import PY2
+
+from grouper.ctl.base import CtlCommand
+
+try:
+    from http.server import BaseHTTPRequestHandler, HTTPServer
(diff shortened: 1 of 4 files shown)


grouper/ctl/user_proxy.py
```diff
@@ -1,79 +1,147 @@
...
+from typing import cast, TYPE_CHECKING
 
-from mrproxy import UserProxyHandler
+from grouper.ctl.base import CtlCommand
 
+try:
+    from http.server import BaseHTTPRequestHandler, HTTPServer
+    from urllib.error import HTTPError, URLError
+    from urllib.request import build_opener, HTTPErrorProcessor, Request
+except ImportError:
```
wrodriguez-dbx: Out of curiosity -- is this a hack to make it work with both Python2 and 3?

rra: Yup! Although now that I look at this again in the morning, I'm going to switch to using `six.moves`, which should make this more obvious and not require as much of an awkward workaround.

PR #486, opened 2019-04-10
Add Dockerfile for running tests

brandon-rhodes:
Minimal Dockerfile for getting the tests to execute

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,56 @@
+# This lets you run Grouper or its tests inside of Docker.  Run it with:
+#
+# docker build -t grouper . && docker run -it -p 8888:8888 grouper
+#
+# You will be given a bash prompt and can press Ctrl-P to review several
+# pre-written commands for running and testing the server.  If you want
+# to edit files on your host and see the results immediately in the
+# container, run it with:
+#
+# docker build -t foo . && docker run -it -p 8888:8888 --mount type=bind,source=$PWD,target=/app foo
+
+# Tried building atop the simpler and smaller "python:2.7-slim" base,
+# but its "MariaDB" replacement for MySQL dies with "Specified key was
+# too long; max key length is 767 bytes" when our tests try creating
+# their tables.
+FROM ubuntu:16.04
+
+RUN apt-get update
+
+ENV DEBIAN_FRONTEND noninteractive
+RUN apt-get install -y libmysqlclient-dev mysql-client mysql-server
+RUN apt-get install -y python2.7-dev python-pip gcc
+RUN apt-get install -y chromium-driver
+RUN apt-get install -y procps unzip wget
+
+WORKDIR /app
+COPY ci/setup.sh /app/ci/setup.sh
+COPY requirements*.txt /app/
+
+RUN /etc/init.d/mysql start && mysql -e "\
+ CREATE USER travis@localhost; \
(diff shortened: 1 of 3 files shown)


brandon-rhodes:
Okay, @rra — do you know why the MySQL installed here won't successfully work with the MySQL version of the tests? They all die with things like:
```
E       OperationalError: (_mysql_exceptions.OperationalError) (1071, 'Specified key was too long; max key length is 767 bytes') [SQL: u'\nCREATE TABLE counters (\n\tid INTEGER NOT NULL AUTO_INCREMENT, \n\tname VARCHAR(255) NOT NULL, \n\tcount INTEGER NOT NULL, \n\tlast_modified DATETIME NOT NULL, \n\tPRIMARY KEY (id), \n\tUNIQUE (name)\n)\n\n']
```

rra:
No, no idea, unfortunately. The error message makes no sense to me. It works both in Travis (which is 14.04, I think) and on Debian unstable, so that's a pretty broad range of MySQL versions.

brandon-rhodes:
I switched to Ubuntu 16.04 and, while the Docker image is bigger now, its MySQL defaults handle the database just fine! Let me know if you think of any other tweaks we should make before landing this for the first time.

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+RUN ( \
+ echo 'MEROU_TEST_DATABASE='mysql://travis:@localhost/merou' py.test -x -v'; \
+ echo 'py.test -x -v'; \
+ echo 'tools/run-dev --user user@example.com --listen-host='; \
+ ) > /root/.bash_history
+
+EXPOSE 8888
+
+# By default, launch the CI tests.
```
rra: I think this comment is no longer correct?

requirements-dev.txt
```diff
@@ -3,6 +3,7 @@ flake8-import-order==0.18.1
 flake8==3.7.7
 groupy>=0.5.0
 mock==2.0.0
+more-itertools==5.0.0
```
rra: What does this do?

Dockerfile
brandon-rhodes: I'll fix!

requirements-dev.txt
brandon-rhodes: Fixes Python 2.7, which the most recent version of this package is now incompatible with.

rra: Oh, you're saying that some other dependency listed in requirements-dev.txt pulls in more-itertools without a version cap, and the latest version is Python-2.7-incompatible?

What's pulling in a newer version? I just recreated my virtualenv and still got 5.0.0, and even `pip install -U more-itertools` says that 5.0.0 is the latest version, so it seems like something is correctly figuring this out. (My Python 3 virtualenv pulls in 7.0.0.)

I'm just reluctant to add dependencies that aren't actually dependencies, since it's easy to lose track of why they're there.

brandon-rhodes: Good point — my first guess, if it's working for you elsewhere, is that Docker is using the version of pip back from 16.04 which maybe isn't careful about Python version compatibility? I'll check!

brandon-rhodes: I can't reproduce the problem. It must have been an artifact of the order in which I was doing things in an earlier version of this Dockerfile!

PR #492, opened 2019-04-11
Enable autoescape in Jinja2 templates

rra:
Turn on autoescaping for all HTML templates.  Stop escaping in the
text email templates where this is meaningless anyway.  Explicitly
mark as safe the one place we inject raw HTML (via xsrf_form from
Tornado).

Remove all the old explicit escaping in the Jinja2 templates, since
it would now be confusing (and is just clutter).

Rework the help_for macro so that it doesn't inject its results
into the middle of a tag, which forces escaping and confuses Jinja2.

Rename all email templates to end in .html or .txt for easier
configuration of auto-escaping, except for the subject line
templates that still end in .tmpl.

Add a simple test for searching in the frontend, and use it to test
that escaping is happening.  The test itself has been tested by
marking the search input as safe and confirming the test then fails.
Add class markings to the search results page so that we can more
easily test with Selenium.

Import some additional models in SetupTest so that it can be used
for additional frontend pages.  This is a temporary hack until schema
creation is moved into a proper Repository object.


itests/fe/search_test.py
```diff
@@ -0,0 +1,56 @@
...
+def test_search_escaping(tmpdir, setup, browser):
+    # type: (LocalPath, SetupTest, Chrome) -> None
+    with setup.transaction():
+        setup.create_user("gary@a.co")
+
+    with frontend_server(tmpdir, "gary@a.co") as frontend_url:
+        browser.get(url(frontend_url, "/"))
+
+        page = UserViewPage(browser)
+        page.search_input.send_keys('SEARCH"><marquee>foo<marquee>')
```
wrodriguez-dbx: Should this be `...</marquee>`?

rra: Oh, yes, indeed! It turns out to work either way (I was worried the HTML had to be vaguely valid), but I'll make that fix. Good eye!

PR #493, opened 2019-04-11
Remove public key tags

rra:
The original intent of this feature was to allow a user to have
multiple public keys, and restrict the scope of each public key
to only certain services using permission filters on tags attached
to the public keys.  It was an interesting idea, but proved harder
to deploy than expected and was never used.  It's rare for sites
supporting SSH public keys to have this feature, likely because the
complexity of documenting how to juggle multiple public keys and
set permissions for them is rather high.

Drop this feature to reduce code complexity and to avoid having to
consider it when making future feature changes.


rra:
Oh, yeah. :)

PR #494, opened 2019-04-11
Add some more mypy type annotations

rra:
The more of the source base has type annotations, the safer making
major refactoring changes will be.  Add type annotations to various
places where I was running into trouble with other refactors.  Fix
the various errors detected by type checking.

This is far from comprehensive, just a random collection of places
I'd touched.

### DIFF
--- a/grouper/expiration.py
+++ b/grouper/expiration.py
@@ -1,14 +1,22 @@
 from datetime import timedelta
+from typing import TYPE_CHECKING
 
 from sqlalchemy import or_
 
 from grouper.constants import ILLEGAL_NAME_CHARACTER
 from grouper.email_util import send_async_email
+from grouper.fe.settings import settings as fe_settings
 from grouper.models.async_notification import AsyncNotification
 from grouper.settings import settings
 
+if TYPE_CHECKING:
+    from datetime import datetime
+    from grouper.model.base.session import Session
+    from typing import Iterable, List, Optional, Tuple
+
 
 def _get_unsent_expirations(session, now_ts):
+    # type: (Session, datetime) -> List[Tuple[str, str, str]]
     """Get upcoming group membership expiration notifications as a list of (group_name,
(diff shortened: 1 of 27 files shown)


grouper/expiration.py
```diff
@@ -49,7 +67,6 @@ def add_expiration(session, expiration, group_name, member_name, recipients, mem
         "member_name": member_name,
         "member_is_user": member_is_user,
     }
-    from grouper.fe.settings import settings as fe_settings
```
wrodriguez-dbx: Know of any reason this was inlined in the first place? Seems like it could be intentional

grouper/fe/util.py
```diff
@@ -305,16 +314,18 @@ def ensure_audit_security(perm_arg):
...
+        # type: (Callable[..., None]) -> Callable[..., None]
         def _decorator(self, *args, **kwargs):
+            # type: (GrouperHandler, *Any, **Any) -> None
             if not any(
                 [
                     name == AUDIT_SECURITY and argument == perm_arg
                     for name, argument, _, _ in user_permissions(self.session, self.current_user)
                 ]
             ):
-                return self.forbidden()
```
wrodriguez-dbx: Does `forbidden` stop execution? Haven't dug into it, just noticed that every other callsite seems to follow the `return self.forbidden()` pattern

grouper/expiration.py
rra: I think it's because `grouper.fe.settings` and `grouper.settings` are separate singletons and, depending on the application, only one of them is refreshed in the background. But I have an upcoming diff that rips out the background refresh anyway, so it won't matter.

grouper/fe/util.py
rra: Oh, whoops, yes, this is a bug. Thank you for the catch. This was left over from something else I reverted.

PR #498, opened 2019-04-12
Be consistent about using current_user attribute

rra:
With Tornado, if you define a get_current_user method, it will use
this to populate current_user on demand and cache the result.  We
were using that in most places, but calling get_current_user
directly in some places, which was confusing.  Use the current_user
attribute everywhere, and ensure that everything works properly if
get_current_user returns None.

### DIFF
--- a/grouper/user_permissions.py
+++ b/grouper/user_permissions.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+from typing import TYPE_CHECKING
 
 from six import itervalues
 from sqlalchemy import asc, or_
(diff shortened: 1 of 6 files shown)


grouper/user_permissions.py
```diff
@@ -32,6 +39,8 @@ def user_has_permission(session, user, permission, argument=None):
     Returns:
         bool: Whether or not this user fulfills the permission.
     """
+    if not user:
```
wrodriguez-dbx: I wonder if making `user` optional has some subtle consequences somewhere, say by making a bad request do _something_ instead of erroring out...

Probably not, but could you note the behavior on no user input in the docstring? :)

rra: I can't remember why I added this and it's not necessary. Both mypy and the test suite are happy with making it non-optional. I'll revert that change before landing. Good catch!

PR #499, opened 2019-04-12
Rewrite settings

rra:
Replace the complex Settings object with a much simpler, more
transparent container object that exports its attributes directly,
allowing for better typing.  Remove the background reload thread,
which was both unnecessary complexity and made rollouts dangerous
because the running service would immediately pick up configuration
changes.

Some horrible hacks were required to maintain the global settings
singleton.  This can be slowly teased apart over time with other
refactors, but is currently deeply embedded in the code.  Passing
settings as an argument required very painful and sweeping changes
and would have made the diff far too large.

Fix a long-standing bug where there were multiple global singletons
and only some were reloaded by the background thread.  There is now
a single global settings singleton, but this required changing
settings to be a function (and thus some widespread code changes).

Shuffle some settings around between sections (mostly moving them
to common, although ctl now has a separate section with the oneoffs)
to reflect where those settings are used and to not require frontend
settings when sending email in the background processor.

### DIFF
--- a/config/dev.yaml
+++ b/config/dev.yaml
@@ -11,189 +11,246 @@ common:
     # Type: str
     database: "sqlite:///grouper.sqlite"
 
-    # If this exists, it should be the path to an executable that Grouper will run. This program
-    # should print a single SqlAlchemy URL and exit 0.
+    # If this exists, it should be the path to an executable that Grouper will
+    # run. This program should print a single SqlAlchemy URL and exit 0.
     #
     # Type: str
     database_source: ""
 
+    # The format in which to display dates in the interface. More details here:
+    # https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior
+    #
+    # Type: str
+    date_format: "%Y-%m-%d %I:%M %p"
+
+    # Number of days before membership expires from a group at which
+    # notification of the upcoming expiration will be sent.
+    #
+    # Type: int
+    expiration_notice_days: 7
+
     # Format for logging output.
-    # See https://docs.python.org/2/library/logging.html#logrecord-attributes
+    # See https://docs.python.org/3/library/logging.html#logrecord-attributes
+    #
     # Type: str
     log_format: "%(asctime)-15s\t%(levelname)s\t%(message)s  [%(name)s]"
 
-    # Directories for plugins. If set, load plugins from these directories. Plugins can affect the
-    # behavior of Grouper.
-    # Type: list of str
+    # Number of days for approvers of audited groups who are not auditors to
+    # become auditors before being expired out of the group
+    #
+    # Type: int
+    nonauditor_expiration_days: 5
+
+    # Directories for plugins. If set, load plugins from these directories.
+    #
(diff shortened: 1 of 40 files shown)


config/dev.yaml
```diff
@@ -11,189 +11,246 @@ common:
...
     log_format: "%(asctime)-15s\t%(levelname)s\t%(message)s  [%(name)s]"
 
-    # Directories for plugins. If set, load plugins from these directories. Plugins can affect the
-    # behavior of Grouper.
-    # Type: list of str
+    # Number of days for approvers of audited groups who are not auditors to
+    # become auditors before being expired out of the group
+    #
+    # Type: int
+    nonauditor_expiration_days: 5
```
wrodriguez-dbx: I believe this is no longer used (since the changes to auto-promote non-auditor approvers) so might as well remove. A quick code search turned up nothing as well.

config/dev.yaml
```diff
@@ -11,189 +11,246 @@ common:
...
     debug: true
 
-    # How often to pull cache data from database in seconds.
+    # How to get help from the people who run this Grouper deployment. Should be
+    # in the form of an imperative sentence.
+    # https://en.wikipedia.org/wiki/Sentence_function#Imperative
+    # For example:"email grouper-admin@example.com"
+    #
+    # Type: str
+    how_to_get_help: "if this is prod, ask someone to fix the how_to_get_help setting"
```
wrodriguez-dbx: heh

grouper/api/settings.py
```diff
@@ -1,7 +1,25 @@
...
-    {"address": None, "debug": False, "num_processes": 1, "port": 8990, "refresh_interval": 60},
-)
+if TYPE_CHECKING:
+    from typing import Optional
+
+
+class ApiSettings(Settings):
+    """Grouper API server settings."""
+
+    def __init__(self):
```
wrodriguez-dbx: As is, each of these Settings classes requires three calls to really set up:
```
settings = XSettings()
settings.update_from_config(filename)
update_global_settings(settings)
```

I'm a bit wary about this as a pattern -- it's really easy to forget that the additional calls are necessary, and tests are the only case I can see where you wouldn't always call all three. Would personally rather have it be a single call, something like `set_global_settings("api", filename)`. Not sure what the best solution would be; just my two cents.

(And same for each of the other Settings classes)

config/dev.yaml
rra: Oh, good call. I'll delete this setting.

config/dev.yaml
rra: For the record, this was pre-existing -- I just moved it around. :)

grouper/api/settings.py
rra: Hm. That's definitely a valid point. I think I was writing it that way because I was hopeful it would be possible to drop the third call (and it is already possible in the tests in some places), but that's not really a good argument against adding a class method that does all of those things at the same time and returns the new `Settings` object. I'll do that before merging this.

PR #500, opened 2019-04-12
Lay the groundwork for showing service permission grants

rra:
Currently, viewing a permission only shows the grants to groups. We want to also show the grants to service accounts. This PR lays the groundwork by converting viewing of a single permission to a usecase.

### DIFF
--- a/grouper/entities/permission_grant.py
+++ b/grouper/entities/permission_grant.py
@@ -1,3 +1,10 @@
 from typing import NamedTuple
 
 PermissionGrant = NamedTuple("PermissionGrant", [("name", str), ("argument", str)])
+GroupPermissionGrant = NamedTuple(
+    "GroupPermissionGrant", [("group", str), ("permission", str), ("argument", str)]
+)
+ServiceAccountPermissionGrant = NamedTuple(
+    "ServiceAccountPermissionGrant",
+    [("service_account", str), ("permission", str), ("argument", str)],
+)
--- a/grouper/repositories/permission_grant.py
+++ b/grouper/repositories/permission_grant.py
@@ -3,12 +3,18 @@ from typing import TYPE_CHECKING
 
 from sqlalchemy import or_
 
-from grouper.entities.permission_grant import PermissionGrant
+from grouper.entities.permission_grant import (
+    GroupPermissionGrant,
+    PermissionGrant,
+    ServiceAccountPermissionGrant,
+)
 from grouper.models.base.constants import OBJ_TYPES
 from grouper.models.group import Group
 from grouper.models.group_edge import GROUP_EDGE_ROLES, GroupEdge
 from grouper.models.permission import Permission
 from grouper.models.permission_map import PermissionMap
+from grouper.models.service_account import ServiceAccount
+from grouper.models.service_account_permission_map import ServiceAccountPermissionMap
 from grouper.models.user import User
 from grouper.repositories.interfaces import PermissionGrantRepository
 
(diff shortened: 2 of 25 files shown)


grouper/repositories/permission_grant.py
```diff
@@ -111,6 +143,24 @@ def permission_grants_for_user(self, username):
         )
         return [PermissionGrant(g.name, g.argument) for g in group_permission_grants]
 
+    def service_account_grants_for_permission(self, name):
```
wrodriguez-dbx: Same notes as `group_grants_for_permission`

itests/fe/permissions_test.py
```diff
@@ -124,3 +139,34 @@ def test_create_button(tmpdir, setup, browser):
             setup.add_user_to_group("gary@a.co", "admins")
         browser.get(url(frontend_url, "/permissions?refresh=yes"))
         assert page.has_create_permission_button
+
+
+def test_view(tmpdir, setup, browser):
```
wrodriguez-dbx: Dang I like how clean these tests read

grouper/entities/permission_grant.py
```diff
@@ -1,3 +1,10 @@
 from typing import NamedTuple
 
 PermissionGrant = NamedTuple("PermissionGrant", [("name", str), ("argument", str)])
```
wrodriguez-dbx: Thoughts on removing this entirely? The only place it's currently used is `permission_grants_for_user`, where it could be replaced by `GroupPermissionGrant` with the added bonus of making it clearer where the grant comes from.

grouper/repositories/permission_grant.py
```diff
@@ -51,10 +66,27 @@ def __init__(self, session):
         # type: (Session) -> None
         self.session = session
 
-    def permission_grants_for_user(self, username):
+    def group_grants_for_permission(self, name):
+        # type: (str) -> List[GroupPermissionGrant]
+        permission = Permission.get(self.session, name=name)
+        if not permission or not permission.enabled:
```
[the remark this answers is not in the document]
rra: This was a great point, and it turned into a whole rat hole. Where I ended up at was:

- Service accounts already delete all their permission grants when disabled, so permissions for disabled service accounts aren't a thing, which keeps things simpler
- Groups do keep permission grants when disabled, which allows undoing disabling more easily. I think we eventually want to drop this too, but I'm not ready to do that yet.
- Disabled permissions shouldn't be granted to anything. This is partly because there's no UI to re-enable a disabled permission, so this is more permanent. If the permission is granted to active groups or service accounts, the person disabling it should be forced to remove all of those grants before disabling the permission, since it ensures they've seen all the consequences. But permission grants to disabled groups should just be silently deleted when the permission is disabled, since there's nothing served by keeping them around.

I now have another PR (which depends on this PR and its child since it uses some of the machinery added there) which implements all of that, at which point this code becomes unconditionally correct because there's no way a disabled permission is granted to anything. I'm therefore leaving this as-is, to be cleaned up in that later PR which I'll put up as soon as we get to that point.

grouper/repositories/permission_grant.py
```diff
@@ -51,10 +66,27 @@ def __init__(self, session):
...
+            return []
+        grants = (
+            self.session.query(Group.groupname, PermissionMap.argument)
+            .filter(
+                PermissionMap.permission_id == permission.id,
+                Group.id == PermissionMap.group_id,
+                Group.enabled == True,
```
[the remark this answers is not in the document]
rra: This will be implemented as a flag in that later PR. Another great point.

grouper/repositories/permission_grant.py
rra: As mentioned above, this turns out to not be an issue once we prevent disabling permissions that are granted to a service account, since service accounts remove all their grants when disabled anyway.

itests/fe/permissions_test.py
rra: Yeah, it's so nice, isn't it?

grouper/entities/permission_grant.py
rra: Good point! I'll make this change before merging.

rra: Oh, now I remember why I didn't do this. It turns out to cause serious performance problems for the SQL version of `permission_grants_for_user`, which I was trying hard to make extremely fast. To include the group names in a naive way requires a SQL query per permission grant the user has, which could be 50x or 100x the number of SQL queries that are otherwise required. There may be some more sophisticated subquery approach that would be faster, but it was complex-enough SQL that I wasn't able to work out a good way to do it.

PR #502, opened 2019-04-16
Stop using global settings in grouper/ctl

rra:
Pass settings in properly to all of the grouper-ctl command
implementations and stop using the global settings object.  (Still
set the global settings object, since other code elsewhere in
Grouper called by this code may still require it.)

### DIFF
--- a/grouper/ctl/dump_sql.py
+++ b/grouper/ctl/dump_sql.py
@@ -6,15 +6,15 @@ from sqlalchemy.schema import CreateIndex, CreateTable
 
 from grouper.models.base.model_base import Model
 from grouper.models.base.session import get_db_engine
-from grouper.settings import settings
 from grouper.util import get_database_url
 
 if TYPE_CHECKING:
     from argparse import _SubParsersAction, Namespace
+    from grouper.settings import Settings
 
 
-def dump_sql_command(args):
-    # type: (Namespace) -> None
+def dump_sql_command(args, settings):
+    # type: (Namespace, Settings) -> None
     db_engine = get_db_engine(get_database_url(settings))
     for table in Model.metadata.sorted_tables:
         print(CreateTable(table).compile(db_engine))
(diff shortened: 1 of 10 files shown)


grouper/ctl/util.py
```diff
@@ -35,42 +36,42 @@ def wrapper(args):
...
+        return f(args, settings)
 
     return wrapper
 
 
-def make_session():
-    # type: () -> Session
-    db_engine = get_db_engine(get_database_url(settings()))
+def make_session(settings):
+    # type: (Settings) -> Session
```
wrodriguez-dbx: What's the decision making process for using a generic `Settings` versus a more specific subclass (`CtlSettings` here)? To me it feels like since this is only called within grouper/ctl, this should be `CtlSettings`, but can also see an argument for making it more generic since no ctl-specific settings are used.

grouper/ctl/dump_sql.py
```diff
@@ -6,15 +6,15 @@
...
 
 if TYPE_CHECKING:
     from argparse import _SubParsersAction, Namespace
+    from grouper.settings import Settings
 
 
-def dump_sql_command(args):
-    # type: (Namespace) -> None
+def dump_sql_command(args, settings):
+    # type: (Namespace, Settings) -> None
```
wrodriguez-dbx: Same question as on `make_session` below

grouper/ctl/main.py
```diff
@@ -66,12 +66,12 @@ def main(sys_argv=sys.argv, session=None):
         sa_log.setLevel(logging.INFO)
 
     usecase_factory = create_sql_usecase_factory(settings, session)
-    command_factory = CtlCommandFactory(usecase_factory)
+    command_factory = CtlCommandFactory(settings, usecase_factory)
```
wrodriguez-dbx: Could you put a comment on the call to `global_settings_from_config` above to note that in this case the settings are used both as the global singleton (for code outside grouper/ctl) and as an object that's passed around?

grouper/ctl/util.py
rra: Yeah, good call, this should be `CtlSettings` as well. Will fix that and the other case.

PR #507, opened 2019-04-17
Clean up disabled permission handling

rra:
Since disabling a permission is currently irreversible in the UI,
require that people manually remove all existing permission grants
before disabling the permission.  This also solves some data
consistency issues by ensuring there will not be grants of a
disabled permission, which then has to be handled in other code.

Remove some old tests that now duplicate the usecase test.

To fully establish the invariant that disabled permissions are not
granted to anything, we have to handle the case of disabled groups.
(Service accounts delete all of their permission grants when
disabled, so are not affected.)  For enabled groups, we require the
person disabling the permission revoke all the grants first, but it
doesn't make sense to ask them to do this for disabled groups, which
are invisible in the UI.

Semantically, it makes sense to silently revoke the permission from
all disabled groups, since their grants are retained only to preserve
the group state if disabling it was a mistake, and removing grants of
a separately-disabled permission is not a mistake.

Do that on disable, and add a test that it happens properly.

### DIFF
--- a/grouper/ctl/permission.py
+++ b/grouper/ctl/permission.py
@@ -44,18 +44,23 @@ class PermissionCommand(CtlCommand, DisablePermissionUI):
         # type: (str) -> None
         logging.info("disabled permission %s", name)
 
-    def disable_permission_failed_existing_group_grants(self, name, grants):
-        # type: (str, List[GroupPermissionGrant]) -> None
-        groups = [g.group for g in grants]
-        logging.critical("permission %s still granted to groups %s", name, ", ".join(groups))
-        sys.exit(1)
-
-    def disable_permission_failed_existing_service_account_grants(self, name, grants):
-        # type: (str, List[ServiceAccountPermissionGrant]) -> None
-        service_accounts = [g.service_account for g in grants]
-        logging.critical(
-            "permission %s still granted to service accounts %s", name, ", ".join(service_accounts)
-        )
+    def disable_permission_failed_existing_grants(
+        self,
+        name,  # type: str
+        group_grants,  # type: List[GroupPermissionGrant]
+        service_account_grants,  # type: List[ServiceAccountPermissionGrant]
+    ):
+        # type: (...) -> None
+        message = ""
+        if group_grants:
+            groups = [g.group for g in group_grants]
+            message = "groups " + ", ".join(groups)
+        if service_account_grants:
+            service_accounts = [g.service_account for g in service_account_grants]
+            if message:
+                message += " and "
+            message += "service accounts " + ", ".join(service_accounts)
+        logging.critical("permission %s still granted to %s", message)
         sys.exit(1)
 
     def disable_permission_failed_not_found(self, name):
--- a/grouper/ctl/permission.py
+++ b/grouper/ctl/permission.py
@@ -7,7 +7,12 @@ from grouper.usecases.disable_permission import DisablePermissionUI
 
 if TYPE_CHECKING:
     from argparse import ArgumentParser, Namespace
+    from grouper.entities.permission_grant import (
+        GroupPermissionGrant,
+        ServiceAccountPermissionGrant,
+    )
     from grouper.usecases.factory import UseCaseFactory
+    from typing import List
 
 
 class PermissionCommand(CtlCommand, DisablePermissionUI):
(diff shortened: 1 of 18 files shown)


grouper/services/permission.py
```diff
@@ -29,6 +29,27 @@ def __init__(self, audit_log, permission_repository, permission_grant_repository
...
+        permissions.  Semantically, it makes sense for those grants to quietly disappear without
+        explicit action.
+
+        Therefore, this service call first deletes all grants of the permission to disabled groups
+        before disabling the underlying permission.
+
+        All permission grants for a service account are deleted when that service account is
+        disabled, so this issue doesn't apply for service accounts.
+        """
+        grants = self.permission_grant_repository.revoke_inactive_group_grants_for_permission(name)
```
wrodriguez-dbx: I'm paranoid and would argue for explicitly checking for invariant violations (grants to enabled groups/service accounts) before this.

grouper/permissions.py
```diff
@@ -193,31 +188,6 @@ def grant_permission_to_service_account(session, account, permission, argument="
...
 
-def disable_permission(session, permission_name, actor_user_id):
-    """Set a permission as disabled.
-
-    Args:
-        session(models.base.session.Session): database session
-        permission_name(str): name of permission in question
-        actor_user_id(int): id of user who is disabling the permission
-    """
-    if permission_name in (entry[0] for entry in SYSTEM_PERMISSIONS):
```
wrodriguez-dbx: I vote we keep this check, because I really don't want to be on-call when someone accidentally disables `grouper.permission.auditor` and something crazy happens like the background processor trying to repeatedly "promote" all non-auditor owners :)

grouper/ctl/permission.py
```diff
@@ -39,6 +44,20 @@ def disabled_permission(self, name):
         # type: (str) -> None
         logging.info("disabled permission %s", name)
 
+    def disable_permission_failed_existing_group_grants(self, name, grants):
+        # type: (str, List[GroupPermissionGrant]) -> None
+        groups = [g.group for g in grants]
+        logging.critical("permission %s still granted to groups %s", name, ", ".join(groups))
+        sys.exit(1)
+
+    def disable_permission_failed_existing_service_account_grants(self, name, grants):
```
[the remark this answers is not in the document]
rra: Yeah, that's a good idea -- I'll do that.

grouper/services/permission.py
rra: Hm. I see why you'd want that, and I considered that, but then followed the rule that services will do anything that's possible in the database and invariant enforcement is the responsibility of the usecases. @cbguder, what do you think here? One could make an argument that one should think of this like a schema violation and enforce it in the repository.

grouper/permissions.py
rra: It's still present. https://github.com/dropbox/merou/blob/master/grouper/usecases/disable_permission.py#L60  This is just removing dead code because every disable permission code path now calls the usecase instead.

wrodriguez-dbx: Ahh I was only looking through the immediately added code, thanks

grouper/services/permission.py
wrodriguez-dbx: Yea had been noodling on this question before this PR. The model that makes the most sense to me is:
 - Usecases answer questions of _should_ something happen per the logic of the usecase.
 - Services answer questions of _can_ something happen, which means checking and enforcing **all** invariants.
 - Repositories just blindly do what they're told. If something would be a schema violation, it should be considered an invariant violation and therefore stopped at the service layer.

In particular, in order to stick to the principle of swappable backing repos, invariants _must_ be the same across repos. This leads to defining them at the service layer, making the service layer into a sort of thin waist of the stack.

With all that being said, I'm totally fine with using a different model. No matter which we choose though, I think we should settle on one soon and make it explicit, so we don't end up having a weird hodgepodge of checks like in the old Grouper code.

rra: It's because it would put Grouper in an inconsistent state: if a permission is granted to a disabled group, and then the permission is itself disabled, and then the group is re-enabled, what happens to the permission grant? Currently, we sprinkle `permission.enabled == True` checks all over the place, but the intended semantics of a permission is that re-enabling the permission doesn't restore any grants to groups. (In fact, we want to require that people remove the permission grant to all groups before disabling the permission.)

The root of this problem is that we preserve permission grants in disabled groups as a safety measure so that we can re-enable the group later and not lose all the permission grants, and do so by just special-casing deactivated groups. A better database model would probably move all of those permission grants to a separate table so they don't have to be excluded from every query, and only restore those grants when the group is re-enabled (or we just drop that feature entirely and add a huge warning to disabling the group). But both of those are major changes, way out of scope for this diff.

rra: Talked with @cbguder in person and he had a much better suggestion: change the service layer to unconditionally delete all grants, so it imposes the invariant without needing to do any checks. Then the usecase can do the check and prevent people from deleting a permission with active grants (business logic), and the service just makes it happen, including our consistency rule. I'll implement that.

PR #508, opened 2019-04-18
Rework grouper-ctl sync_db and start on schema service

rra:
Move initialization of the schema into a usecase, and rework the
sync_db grouper-ctl command as a new-style CtlCommand.  Not all of
its functions have moved to usecases yet, but it's a start.

This uncovered a problem with grouper-ctl testing: we were passing
in the config/dev.yaml configuration file, but it uses a different
database URL than tests.  Fix this by adding a -d flag to
grouper-ctl to specify a database URL, consistent with the API
and frontend servers, and have the test suite pass that in to
override the database URL.  This required adding a new parameter
to the legacy run_main test method.

Reorganize the test suite to separate sync_db into its own test and
test more of its functionality.

### DIFF
--- a/grouper/ctl/sync_db.py
+++ b/grouper/ctl/sync_db.py
@@ -9,88 +9,94 @@ from grouper.constants import (
     SYSTEM_PERMISSIONS,
     USER_ADMIN,
 )
+from grouper.ctl.base import CtlCommand
 from grouper.ctl.util import make_session
-from grouper.models.base.model_base import Model
-from grouper.models.base.session import get_db_engine
 from grouper.models.group import Group
 from grouper.permissions import create_permission, get_permission, grant_permission
 from grouper.util import get_auditors_group_name
 
 if TYPE_CHECKING:
-    from argparse import Namespace
+    from argparse import ArgumentParser, Namespace
     from grouper.ctl.settings import CtlSettings
+    from grouper.usecases.factory import UseCaseFactory
 
 
-def sync_db_command(args, settings):
-    # type: (Namespace, CtlSettings) -> None
-    # Models not implicitly or explictly imported above are explicitly imported here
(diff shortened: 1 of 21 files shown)


grouper/ctl/sync_db.py
```diff
@@ -9,88 +9,94 @@
...
+            try:
+                create_permission(session, name, description)
+                session.flush()
+            except IntegrityError:
+                session.rollback()
+                raise Exception("Failed to create permission: %s" % (name,))
+            session.commit()
+
+        # This group is needed to bootstrap a Grouper installation.
+        admin_group = Group.get(session, name="grouper-administrators")
```
wrodriguez-dbx: While you're here, may be worth defining "grouper-administrators" in grouper/constants.py

tests/ctl_util.py
```diff
@@ -1,20 +1,27 @@
...
-def call_main(session, *args):
-    # type: (Session, *str) -> None
-    argv = ["grouper-ctl", "-c", src_path("config", "dev.yaml")] + list(args)
+def call_main(session, tmpdir, *args):
+    # type: (Session, LocalPath, *str) -> None
+    argv = ["grouper-ctl", "-c", src_path("config", "dev.yaml"), "-d", db_url(tmpdir)] + list(args)
     main(sys_argv=argv, session=session)
 
 
 def run_ctl(setup, *args):
```
wrodriguez-dbx: To confirm I'm understanding correctly -- `run_ctl` is the new way of testing that we're slowly migrating to, right? If so, could you throw a comment on `call_main` saying essentially "this isn't what you want"? :)

grouper/ctl/sync_db.py
rra: Good idea. Done.

tests/ctl_util.py
rra: Will do.

PR #510, opened 2019-04-18
Start injecting PluginProxy where needed

rra:
Start the process of migrating away from a global PluginProxy
object to dependency injection.  Move the main entry point for
loading plugins to a static method on PluginProxy, use it to
create a PluginProxy object in each executable, and (for now)
explicitly set it as the global plugin proxy.  Similarly set
an empty one in test initialization.

Move the plugin loading code out of the __init__.py file to
avoid a circular dependency with grouper.plugin.proxy, and also
because code in __init__.py is confusing.  Now only the legacy
global object handling is there.

PluginProxy objects are created by each executable rather than
by the global factory functions because they take the executable
name as a parameter.  Use hyphens instead of underscores for
grouper-api and grouper-fe (a backward-incompatible change) for
consistency (and because Dropbox internal plugins were rewriting
the underscore back to a hyphen anyway).

### DIFF
--- a/grouper/api/main.py
+++ b/grouper/api/main.py
@@ -17,8 +17,9 @@ from grouper.error_reporting import get_sentry_client, setup_signal_handlers
 from grouper.graph import Graph
 from grouper.initialization import create_graph_usecase_factory
 from grouper.models.base.session import get_db_engine, Session
-from grouper.plugin import initialize_plugins
+from grouper.plugin import set_global_plugin_proxy
 from grouper.plugin.exceptions import PluginsDirectoryDoesNotExist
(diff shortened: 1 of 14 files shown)


rra:
The session, the graph, the plugin proxy, and the settings object are the only four that I'm aware of and that I've found to date. We're now injecting or creating all of those, so hopefully there shouldn't be much more of this. I think the settings and the plugin proxy are the only two things that need to vary by application and thus need to be under the control of the application `main`. (Right now we're injecting the session and graph for legacy reasons, but we'll eventually be able to stop.)

PR #511, opened 2019-04-18
Move creating the session into the repository

rra:
Creating the database session requires similar calls to initializing
or dropping the schema, so move that into the SchemaRepository as
well.


rra:
Yeah, it's kind of unsatisfying. I'll keep noodling about it and see if I can come up with anything better, and in the meantime at least update the description.

cbguder:
@rra How about pulling it into its own factory? As for the schema repository, I feel like we don't have to call it a repository if it doesn't look like the other repositories.

rra:
I think the other bits of the schema repository make sense to be a repository (creating and dropping the schema are database-level implementations of actions required by use cases), but moving the session creation into a separate factory makes sense. I'll do that.

PR #512, opened 2019-04-19
Only support datetime in template formatting

rra:
The print_date, expires_when_str, and long_ago_str template utility
functions supported formatting datetime, seconds since epoch, and
a specific string value for a datetime, which added a fair bit of
complexity.  Before refactoring HTML template handling for better
web security, clean this up, since it's not necessary.

We were passing dates other than datetime into these functions for
only two reasons:

* Group.my_members() was doing tricky SQLAlchemy work to merge users
  and groups, which for reasons I don't understand was coercing the
  expiration to a string.  Use much simpler logic to avoid this, at
  the minor cost of one additional round trip to the database.
* Permission information was taken from the graph, which uses seconds
  since epoch for time values.  This is easy to convert to datetime
  before doing template formatting.

Add explicit formatting to email audit templates instead of using the
default formatting for Jinja2 (which doesn't honor Grouper settings),
caught by review of this diff.

Remove some old templates that are no longer referenced.


grouper/fe/template_util.py
```diff
@@ -4,104 +4,78 @@
...
-        date_obj = datetime.strptime(date_obj, "%Y-%m-%d %H:%M:%S.%f")
-        date_obj = date_obj.replace(tzinfo=UTC)
-
-    assert isinstance(date_obj, datetime)
+def highest_period_delta_str(delta):
+    # type: (relativedelta) -> Optional[str]
+    """Given a relativedelta, return a string version of only the longest interval."""
+    for component in DELTA_COMPONENTS:
+        value = getattr(delta, "{}s".format(component))
+        if value > 0:
```
wrodriguez-dbx: I think an explicit `if value < 0: return None` makes this a little easier to understand. Alternatively a note in the docstring that it returns None on negative (or zero) relativedeltas.

grouper/fe/template_util.py
```diff
@@ -4,104 +4,78 @@
...
-from six import string_types
 
 from grouper.fe.settings import settings
 
 if TYPE_CHECKING:
-    from typing import Any, Callable, Dict, Optional, Union
+    from typing import Any, Callable, Dict, Optional
 
+# Components of a relativedelta, in order from longest interval to shortest.
+DELTA_COMPONENTS = ["year", "month", "day", "hour", "minute", "second"]
```
wrodriguez-dbx: Guessing we're not including microseconds for human readability? If so, a comment mentioning that would make the test case more obvious (see below)

tests/fe/util_test.py
```diff
@@ -12,35 +12,47 @@ def test_expires_when_str():
 
     assert expires_when_str(None) == "Never", "no datetime means no expires"
 
-    for date_, expected, msg in [
-        ("2015-08-11 11:00:00.000", "Expired", "long before should expire"),
-        ("2015-08-11 12:00:00.000", "Expired", "same time should expire"),
-        ("2015-08-11 11:59:59.000", "Expired", "minute after should expire"),
-        ("2015-08-11 12:00:00.100", "Expired", "milliseonds before should expire"),
```
wrodriguez-dbx: This case + the equivalent one in `test_long_ago_str` are a bit confusing without a comment somewhere on how we're (manually) only precise to the second, given that `datetime` and `relativedelta` on their own can handle the finer precision

PR #513, opened 2019-04-20
Move the rest of sync_db into a usecase

rra:
The InitializeSchema usecase now does all of the work of sync_db,
including creating the administrator and auditors groups and the
system permissions.  This required creating a GroupService and
plumbing Settings through to ServiceFactory.

### DIFF
--- a/grouper/repositories/group.py
+++ b/grouper/repositories/group.py
@@ -2,14 +2,13 @@ from typing import TYPE_CHECKING
 
 from grouper.entities.group import Group, GroupJoinPolicy
 from grouper.models.group import Group as SQLGroup
-from grouper.repositories.interfaces import GroupRepository
 
 if TYPE_CHECKING:
     from grouper.models.base.session import Session
     from typing import Optional
 
 
-class SQLGroupRepository(GroupRepository):
+class GroupRepository(object):
     """Storage layer for groups."""
 
     def __init__(self, session):
--- a/grouper/repositories/group.py
+++ b/grouper/repositories/group.py
@@ -1,0 +1,33 @@
+from typing import TYPE_CHECKING
+
+from grouper.entities.group import Group, GroupJoinPolicy
+from grouper.models.group import Group as SQLGroup
+from grouper.repositories.interfaces import GroupRepository
+
+if TYPE_CHECKING:
+    from grouper.models.base.session import Session
+    from typing import Optional
+
+
+class SQLGroupRepository(GroupRepository):
+    """Storage layer for groups."""
+
+    def __init__(self, session):
+        # type: (Session) -> None
+        self.session = session
+
+    def create_group(self, name, description, can_join):
+        # type: (str, str, GroupJoinPolicy) -> None
+        group = SQLGroup(groupname=name, description=description, canjoin=can_join.value)
+        group.add(self.session)
+
+    def get_group(self, name):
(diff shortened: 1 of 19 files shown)


grouper/services/group.py
```diff
@@ -0,0 +1,58 @@
...
+        # type: (Settings, GroupRepository, PermissionGrantRepository) -> None
+        self.settings = settings
+        self.group_repository = group_repository
+        self.permission_grant_repository = permission_grant_repository
+
+    def create_group(self, name, description, join_policy):
+        # type: (str, str, GroupJoinPolicy) -> None
+        self.group_repository.create_group(name, description, join_policy)
+
+    def grant_permission_to_group(self, permission, argument, group):
```
wrodriguez-dbx: I think we still need a check somewhere that the argument fits our validation regex, a la https://github.com/dropbox/merou/blob/3bcdf60c8bde6a71303299e853681b1ff433867b/grouper/permissions.py#L156

There's the same question of where it should go as in the thread on the other PR though; in the model I still have in my head it's here but if the usecase is the right place then nevermind :)

grouper/services/group.py
```diff
@@ -0,0 +1,58 @@
...
+        # type: (str, str, str) -> None
+        self.permission_grant_repository.grant_permission_to_group(permission, argument, group)
+
+    def group_exists(self, name):
+        # type: (str) -> bool
+        return True if self.group_repository.get_group(name) else False
+
+    def initialize_administrator_group(self):
```
wrodriguez-dbx: For my own understanding, what's the reasoning behind putting the logic here (and in `initialize_auditors_group`) in the service instead of being in the usecase? As an eye test it looks like a bit of actual logic that consists of calling a bunch of service functions. Guessing there's also only a single usecase that will end up calling this.

grouper/services/permission.py
```diff
@@ -27,6 +27,16 @@ def __init__(self, audit_log, permission_repository, permission_grant_repository
         self.permission_repository = permission_repository
         self.permission_grant_repository = permission_grant_repository
 
+    def create_permission(self, name, description=""):
+        # type: (str, str) -> None
+        self.permission_repository.create_permission(name, description)
+
+    def create_system_permissions(self):
```
wrodriguez-dbx: Same question as on the `initialize_X_group`s above (with probably the same answer 😄)

grouper/services/group.py
rra: Yeah, good call, I'll add this.

grouper/repositories/group.py
```diff
@@ -0,0 +1,33 @@
...
+from grouper.entities.group import Group, GroupJoinPolicy
+from grouper.models.group import Group as SQLGroup
+from grouper.repositories.interfaces import GroupRepository
+
+if TYPE_CHECKING:
+    from grouper.models.base.session import Session
+    from typing import Optional
+
+
+class SQLGroupRepository(GroupRepository):
```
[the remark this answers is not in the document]
rra: I did borrow a bit of work from the future here. I'll unwind this.

grouper/services/group.py
rra: Yeah, in retrospect, my justification for doing this (not wanting to plumb settings into the usecase) was a bad one. I'll change this.

PR #522, opened 2019-04-24
Avoid microseconds in default timestamps

rra:
Behavior of database backends varies when given timestamps that
include microseconds.  MySQL strips them but SQLite stores them.
This can cause confusing test behavior between different backends.
Grouper has no need of sub-second timestamp accuracy, so solve
this problem more systematically by stripping microseconds from
the default timestamps in the database model.

### DIFF
--- a/grouper/models/audit.py
+++ b/grouper/models/audit.py
@@ -1,11 +1,9 @@
-from datetime import datetime
-
 from six import itervalues
 from sqlalchemy import Boolean, Column, DateTime, ForeignKey, Integer
 from sqlalchemy.orm import relationship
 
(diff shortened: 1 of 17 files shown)


wrodriguez-dbx:
Is the only problem we're trying to solve test failures caused by checking for exact timestamp matches? If so, feels like the right solution is either to change the tests to not care about microseconds or to fully ensure microseconds won't show up anywhere. As is, no matter the default timestamp function, the live database can still end up seeing datetimes with microseconds.

rra:
Yeah, on further thought, this was a bad idea and we should just fix the tests.

PR #526, opened 2019-04-25
Mostly fix port binding races in integration tests

rra:
The integration tests found an unused port on which to run the
service under test by binding a socket with system port allocation,
retrieving the bound port number, and then closing that socket.
This has two problems that were causing race conditions:

* We weren't setting SO_REUSEADDR, so depending on how quickly the
  child process started, the kernel may not be willing to let it
  rebind the same port we just closed.
* Even with port reuse, there's an inherent race condition when
  tests are running in parallel or anything else is running on the
  same host: someone else may allocate the same port between the
  point when we close it and the point at which the child server
  rebinds to it.

For the API and Frontend servers, implement a new --listen-stdin
option that tells the service to not bind its own port and instead
use the socket passed as standard input as the network socket.
This allows the test to bind a socket with system port allocation
and pass it open to the server under test, eliminating the race.

This unfortunately is not possible with grouper-ctl user_proxy yet
because it's built on Python's internal HTTPServer, which in turn
does not support starting a server with a user-supplied socket
instead of binding its own socket.  It will need to be rewritten to
use Tornado to fix this.  In the meantime, there will still be a
race, but narrow it by setting SO_REUSEADDR and closing the socket
as close as possible to starting the child process.

Replace the older test fixtures with simple wrappers around the new
setup functions so that they also get the benefit of this change.

Use the current instead of the deprecated way of starting the
Tornado IOLoop (unrelated but noticed it while working on this
change).

### DIFF
--- a/itests/setup.py
+++ b/itests/setup.py
@@ -10,7 +10,7 @@ import socket
 import subprocess
 import sys
 import time
-from contextlib import closing, contextmanager
+from contextlib import contextmanager
 from typing import TYPE_CHECKING
 
 from selenium.webdriver import Chrome, ChromeOptions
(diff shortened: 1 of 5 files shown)


itests/setup.py
```diff
@@ -60,21 +56,24 @@ def _wait_until_accept(port, timeout=3.0):
...
     # type: (LocalPath) -> Iterator[str]
-    api_port = _get_unused_port()
+    api_socket = _bind_socket()
+    api_port = api_socket.getsockname()[1]
 
     cmd = [
         sys.executable,
         src_path("bin", "grouper-api"),
+        "-v",
+        "-v",
```
wrodriguez-dbx: Nit: "-vv"?

rra: Yeah, I'll make that consistent with the other ones that use `-vvc`. (Two `-v` is required to make it log at debug level, IIRC.)

cbguder:
❤️

PR #528, opened 2019-04-26
Use permission repository to create test permissions

rra:
Now that the permission repository supports creating permissions,
use it in SetupTest.  Use the permission repository directly rather
than the service because we want to force specific permission
creation times, which isn't something that seems useful to expose
through the permission service.

### DIFF
--- a/grouper/repositories/interfaces.py
+++ b/grouper/repositories/interfaces.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING
 from six import with_metaclass
 
 if TYPE_CHECKING:
+    from datetime import datetime
     from grouper.entities.pagination import PaginatedList, Pagination
     from grouper.entities.permission import Permission
     from grouper.entities.permission_grant import (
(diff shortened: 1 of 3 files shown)


wrodriguez-dbx:
What are your thoughts on manually setting `created_on` in the test setup `create_permission` instead? Doing so would allow the setup code to call into the service instead of the repo, testing the whole code path. Also seems nice that the repo doesn't need to support specifying the creation date, since then aren't changing the live code to accommodate the test.

rra:
This is a good point, and a good concern, but I want to eventually establish a rule that all code that directly manipulates the database lives in the repository layer so that we only have to touch the repository layer if we change the database schema. I'm worried that having tests mess directly with the database will cause a maintenance burden down the road when we fix some of the bugs in our repository layer. It feels to me like the benefits of having all knowledge of the database layout stop at the repository layer outweighs the messiness of having to add test-only parameters to some of the repository methods.

PR #529, opened 2019-04-26
Fix creation of group with clickthru

rra:
Fix various issues when creating a group with the require clickthru
form field:

* The help popover had a lower z-index than the modal, so displayed
  behind the modal, which wasn't useful to anyone.
* The require clickthru form field was ignored and not persisted to
  the database.
* The fields were too narrow, resulting in odd wrapping.
* A bug in the verification logic meant that we weren't checking
  that the user really did confirm the clickthru agreement (although
  it would have been hard to submit without confirming through a
  browser UI without taking special action).
* The button construction for the join form was odd and overly
  complex, which made testing difficult.

Add an integration test for creating and then requesting membership
in a group with a required clickthru agreement.

### DIFF
--- a/grouper/fe/templates/group-join.html
+++ b/grouper/fe/templates/group-join.html
@@ -32,26 +32,22 @@
                     {% include "forms/group-join.html" %}
                     {% if group.require_clickthru_tojoin %}
                     <input class="clickthru-checkbox" type="checkbox" name="clickthru_agreement"/>
-                    {% else %}
+                    {% endif %}
                     <div class="form-group">
                         <div class="col-sm-offset-3 col-sm-4">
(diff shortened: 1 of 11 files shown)


grouper/fe/templates/group-join.html
```diff
@@ -90,7 +86,6 @@ <h4 class="modal-title">Join Group Agreement</h4>
                     Agree
                 </button>
             </div>
-                </form>
```
wrodriguez-dbx: I'm so confused -- what was this closing?

rra: I know, right? I haven't gone through the commit history to figure out what happened.

PR #531, opened 2019-05-01
Set Content-Security-Policy and Referrer-Policy

rra:
Build a Content-Security-Policy header from external resources and
the configured CDNJS prefix and set it on every request, including
some restrictive default settings.  Also set Referrer-Policy to
same-origin on every request.

Unfortunately, set_default_headers is called before initialize in
request handlers, so the template engine had to move to the Tornado
settings to be accessible early enough.  Since we're making that
change anyway, also move the session class and drop the parameters
to initialize.  This makes us somewhat more vulnerable to changes in
the attributes in newer versions of Tornado, but avoids globals or
messing about with subclassing Application.

### DIFF
--- a/grouper/fe/templating.py
+++ b/grouper/fe/templating.py
@@ -1,9 +1,10 @@
-from typing import NamedTuple, TYPE_CHECKING
+from typing import cast, NamedTuple, TYPE_CHECKING
 
+from grouper.fe.settings import FrontendSettings
 from grouper.templating import BaseTemplateEngine
 
 if TYPE_CHECKING:
-    from grouper.fe.settings import FrontendSettings
+    from typing import Iterable, List
 
 # An external web resource with SRI.
 ExternalResource = NamedTuple("ExternalResource", [("url", str), ("integrity", str)])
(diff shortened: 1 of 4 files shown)


grouper/fe/templating.py
```diff
@@ -78,3 +122,19 @@ def __init__(self, settings, deployment_name):
...
+        policy += "; img-src 'self' " + " ".join(self._cdnjs_urls(EXTERNAL_IMG))
+        policy += "; script-src 'self' " + " ".join(self._cdnjs_urls(EXTERNAL_JS))
+        policy += "; style-src 'self' " + " ".join(self._cdnjs_urls(EXTERNAL_CSS))
+        policy += "; font-src " + " ".join(self._cdnjs_urls(EXTERNAL_FONTS))
+        return policy
+
+    def _cdnjs_urls(self, resources):
+        # type: (Iterable[ExternalResource]) -> List[str]
+        """Return the URLs on the preferred CDNJS mirror for the given resources."""
+        cdnjs_prefix = cast(FrontendSettings, self.settings).cdnjs_prefix
```
wrodriguez-dbx: Feels wrong that the cast is necessary here... don't see a better option though

rra: Yeah, I feel like this is a bug in mypy. Even if I set `self.settings` explicitly in the constructor here, it still uses the type from the parent class and still produces an error without the `cast`. mypy in general doesn't deal with different method signatures on parent and child objects very well.

PR #533, opened 2019-05-01
Add SRI for internal resources, require SRI

rra:
Enable require-sri-for both script and style.  This requires every
loaded resource have SRI, so calculate integrity values for the
internal JavaScript and CSS on initialization of the template engine
and pass them into the template.


itests/fe/permissions_test.py
```diff
@@ -62,7 +62,7 @@ def create_test_data(setup):
 def test_list(tmpdir, setup, browser):
     # type: (LocalPath, SetupTest, Chrome) -> None
     permissions = create_test_data(setup)
-    template_engine = FrontendTemplateEngine(FrontendSettings(), "")
+    template_engine = BaseTemplateEngine(Settings(), "grouper.fe")
```
wrodriguez-dbx: Since this is a test of the frontend's output, I think this should still be a `FrontendTemplateEngine`. In case, for example, we want to use different `print_date` functions between the fe and emails

tests/fe/templating_test.py
```diff
@@ -0,0 +1,29 @@
...
+import grouper.fe
+from grouper.fe.settings import FrontendSettings
+from grouper.fe.templating import FrontendTemplateEngine
+
+
+def test_included_resources():
+    # type: () -> None
+    settings = FrontendSettings()
+    static_path = os.path.join(os.path.dirname(grouper.fe.__file__), "static")
+    engine = FrontendTemplateEngine(settings, "", static_path, package="tests.fe")
```
wrodriguez-dbx: Are we going to have different templates in tests.fe than grouper.fe?

itests/fe/permissions_test.py
rra: The problem is that `FrontendTemplateEngine` has various other prerequisites that I didn't want to initialize, since this test isn't really using the template engine at all, just the date formatting. Instead, I'll just copy the single line of date formatting logic over to the test, since it's highly unlikely to change, and avoid instantiating any template engine at all.

tests/fe/templating_test.py
rra: Yup, that's where the `base.json.tmpl` template used in the test comes from. It's nonsensical for the production code, just a way to generate a template and then be able to easily inspect, from Python, what template variables were set.

PR #535, opened 2019-05-03
Mask database password in logging

rra:
We log the full database URL at debug level when running an external
program, but it can contain a database password.  Remove the password
from the URL before logging it.

### DIFF
--- a/tests/settings_test.py
+++ b/tests/settings_test.py
@@ -1,9 +1,11 @@
+import logging
 import subprocess
 from typing import TYPE_CHECKING
 
 import pytest
 import pytz
 from mock import call, patch
+from six.moves import StringIO
 
 from grouper.settings import DatabaseSourceException, InvalidSettingsError, Settings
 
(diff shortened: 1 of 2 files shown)


tests/settings_test.py
```diff
@@ -125,3 +127,23 @@ def test_database():
...
+    expected_url = "mysql://user:???@example.com:8888/merou"
+
+    log_output = StringIO()
+    log_handler = logging.StreamHandler(log_output)
+    settings._logger.addHandler(log_handler)
+    settings._logger.setLevel(logging.DEBUG)
+
+    with patch("subprocess.check_output") as mock_subprocess:
+        mock_subprocess.return_value = "mysql://user:password@example.com:8888/merou"
+        assert settings.database == test_url
```
wrodriguez-dbx: Nit: a comment calling out that this is where the logging actually happens would be nice :)

rra: Yeah, good idea.

PR #537, opened 2019-05-15
Clarify that the manager role is a superset of the member role.

benjaminp:
It was previously unclear to me that the manager also inherited permissions of the group.


rra:
We should just delete the manager concept entirely as a failed experiment. The distinction between manager and owner has never proven useful. But that requires a database migration, so is a pile of work.

benjaminp:
Yeah, my experience with it is that people with the title "manager" think they should request it when they really want np-owner.

PR #539, opened 2019-05-20
Add statistic for graph updates

rra:
Log the elapsed milliseconds for each graph update.  This will help
us catch regressions and evaluate the impact of further graph
optimizations.

### DIFF
--- a/grouper/graph.py
+++ b/grouper/graph.py
@@ -10,6 +10,7 @@ from sqlalchemy import or_
 from sqlalchemy.orm import aliased
 from sqlalchemy.sql import label, literal
 
+from grouper import stats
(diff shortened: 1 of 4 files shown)


grouper/graph.py
```diff
@@ -172,6 +175,9 @@ def update_from_db(self, session):
                 self.group_tuples = group_tuples
                 self.disabled_group_tuples = disabled_group_tuples
 
+            duration = datetime.utcnow() - start_time
+            stats.log_rate("update_ms", int(duration.total_seconds() * 1000))
```
wrodriguez-dbx: Nit: I'd argue for graph_update_ms

rra: Oh, good call.

PR #540, opened 2019-05-20
Remove unused cutoff API feature

rra:
The API server supported giving an optional cutoff as a parameter
to several routes, which would restrict graph calculations for
group membership and permissions to paths shorter than that cutoff.
This was never supported by Groupy and not tested except as a side
effect of another test, and I see no sign that we ever used it.

Drop this feature to simplify the graph code.

Add type annotations to all changed functions so that mypy has a
better chance of catching errors.  This required cleaning up some
typing in the graph code.

### DIFF
--- a/grouper/graph.py
+++ b/grouper/graph.py
@@ -29,7 +29,7 @@ from grouper.util import singleton
 if TYPE_CHECKING:
     from grouper.models.base.session import Session
     from grouper.service_account import ServiceAccountPermission
-    from typing import Any, Dict, Iterable, List, Optional, Set
+    from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
 
 MEMBER_TYPE_MAP = {"User": "users", "Group": "subgroups"}
 EPOCH = datetime(1970, 1, 1)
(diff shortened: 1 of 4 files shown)


grouper/graph.py
```diff
@@ -718,17 +718,17 @@ def get_user_details(self, username, cutoff=None, expose_aliases=True):
                     "rolename": GROUP_EDGE_ROLES[role],
                 }
 
-                for permission in self.permission_metadata[parent_name]:
+                for service_permission in self.permission_metadata[parent_name]:
```
wrodriguez-dbx: What's the `service` in `service_permission` referring to?

tests/api/handlers_test.py
```diff
@@ -56,8 +56,6 @@ def test_users(users, http_client, base_url):  # noqa: F811
     assert body["status"] == "ok"
     assert sorted(body["data"]["users"]) == all_users
 
-    # TODO: test cutoff
```
wrodriguez-dbx: :D

grouper/graph.py
rra: "service account". The service account permission grants are a different data type than group permission grants, so need to be stored in a separate variable. The variable naming here is not the greatest; I'm hoping to clean that up in a later diff once I've simplified this code some more.

wrodriguez-dbx: Huh, I'm getting lost somewhere in the code -- is `permission_metadata` not explicitly for users and not service accounts? My impression is that the rename should be on https://github.com/rra/merou/blob/c3f022fbf320da505453e2fd985190243789c51e/grouper/graph.py#L675 rather than this loop

rra: Oh, whoops, you're entirely correct, and I changed the wrong loop. Will fix in my next diff.

PR #546, opened 2019-05-22
Switch to PermissionGrant in the graph

rra:
Rather than using a custom MappedPermission named tuple to store
permission grants, use the standard PermissionGrant entity.  This
requires splitting permission grants from permission metadata,
which now contains only information about the permission itself,
and reworking how the graph checks whether permissions are audited.

Rename the name field of PermissionGrant to permission, which is
more consistent with the other versions of this entity and makes
for more obvious code in the graph.

Add a test for granting and revoking service account permissions to
ensure that we're exercising the template code that creates links
based on the permission grant ID.

### DIFF
--- a/grouper/repositories/permission_grant.py
+++ b/grouper/repositories/permission_grant.py
@@ -53,7 +53,10 @@ class GraphPermissionGrantRepository(PermissionGrantRepository):
         permissions = []
         for permission_data in user_details["permissions"]:
             permission = PermissionGrant(
-                name=permission_data["permission"], argument=permission_data["argument"]
+                permission=permission_data["permission"],
+                argument=permission_data["argument"],
+                granted_on=datetime.utcfromtimestamp(permission_data["granted_on"]),
+                is_alias=permission_data["alias"],
             )
             permissions.append(permission)
         return permissions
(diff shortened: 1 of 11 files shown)


grouper/repositories/permission_grant.py
```diff
@@ -178,14 +181,17 @@ def permission_grants_for_user(self, name):
...
+            self.session.query(Permission.name, PermissionMap.argument, PermissionMap.granted_on)
             .filter(
                 Permission.id == PermissionMap.permission_id,
                 PermissionMap.group_id.in_(seen_group_ids),
             )
             .all()
         )
-        return [PermissionGrant(g.name, g.argument) for g in group_permission_grants]
+        return [
+            PermissionGrant(g.name, g.argument, g.granted_on, is_alias=False)
```
wrodriguez-dbx: Tracing through the code, I think that the graph and sql versions of `permission_grants_for_user` differ -- it looks like the graph version drills down to [here](https://github.com/dropbox/merou/blob/c1a349429ef3e53b3b1be641f4ae631d05e6749d/grouper/graph.py#L255) which does properly expand permission aliases, but this sql version never takes aliases into account.

rra: Yes, indeed, good catch. I noticed that too but forgot to add a comment. It's on my to-do list to fix in a later diff (the SQL version of `permission_grants_for_user` isn't used yet).

PR #548, opened 2019-05-22
Fix error from disabling granted permissions

rra:
Also add a test, and rework settings_test.py to use the same pytest
fixture facility instead of messing around with logging handlers.


tests/ctl/permission_test.py
```diff
@@ -7,6 +7,7 @@
 from tests.ctl_util import run_ctl
 
 if TYPE_CHECKING:
+    from pytest.logging import LogCaptureFixture
```
wrodriguez-dbx: Learning about so many random python features from reviewing these diffs :)

PR #549, opened 2019-05-23
Improve error page for invalid users

rra:
Provide a readable error page for syntactically invalid users and
for attempting to visit Grouper as a service account.  This requires
avoiding the magic Tornado get_current_user method override, since
Tornado catches exceptions from that method.  Instead, call our own
method and catch exceptions in prepare.

Add id attributes to page headings and subheadings to make finding
them in Selenium tests more efficient.

Stop pre-creating the user used in frontend Selenium tests so that
we regularly exercise the automatic creation of a new user and thus
don't need a separate explicit test for it.

### DIFF
--- a/grouper/fe/util.py
+++ b/grouper/fe/util.py
@@ -171,23 +171,19 @@ class GrouperHandler(SentryHandler):
     def prepare(self):
         # type: () -> None
         username = self.request.headers.get(settings().user_auth_header)
+
         try:
             user = self.get_or_create_user(username)
         except InvalidUser as e:
-            self.set_status(403)
-            self.raise_and_log_exception(tornado.web.HTTPError(403))
-            self.render(
-                "errors/baduser.html", message=str(e), how_to_get_help=settings().how_to_get_help
-            )
+            self.baduser(str(e))
             self.finish()
             return
 
-        if not user or not user.enabled:
-            self.forbidden()
+        if user and user.enabled:
+            self.current_user = user
+        else:
+            self.baduser("{} is not an active account".format(username))
             self.finish()
-            return
-
-        self.current_user = user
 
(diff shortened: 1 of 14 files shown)


grouper/fe/util.py
```diff
@@ -157,17 +164,31 @@ def get_current_user(self):
...
+        except InvalidUser as e:
+            self.set_status(403)
+            self.raise_and_log_exception(tornado.web.HTTPError(403))
+            self.render(
+                "errors/baduser.html", message=str(e), how_to_get_help=settings().how_to_get_help
+            )
+            self.finish()
+            return
+
+        if not user or not user.enabled:
```
wrodriguez-dbx: Maybe make this also go to the baduser page instead of forbidden

tests/setup.py
```diff
@@ -250,6 +250,12 @@ def grant_permission_to_service_account(self, permission, argument, service_acco
         )
         grant.add(self.session)
 
+    def disable_user(self, user):
+        # type: (str) -> None
+        user_obj = User.get(self.session, name=user)
```
wrodriguez-dbx: Since the user repo/service already exists, feels better to call into the user service `disable_user` function (and similarly for a bunch of the other setup functions that are now able to call into services instead of reimplementing logic)

grouper/fe/util.py
rra: Yeah, thinking about this some more, that's a good idea. Will add that in before landing.

tests/setup.py
rra: Good catch, I missed that we already had a service call for that. (I think I made the other setup functions use a service or repository if one exists. When I have some time, I'll go through and move that code into repositories systematically.)

PR #550, opened 2019-05-23
Remove ServiceAccountPermission named tuple

rra:
Replace it with ServiceAccountPermissionGrant for individual service
accounts and PermissionGrant when retrieving all grants for the graph.

This required adding granted_on and mapping_id to
ServiceAccountPermissionGrant.  Do the same with GroupPermissionGrant,
since this will eventually be required as well.

### DIFF
--- a/grouper/service_account.py
+++ b/grouper/service_account.py
@@ -10,11 +10,12 @@ Account abstraction.  A User that's just the account of a ServiceAccount is flag
 service_account boolean field.
 """
 
-from collections import defaultdict, namedtuple
+from collections import defaultdict
 from typing import TYPE_CHECKING
 
 from sqlalchemy.exc import IntegrityError
 
+from grouper.entities.permission_grant import PermissionGrant, ServiceAccountPermissionGrant
 from grouper.group_service_account import add_service_account
 from grouper.models.audit_log import AuditLog
 from grouper.models.counter import Counter
@@ -30,11 +31,6 @@ if TYPE_CHECKING:
     from grouper.models.base.session import Session
     from grouper.models.group import Group
     from typing import Dict, List, Union
-
-# A single service account permission.
-ServiceAccountPermission = namedtuple(
-    "ServiceAccountPermission", ["permission", "argument", "granted_on", "mapping_id"]
-)
 
 
 class BadMachineSet(Exception):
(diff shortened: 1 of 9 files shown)


grouper/service_account.py
```diff
@@ -199,51 +196,60 @@ def enable_service_account(session, actor, service_account, owner):
...
-    for permission in permissions:
+    for grant in grants:
         out.append(
-            ServiceAccountPermission(
-                permission=permission[0].name,
-                argument=permission[1].argument,
-                granted_on=permission[1].granted_on,
-                mapping_id=permission[1].id,
+            ServiceAccountPermissionGrant(
+                service_account=service_account,
```
wrodriguez-dbx: Is there a way this is automatically getting the username from `service_account`, or should it be `service_account.user.username`? If the latter I don't see how mypy isn't yelling unless it calls [`__repr__`](https://github.com/dropbox/merou/blob/4f986f60c7ce8c00b5717a890f3890eeb4b5c087/grouper/models/service_account.py#L43)

grouper/service_account.py
```diff
@@ -199,51 +196,60 @@ def enable_service_account(session, actor, service_account, owner):
...
+                granted_on=grant.granted_on,
+                mapping_id=grant.id,
             )
         )
     return out
 
 
 def all_service_account_permissions(session):
-    # type: (Session) -> Dict[str, List[ServiceAccountPermission]]
+    # type: (Session) -> Dict[str, List[PermissionGrant]]
```
wrodriguez-dbx: Hmm, I'm not clear on when `PermissionGrant` is appropriate vs. `ServiceAccountPermissionGrant`. Feels like there ought to be some structuring that avoids awkardness like `service_account_permissions` returning a different type than `all_service_account_permissions`, and like multiple functions wanting to return `PermissionGrant` so filling in `is_alias` with a dummy false

grouper/service_account.py
rra: Oh, good catch! And I think the SQLAlchemy model objects aren't fully typed; I've seen it miss other things like that.

grouper/service_account.py
rra: Yeah, you're right, this is ugly. This made me finally (after a couple of tries) realize that `PermissionGrant` is indeed superfluous and we can use `GroupPermissionGrant` and `ServiceAccountPermissionGrant` everywhere without too many difficulties. I'll push another patch on this PR that does that.

I think the `is_alias` information is real -- it's just that we have a lot of code that isn't alias-aware yet, so there are no paths to add the aliases. But over time that attribute will become more useful.

PR #553, opened 2019-05-24
Unify some duplicate graph data structures

rra:
Add new private _groups and _permissions dicts that map group and
permission names to Group and Permission objects, and use them to
satisfy requests for the groups and permissions properties on the
fly.  Replace group_tuples, permission_tuples, and permission_metadata
with those new data structures.

Replace the separate users data structure with a property that
generates the list on demand from user_metadata.

Return name and email address metadata about a group as part of
get_group_details, and include email_address in the Group data transfer
object to support this.  This eliminates the separate group_metadata
data structure.

Mark more graph data structures that are not used outside the graph
code as private to make the API cleaner and easier to understand.

These changes eliminate three more database queries from the graph
construction code.

### DIFF
--- a/grouper/graph.py
+++ b/grouper/graph.py
@@ -68,6 +68,10 @@ class GroupGraph(object):
     groups and service accounts to the permission grants they have, and the internal graph and
     rgraph directed graphs.
 
+    The cached user metadata includes metadata for disabled users, and the disabled_groups internal
+    attribute holds a dictionary of group names to Group named tuples for disabled groups.  Other
+    than those exceptions, stored graph metadata includes only enabled objects.
+
     The directed graphs are used to calculate inherited membership and permissions.  _graph is a
     directed graph with all users and groups as nodes, and with directed edges from groups to their
     members (whether users or other groups).  _rgraph is the same graph reversed, so the edges
@@ -84,6 +88,7 @@ class GroupGraph(object):
         permissions: Names of all enabled permissions
         user_metadata: Full information about each user
         permission_grants: Permission grant information for users
+
     """
 
     def __init__(self):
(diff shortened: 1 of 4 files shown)


grouper/graph.py
```diff
@@ -60,46 +60,72 @@ class NoSuchGroup(Exception):
 class GroupGraph(object):
     """The cached permission graph.
 
+    The graph is internally represented by four major components: the users, groups, and
```
wrodriguez-dbx: Nit: Maybe note that none of these components include disabled anything? In general I think getting more explicit about how we handle disabled things is a good idea, since right now I have very low confidence that we're getting it all right

grouper/graph.py
```diff
@@ -60,46 +60,72 @@ class NoSuchGroup(Exception):
...
+        # complicated object, which hasn't been written yet.
         self.user_metadata = {}  # type: Dict[str, Dict[str, Any]]
-        self.group_metadata = {}  # type: Dict[str, Dict[str, Any]]
-        self.group_service_accounts = {}  # type: Dict[str, List[str]]
+
+        # Collection of all disabled groups.
+        self._disabled_groups = set()  # type: Set[Group]
+
+        # Map of groups to their permission grants.
         self.permission_grants = {}  # type: Dict[str, List[GroupPermissionGrant]]
```
wrodriguez-dbx: Should this be made private as well? Doesn't seem to be used directly anywhere else, unless I'm missing something

grouper/graph.py
```diff
@@ -60,46 +60,72 @@ class NoSuchGroup(Exception):
...
+    owns; the group_permission_grants and service_account_permission_grants dictionaries that map
+    groups and service accounts to the permission grants they have, and the internal graph and
+    rgraph directed graphs.
+
+    The directed graphs are used to calculate inherited membership and permissions.  _graph is a
+    directed graph with all users and groups as nodes, and with directed edges from groups to their
+    members (whether users or other groups).  _rgraph is the same graph reversed, so the edges
+    point from users and groups to the groups of which they are a member.
+
+    Finally, disabled_groups is a sorted list of Group named tuples for all disabled groups.
```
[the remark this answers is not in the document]
rra: Yeah, probably no reason not to. It consumes a bit of extra, unnecessary memory, but the consistency seems worthwhile. Will push another commit that does this.

grouper/graph.py
rra: Good idea. Added some additional commentary about that.

grouper/graph.py
rra: It's used in tests, so I was deferring cleanup of that until I could figure out a better way to implement those tests. I've added that to my to-do list for cleanups.

PR #554, opened 2019-05-29
Add API for retrieving permission grants

rra:
/grants will return all permission grants in a JSON dict with the
permissions as keys, followed by users or service_accounts, followed
by the name of the user or service account, followed by a list of
arguments.

/grants/<permission> will return the same for a specific permission,
without the first layer of dict but with a separate top-level key of
permission, indicating the requested permission.

The test uses the low-level API directly since Groupy doesn't have
high-level support for grants.  (It's not clear yet whether it's
worth the trouble of adding that.)

### DIFF
--- a/grouper/api/routes.py
+++ b/grouper/api/routes.py
@@ -6,6 +6,7 @@ additional arguments will be injected when the Tornado Application object is cre
 """
 
 from grouper.api.handlers import (
+    Grants,
     Groups,
     MultiUsers,
     NotFound,
@@ -19,17 +20,19 @@ from grouper.constants import NAME_VALIDATION, PERMISSION_VALIDATION
 from grouper.handlers.health_check import HealthCheck
 
 HANDLERS = [
-    (r"/users", Users),
-    (r"/users/{}".format(NAME_VALIDATION), Users),
-    (r"/token/validate".format(NAME_VALIDATION), TokenValidate),
-    (r"/public-keys", UsersPublicKeys),
+    (r"/debug/health", HealthCheck),
+    (r"/grants", Grants),
+    (r"/grants/{}".format(PERMISSION_VALIDATION), Grants),
     (r"/groups", Groups),
     (r"/groups/{}".format(NAME_VALIDATION), Groups),
     (r"/permissions", Permissions),
     (r"/permissions/{}".format(PERMISSION_VALIDATION), Permissions),
+    (r"/public-keys", UsersPublicKeys),
     (r"/service_accounts", ServiceAccounts),
     (r"/service_accounts/{}".format(NAME_VALIDATION), ServiceAccounts),
+    (r"/users", Users),
+    (r"/users/{}".format(NAME_VALIDATION), Users),
     (r"/multi/users", MultiUsers),
-    (r"/debug/health", HealthCheck),
+    (r"/token/validate".format(NAME_VALIDATION), TokenValidate),
     (r"/.*", NotFound),
 ]
--- a/grouper/api/routes.py
+++ b/grouper/api/routes.py
@@ -33,6 +33,6 @@ HANDLERS = [
     (r"/users", Users),
     (r"/users/{}".format(NAME_VALIDATION), Users),
     (r"/multi/users", MultiUsers),
-    (r"/token/validate".format(NAME_VALIDATION), TokenValidate),
+    (r"/token/validate", TokenValidate),
     (r"/.*", NotFound),
 ]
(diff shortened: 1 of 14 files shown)


grouper/api/routes.py
```diff
@@ -19,17 +20,19 @@
...
     (r"/permissions", Permissions),
     (r"/permissions/{}".format(PERMISSION_VALIDATION), Permissions),
+    (r"/public-keys", UsersPublicKeys),
     (r"/service_accounts", ServiceAccounts),
     (r"/service_accounts/{}".format(NAME_VALIDATION), ServiceAccounts),
+    (r"/users", Users),
+    (r"/users/{}".format(NAME_VALIDATION), Users),
     (r"/multi/users", MultiUsers),
-    (r"/debug/health", HealthCheck),
+    (r"/token/validate".format(NAME_VALIDATION), TokenValidate),
```
wrodriguez-dbx: Not from this diff, but does this `format` do anything?

grouper/entities/permission_grant.py
```diff
@@ -23,3 +23,15 @@
...
+# have been granted that permission to lists of all the arguments to that permission that they have
+# been granted.
+AllGrantsOfPermission = NamedTuple(
+    "AllGrantsOfPermission",
+    [("users", Dict[str, List[str]]), ("service_accounts", Dict[str, List[str]])],
+)
+
+# The same, but for every permission, stored in a dictionary mapping permission names to those
+# grants.
+AllGrants = Dict[str, AllGrantsOfPermission]
```
wrodriguez-dbx: I have a vague feeling that this definition is overkill, probably since it can be described entirely (and cleanly) as a type annotation. By contrast, the definition of `AllGrantsOfPermission` does feel valuable since it a) does provide additional information (the key names) and b) would be very ugly if expanded into a type annotation everywhere.

grouper/graph.py
```diff
@@ -123,6 +123,9 @@ def __init__(self):
         # Map of groups to their permission grants.
         self.permission_grants = {}  # type: Dict[str, List[GroupPermissionGrant]]
```
wrodriguez-dbx: Nit: renaming to something like `grants_by_group` would help avoid confusion with `_grants_by_permission`. Also I think the class comment is incorrect ("`permission_grants: Permission grant information for users`")

grouper/entities/permission_grant.py
```diff
@@ -23,3 +23,15 @@
         ("grant_id", Optional[int]),
     ],
 )
+
+# Represents all grants of a specific permission, as a dictionary of users and service accounts who
+# have been granted that permission to lists of all the arguments to that permission that they have
+# been granted.
+AllGrantsOfPermission = NamedTuple(
```
wrodriguez-dbx: One thing that I think is a bit confusing -- but that I don't have any suggestion on how to improve -- is the use of "grant" to both mean the details of the grant itself, as well as to just represent that a user/service account/group has a permission. So under the former definition, a user with the same (permission, argument) through two different groups has two grants, whereas under the latter it only has one.

grouper/graph.py
```diff
@@ -436,6 +444,62 @@ def _get_edges(session):
...
+        service_grants = defaultdict(
+            lambda: defaultdict(set)
+        )  # type: Dict[str, Dict[str, Set[str]]]
+        for account, service_grant_list in iteritems(service_account_grants):
+            for service_grant in service_grant_list:
+                service_grants[service_grant.permission][account].add(service_grant.argument)
+
+        # For each group that has a permission grant, determine all of its users from the graph,
+        # and then record each permission grant of that group as a grant to all of those users.
+        # This ensures that each permission grant is recorded for all affected users exactly once.
```
wrodriguez-dbx: Teeny tiny nit: the "exactly" comes from the use of sets rather than the algorithm (as the comment sort of implies) I believe

grouper/api/routes.py
rra: Good catch. I'll add a diff on to this PR that gets rid of that.

grouper/entities/permission_grant.py
rra: Yup, good call. I've now gotten rid of `AllGrants`.

grouper/entities/permission_grant.py
rra: This is an interesting point. I thought about it for a bit and then renamed this to `UniqueGrantsOfPermission`. Hopefully the "unique" part will hint at that difference: this isn't every grant, just the ones that provide unique permissions.

grouper/graph.py
rra: I renamed this to `_group_grants` (to parallel `_service_account_grants`) and made it private and fixed the tests that were using it.

grouper/graph.py
rra: Thanks, fixed!

PR #568, opened 2019-07-24
Improve log_exception plugin hook, use it

rra:
This hook currently isn't used, so before we add an initial use of
it, clean up the types and parameters.  Pass in the full tuple that
Python normally includes with exceptions, in addition to the request
and status if any, and allow the request and status to be omitted
for non-Tornado exceptions.

Remove Sentry support in the background processor and instead just
call the log_exception plugins.

Add a log_exception handler that logs via the plugin log_exception
method and then still runs the normal Tornado exception logging.

Cleaning up remaining Sentry support will be in a subsequent diff.

### DIFF
--- a/grouper/api/handlers.py
+++ b/grouper/api/handlers.py
@@ -29,7 +29,8 @@ if TYPE_CHECKING:
     from grouper.entities.user import User
     from grouper.graph import GroupGraph
     from grouper.usecases.factory import UseCaseFactory
-    from typing import Any, Dict, Iterable, Optional, Tuple
+    from types import TracebackType
+    from typing import Any, Dict, Iterable, Optional, Tuple, Type
 
 
 def get_individual_user_info(handler, name, service_account):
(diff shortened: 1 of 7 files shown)


tests/exception_test.py
```diff
@@ -0,0 +1,70 @@
...
+    ):
+        # type: (...) -> None
+        self.request = request
+        self.status = status
+        self.exc_type = exc_type
+        self.exc_value = exc_value
+        self.exc_tb = exc_tb
+
+
+def test_exception_plugin():
```
wrodriguez-dbx: Is anything specific to exception logging actually being tested here? It seems that there's only one piece of logic being run that isn't defined in this file, which is how `proxy.log_exception` does in fact call `plugin.log_exception` for the one passed in `plugin`.

grouper/api/handlers.py
```diff
@@ -93,6 +94,20 @@ def on_finish(self):
...
         stats.log_rate("response_status_{}_{}".format(self.__class__.__name__, response_status), 1)
 
+    def log_exception(
+        self,
+        exc_type,  # type: Optional[Type[BaseException]]
+        exc_value,  # type: Optional[BaseException]
+        exc_tb,  # type: Optional[TracebackType]
+    ):
+        # type: (...) -> None
+        if isinstance(exc_value, HTTPError):
```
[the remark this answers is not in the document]
rra: Using `exc_type` confuses mypy because it doesn't know to infer that `exc_value` therefore has a `status_code` element. Yes, there's no purpose in including `exc_type` other than that Python always does, but that seemed like a good enough reason since this is an external API.

tests/exception_test.py
rra: It isn't currently. :( I'd like it to really test exception handling, but haven't yet been able to figure out a good way to do so. The name of the test is therefore somewhat aspirational.

rra:
Going to merge this for now with the below-standard test suite and hope that I can make it better later.

PR #569, opened 2019-07-24
Bump networkx version

rra:
Eliminates a warning during tests.


wrodriguez-dbx:
Yea `This is our first Python 3 only release` :/

https://networkx.github.io/documentation/stable/news.html

rra:
Yeah, indeed. We can't do this until we go to pure Python 3.

PR #570, opened 2019-07-24
Remove Sentry support

rra:
Catch a few more places where exceptions were being logged only to
Sentry and log them to the plugin infrastructure instead.

### DIFF
--- a/grouper/api/handlers.py
+++ b/grouper/api/handlers.py
@@ -7,11 +7,10 @@ from datetime import datetime
 from typing import TYPE_CHECKING
 
 from six import iteritems, StringIO
-from tornado.web import HTTPError
+from tornado.web import HTTPError, RequestHandler
 
 from grouper import stats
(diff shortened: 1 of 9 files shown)


grouper/api/handlers.py
```diff
@@ -68,11 +68,12 @@ def get_individual_user_info(handler, name, service_account):
         return out
 
 
-class GraphHandler(SentryHandler):
+class GraphHandler(RequestHandler):
     def initialize(self, *args, **kwargs):
         # type: (*Any, **Any) -> None
         self.graph = kwargs["graph"]  # type: GroupGraph
         self.usecase_factory = kwargs["usecase_factory"]  # type: UseCaseFactory
+        self.plugins = kwargs["plugins"]  # type: PluginProxy
```
wrodriguez-dbx: Oh huh, nice catch

PR #572, opened 2019-08-07
Refactor service account creation tests

rra:
In preparation for refactoring service account creation, switch the
existing service account creation tests to the new syntax.

### DIFF
--- a/itests/fe/service_accounts_test.py
+++ b/itests/fe/service_accounts_test.py
@@ -1,6 +1,6 @@
 from typing import TYPE_CHECKING
 
-from itests.fixtures import async_server  # noqa: F401
+from grouper.constants import USER_ADMIN
 from itests.pages.groups import GroupViewPage
 from itests.pages.service_accounts import (
     ServiceAccountCreatePage,
@@ -10,15 +10,6 @@ from itests.pages.service_accounts import (
 )
 from itests.pages.users import UsersViewPage
 from itests.setup import frontend_server
-from tests.fixtures import (  # noqa: F401
-    graph,
-    groups,
-    permissions,
-    service_accounts,
-    session,
-    standard_graph,
-    users,
-)
 from tests.url_util import url
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


itests/fe/service_accounts_test.py
```diff
@@ -27,37 +18,41 @@
     from tests.setup import SetupTest
 
 
-def test_service_account_lifecycle(async_server, browser):  # noqa: F811
-    browser.get(url(async_server, "/groups/user-admins"))
-
-    page = GroupViewPage(browser)
-    page.click_add_service_account_button()
+def test_service_account_lifecycle(tmpdir, setup, browser):
```
wrodriguez-dbx: Not review, just a general question:

What are your thoughts on basing tests on real workflows like this? It's very natural and I think tends to end up reading very nicely, but I've always stayed away from them to prioritize being comprehensive. So here for example, there are a bunch of small corner cases that aren't exercised.

tests/ctl/service_account_test.py
```diff
@@ -1,127 +1,151 @@
...
+    assert service_account is not None
+    assert service_account.user.name == "good-service@a.co"
+    assert service_account.machine_set == "foo +bar -(org)"
+    assert service_account.description == "this is a service account.\n\n it is for testing"
+    group = Group.get(setup.session, name="some-group")
+    assert get_service_accounts(setup.session, group) == [service_account]
+
+    # If the account already exists, creating it again does nothing.
+    #
+    # TODO(rra): Change this behavior to return an error message instead, since it's confusing.
```
wrodriguez-dbx: Also general question rather than review:

Do you have a strategy for actually go backing and hitting TODOs? I avoid leaving them in final checked-in code because I'm pretty sure I'll just forget about them, but they do seem very useful in certain cases

itests/fe/service_accounts_test.py
rra: I like them. My preferred style is to have one of these tests and then separate tests for the interesting edge cases. The drawback is that the Selenium tests are very slow, so ideally as much of the edge case testing as possible is done at the usecase test level, and then the Selenium test only tests the full workflow happy path and then any UI-specific edge cases.

In general, though, I'm happy to duplicate test coverage to have both edge-case and full workflow tests.

tests/ctl/service_account_test.py
rra: In general, I agree with you most of the time. In this specific case, subsequent diffs in this same patch series are going to fix the TODOs, so they won't be there long.

I will sometimes leave long-lived TODOs in the code not because I have any specific plans to go back and fix them but because I want to make explicit to someone reading the code that this bit of code is an error, or at least sub-optimal, and just hasn't been fixed. Otherwise, it can be hard to be sure whether the code is actually ideal because of some subtlety that isn't obvious. One can explain that in English, but TODO makes it totally clear that no, this is something that should be fixed, but there just hasn't been time yet.

PR #573, opened 2019-08-13
Add use case for creating a service account

rra:
Implement creating a service account as a use case and add tests
for the success case and the various error conditions.

Convert the grouper-ctl service_account create command to the new
usecase and the new grouper-ctl command architecture, and update
its tests accordingly.

### DIFF
--- a/grouper/repositories/permission_grant.py
+++ b/grouper/repositories/permission_grant.py
@@ -55,6 +55,22 @@ class GraphPermissionGrantRepository(PermissionGrantRepository):
         # type: (str) -> List[ServiceAccountPermissionGrant]
         return self.repository.service_account_grants_for_permission(name)
 
+    def permission_grants_for_service_account(self, name):
+        # type: (str) -> List[ServiceAccountPermissionGrant]
+        user_details = self.graph.get_user_details(name)
+        permissions = []
+        for permission_data in user_details["permissions"]:
+            permission = ServiceAccountPermissionGrant(
+                service_account=name,
+                permission=permission_data["permission"],
+                argument=permission_data["argument"],
+                granted_on=datetime.utcfromtimestamp(permission_data["granted_on"]),
+                is_alias=permission_data["alias"],
+                grant_id=None,
+            )
+            permissions.append(permission)
+        return permissions
+
     def permission_grants_for_user(self, name):
         # type: (str) -> List[GroupPermissionGrant]
         user_details = self.graph.get_user_details(name)
@@ -78,6 +94,13 @@ class GraphPermissionGrantRepository(PermissionGrantRepository):
     def revoke_all_service_account_grants(self, permission):
         # type: (str) -> List[ServiceAccountPermissionGrant]
         return self.repository.revoke_all_service_account_grants(permission)
+
+    def service_account_has_permission(self, service, permission):
+        # type: (str, str) -> bool
+        for grant in self.permission_grants_for_service_account(service):
+            if permission == grant.permission:
+                return True
+        return False
 
     def user_has_permission(self, user, permission):
         # type: (str, str) -> bool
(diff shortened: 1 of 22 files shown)


grouper/repositories/permission_grant.py
```diff
@@ -79,6 +95,13 @@ def revoke_all_service_account_grants(self, permission):
         # type: (str) -> List[ServiceAccountPermissionGrant]
         return self.repository.revoke_all_service_account_grants(permission)
 
+    def service_account_has_permission(self, service, permission):
+        # type: (str, str) -> bool
+        for grant in self.permission_grants_for_service_account(service):
```
wrodriguez-dbx: Guessing this is little enough of a bottleneck that it isn't worth making the optimizations of replicating the loop from `permission_grants_for_service_account` to short-circuit and have only a single loop?

grouper/repositories/permission_grant.py
```diff
@@ -175,6 +198,37 @@ def service_account_grants_for_permission(self, name):
             for g in grants.all()
         ]
 
+    def permission_grants_for_service_account(self, name):
+        # type: (str) -> List[ServiceAccountPermissionGrant]
+        """Return all permission grants for a service account.
+
+        TODO(rra): Currently does not expand permission aliases, and therefore doesn't match the
+        graph behavior.  Use with caution until that is fixed.
```
wrodriguez-dbx: Is there any correct usage of function? If not, feels ideal to actually guarantee this doesn't get used accidentally

grouper/usecases/create_service_account.py
```diff
@@ -0,0 +1,124 @@
...
+            self.ui.create_service_account_failed_invalid_name(service, error)
+            return
+
+        # Creation is permitted if the actor is a user admin or if the actor is a user (not a
+        # service account) and is a member of the owning group.
+        if self.service_account_service.service_account_exists(self.actor):
+            allowed = self.service_account_service.service_account_is_user_admin(self.actor)
+        else:
+            allowed = owner in self.user_service.groups_of_user(self.actor)
+            if not allowed:
```
wrodriguez-dbx: Ah, that's one solution to multi-line boolean expressions being ugly in Python :)

grouper/repositories/permission_grant.py
```diff
@@ -55,6 +55,22 @@ def service_account_grants_for_permission(self, name):
...
+        user_details = self.graph.get_user_details(name)
+        permissions = []
+        for permission_data in user_details["permissions"]:
+            permission = ServiceAccountPermissionGrant(
+                service_account=name,
+                permission=permission_data["permission"],
+                argument=permission_data["argument"],
+                granted_on=datetime.utcfromtimestamp(permission_data["granted_on"]),
+                is_alias=permission_data["alias"],
+                grant_id=None,
```
wrodriguez-dbx: I feel like I asked this on a previous PR, but do we have a sense for when/how `grant_id` is necessary and why it's okay to not set it if we're using a graph-backed repo?

grouper/repositories/permission_grant.py
rra: Yeah, that was my thought process. We can always make it faster later if it looks like it matters. I'm hoping before that happens we modify the graph to store the same data transfer objects, so then it's a simple loop either way.

grouper/repositories/permission_grant.py
rra: It's used by `grouper-ctl` to check whether a service account has the `USER_ADMIN` permission and thus can create other service accounts. The only thing it doesn't support is permission aliases, which are a graph-only feature (which means they're just in general not supported by a bunch of stuff right now).

grouper/repositories/permission_grant.py
rra: It's used by the frontend to uniquely identify a grant, since otherwise it has to be represented by the tuple of (group, permission, argument) or (service account, permission, argument), which is painful to work with in URLs. I think we'll eventually need to store it in the graph as well (probably by caching the full `ServiceAccountPermissionGrant` objects in the graph). This is just a temporary workaround to avoid doing too much work now.

PR #575, opened 2019-08-16
Allow overriding the database source command

rra:
We want to stop having every Grouper binary call the same piece of
code to get the database URL so that we can stage rollouts of changes
to the database URL handling as well.  However, given our internal
deployment mechanisms, this means that the path to the command to run
to get the database URL will be different for each component.  To
avoid needing four different config files (eight for stage support),
add a --database-source command-line option to override the
database_source setting in the config and specify a command to run.

grouper-background didn't previously support overriding the database
URL with a command-line flag, so add that.

This doesn't include tests because setting up a proper test of
something that runs an external program is irritatingly difficult,
and I'm bailing on that for now.

### DIFF
--- a/grouper/api/main.py
+++ b/grouper/api/main.py
@@ -52,6 +52,8 @@ def start_server(args, settings, plugins):
     logging.debug("configure database session")
     if args.database_url:
         settings.database = args.database_url
+    elif args.database_source:
+        settings.database_source = args.database_source
     Session.configure(bind=get_db_engine(settings.database))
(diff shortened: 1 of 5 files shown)


rra:
Adding separate commands to the per-service stanzas is an entirely superior approach to solving the problem that I had, so in the name of not adding code for which we have no use case, I'm going to abandon this PR. We can revisit common option parsing later when we either get there in the refactor or have another use case.

PR #576, opened 2019-08-19
Fix audit view performance

cauthu:
Introduce `get_group_audit_members_infos()` in to `grouper/audit.py` that uses a small number of queries to collect all the audit information about a group's members that the frontend needs to display the audit modal. Avoid using `AuditMember`'s `member()` property because displaying a large group would require several queries per member of the group; in fact, remove that business logic function altogether (as well as `my_members()`) from the AuditMember model.

Also, use a dedicated query to count a group's members that still have pending audit status, instead of retrieving all the members and then checking their audit statuses.

### DIFF
--- a/grouper/audit.py
+++ b/grouper/audit.py
@@ -227,8 +227,7 @@ def get_group_audit_members_infos(session, group):
             AuditMember.edge_id == GroupEdge.id,
             GroupEdge.member_type == OBJ_TYPES["User"],
             GroupEdge.member_pk == User.id,
-            # but only those members who are still current members of the group, because members
-            # that need audits might have left the group after the audit was started
+            # only those members who have not left the group after the audit started
             AuditMember.edge_id.in_(members_edge_ids),
         )
         .all()
@@ -241,8 +240,7 @@ def get_group_audit_members_infos(session, group):
             AuditMember.edge_id == GroupEdge.id,
             GroupEdge.member_type == OBJ_TYPES["Group"],
             GroupEdge.member_pk == Group.id,
-            # but only those members who are still current members of the group, because members
-            # that need audits might have left the group after the audit was started
+            # only those members who have not left the group after the audit started
             AuditMember.edge_id.in_(members_edge_ids),
         )
         .all()
@@ -256,25 +254,23 @@ def get_group_audit_members_infos(session, group):
     ]
 
 
-def get_group_audit_members_count(session, group, status):
-    # type: (Session, Group, str) -> int
-    """Count the memberships of a group with a specific audit status
+def group_has_pending_audit_members(session, group):
+    # type: (Session, Group) -> bool
+    """Check if a group still has memberships with "pending" audit status
 
     Arg(s):
         session: The SQL session
         group: The group
-        status: The type of audit status to count.
 
     Return:
-        The number of memberships of a group with a specific audit status.
+        True if the group still has memberships with "pending" audit status
 
     """
     members_edge_ids = {member.edge_id for member in itervalues(group.my_members())}
     audit_members_statuses = session.query(AuditMember.status).filter(
         AuditMember.audit_id == group.audit_id,
-        AuditMember.status == status,
-        # but only those members who are still current members of the group, because members that
-        # need audits might have left the group after the audit was started
+        AuditMember.status == "pending",
+        # only those members who have not left the group after the audit started
         AuditMember.edge_id.in_(members_edge_ids),
     )
     return audit_members_statuses.count()
--- a/grouper/audit.py
+++ b/grouper/audit.py
@@ -1,19 +1,23 @@
-from typing import TYPE_CHECKING
+import itertools
+from typing import List, NamedTuple, TYPE_CHECKING, Union
 
-from six import iteritems
+from six import iteritems, itervalues
 
 from grouper.constants import PERMISSION_AUDITOR
 from grouper.graph import Graph, NoSuchGroup
 from grouper.models.audit import Audit
+from grouper.models.audit_member import AuditMember
+from grouper.models.base.constants import OBJ_TYPES
 from grouper.models.group import Group
+from grouper.models.group_edge import GroupEdge
+from grouper.models.user import User
 from grouper.util import get_auditors_group_name
 
 if TYPE_CHECKING:
     from grouper.models.base.session import Session
-    from grouper.models.user import User
     from grouper.settings import Settings
     from sqlalchemy.orm.query import Query
-    from typing import Set, Union
+    from typing import Set
 
 
 class UserNotAuditor(Exception):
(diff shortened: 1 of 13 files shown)


cauthu:
for manual testing:
* make the db engine echo the queries
* make a small change to the group_view.py handler to move the get template variables call before the render call, and
* add a few logging statements to demarcate the render call.

in master:
```
(merou-venv) > cauthu @ cauthu-dbx:merou$ git diff
diff --git grouper/fe/handlers/group_view.py grouper/fe/handlers/group_view.py
index f4da65c..faecef2 100644
--- grouper/fe/handlers/group_view.py
+++ grouper/fe/handlers/group_view.py
@@ -23,8 +23,12 @@ class GroupView(GrouperHandler):
         if is_role_user(self.session, group=group):
             return self.redirect("/service/{}".format(group.groupname))
 
+        gvtv = get_group_view_template_vars(self.session, self.current_user, group, self.graph)
+        import logging
+        logging.warn('___ no fix begin render')
         self.render(
             "group.html",
             group=group,
-            **get_group_view_template_vars(self.session, self.current_user, group, self.graph)
+            **gvtv,
         )
+        logging.warn('___ no fix done render')
diff --git grouper/models/base/session.py grouper/models/base/session.py
index 3b171f4..2d1cce1 100644
--- grouper/models/base/session.py
+++ grouper/models/base/session.py
@@ -31,7 +31,7 @@ def flush_transaction(method):
 
 def get_db_engine(url):
     # type: (str) -> Engine
-    return create_engine(url, pool_recycle=300)
+    return create_engine(url, pool_recycle=300, echo=True)
 
 
 class SessionWithoutAdd(_Session):
(merou-venv) > cauthu @ cauthu-dbx:merou$
```

in the fix branch:
```
(merou-venv) > cauthu @ cauthu-dbx:merou$ git diff
diff --git grouper/fe/handlers/group_view.py grouper/fe/handlers/group_view.py
index a33fc87..efa2d9e 100644
--- grouper/fe/handlers/group_view.py
+++ grouper/fe/handlers/group_view.py
@@ -24,9 +24,13 @@ class GroupView(GrouperHandler):
         if is_role_user(self.session, group=group):
             return self.redirect("/service/{}".format(group.groupname))
 
+        gvtv = get_group_view_template_vars(self.session, self.current_user, group, self.graph)
+        import logging
+        logging.warn('___ with fix begin render')
         self.render(
             "group.html",
             group=group,
             audit_members_infos=get_group_audit_members_infos(self.session, group),
-            **get_group_view_template_vars(self.session, self.current_user, group, self.graph)
+            **gvtv,
         )
+        logging.warn('___ with fix done render')
diff --git grouper/models/base/session.py grouper/models/base/session.py
index 3b171f4..2d1cce1 100644
--- grouper/models/base/session.py
+++ grouper/models/base/session.py
@@ -31,7 +31,7 @@ def flush_transaction(method):
 
 def get_db_engine(url):
     # type: (str) -> Engine
-    return create_engine(url, pool_recycle=300)
+    return create_engine(url, pool_recycle=300, echo=True)
 
 
 class SessionWithoutAdd(_Session):
(merou-venv) > cauthu @ cauthu-dbx:merou$ 
```

----
then load a group with an audit involving 6 members
image

then count the number of sql queries that happen during the render. 
```
(merou-venv) > cauthu @ cauthu-dbx:merou$ sed -n "/__ no fix begin render/,/__ no fix done render/p" before.txt | grep -c "INFO sqlalchemy.engine.base.Engine SELECT"
26
(merou-venv) > cauthu @ cauthu-dbx:merou$ 
(merou-venv) > cauthu @ cauthu-dbx:merou$ sed -n "/__ with fix begin render/,/__ with fix done render/p" after.txt | grep -c "INFO sqlalchemy.engine.base.Engine SELECT"
7
(merou-venv) > cauthu @ cauthu-dbx:merou$ 
```

grouper/audit.py
```diff
@@ -185,3 +201,80 @@ def get_auditors_group(settings, session):
...
+
+    return [
+        AuditMemberInfo(audit_member, audit_member_role, member_obj)
+        for audit_member, audit_member_role, member_obj in itertools.chain(
+            user_members, group_members
+        )
+    ]
+
+
+def get_group_audit_members_count(session, group, status):
```
wrodriguez-dbx: How about making this something like `group_has_pending_audit_members`? Since there's only the single callsite, making it generic across statuses and returning the full count seems a bit unnecessary

grouper/fe/handlers/audits_complete.py
```diff
@@ -31,37 +31,37 @@ def post(self, audit_id):
             if argument.startswith("audit_"):
                 edges[int(argument.split("_")[1])] = self.request.arguments[argument][0].decode()
 
-        for member in audit.my_members():
-            if member.id in edges:
+        for ami in get_group_audit_members_infos(self.session, audit.group):
```
wrodriguez-dbx: In the spirit of the new architecture and its extreme verbosity, personal preference for `ami` -> `audit_member_info` :)

grouper/fe/handlers/audits_complete.py
```diff
@@ -31,37 +31,37 @@ def post(self, audit_id):
...
+            if ami.audit_member_obj.id in edges:
                 # You can only approve yourself (otherwise you can remove yourself
                 # from the group and leave it ownerless)
-                if member.member.id == self.current_user.id:
```
wrodriguez-dbx: Well I'm happy with this change to remove `member.member` being a thing if nothing else

grouper/fe/templates/group.html
```diff
@@ -207,19 +207,6 @@ <h4 class="modal-title">Remove User</h4>
 {% if group.audit and not group.audit.complete %}
 {# TODO(herb): audit db calls in this block #}
```
wrodriguez-dbx: Seems like this can be removed 🎉

tests/audit_test.py
```diff
@@ -170,15 +171,33 @@ def test_audit_end_to_end(session, users, groups, http_client, base_url, graph):
     assert groupname in [x.group.name for x in open_audits], "group we expect also gets audit"
 
     # pull all the info we need to resolve audits, avoids detached sqlalchemy sessions
-    AuditMember = namedtuple("AuditMember", "am_id, edge_type, edge_id")
-    Audit = namedtuple("Audit", "audit_id, owner_name, group_name, audit_members")
+    # (DetachedInstanceError)
+    MyAuditMemberInfo = NamedTuple(
+        "MyAuditMemberInfo", [("am_id", int), ("edge_type", int), ("edge_id", int)]
```
wrodriguez-dbx: Nit: Preference for `am_id` -> `member_id` for clarity

grouper/audit.py
```diff
@@ -185,3 +201,80 @@ def get_auditors_group(settings, session):
...
+    user_members = (
+        session.query(AuditMember, GroupEdge._role, User)
+        .filter(
+            AuditMember.audit_id == group.audit_id,
+            AuditMember.edge_id == GroupEdge.id,
+            GroupEdge.member_type == OBJ_TYPES["User"],
+            GroupEdge.member_pk == User.id,
+            # but only those members who are still current members of the group, because members
+            # that need audits might have left the group after the audit was started
+            AuditMember.edge_id.in_(members_edge_ids),
```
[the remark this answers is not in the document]
cauthu: indeed that is the behavior we want. will update.

grouper/audit.py
cauthu: sgtm

grouper/fe/handlers/audits_complete.py
cauthu: hah

grouper/fe/templates/group.html
cauthu: :tada:

tests/audit_test.py
cauthu: no, member_id will be confusing with the id of the User/Group object. am_id is clearer that its the id of the audit_member.

cauthu:
yea working on the conversion to usecase. kinda non-trivial heh

PR #577, opened 2019-08-21
Add plugin support for checking service account names

rra:
New plugin callback to provide an opportunity to reject service
account names.  Pass the prospective owner into the plugin in case
the plugin wants to enforce any owner-based naming scheme, which
requires plumbing the owner through in a few other places.

Clean up some use of mocks in the usecase test.

Update comments in BasePlugin to use a standard comment style,
and move configure to the top of the method list since it's more of an
initialization call than a part of the functionality.

### DIFF
--- a/grouper/initialization.py
+++ b/grouper/initialization.py
@@ -32,7 +32,7 @@ def create_graph_usecase_factory(
     if not session_factory:
         session_factory = SessionFactory(settings)
     repository_factory = GraphRepositoryFactory(settings, plugins, session_factory, graph)
-    service_factory = ServiceFactory(settings, repository_factory)
+    service_factory = ServiceFactory(settings, plugins, repository_factory)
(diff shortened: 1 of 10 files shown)


rra:
Yeah, to be clear I have zero plan to adopt owner-based naming schemes here at Dropbox. It just seemed easy enough to pass in the owner even though we have no plans to use it, and there didn't seem to be a reason not to do so, just in case some other user of Grouper wanted to use that as additional information in making the naming decision.

wrodriguez-dbx:
Fair. I'd argue there's a slight reason not to do so, which is that we'd be implying a guarantee we aren't actually making: that it's impossible to introduce a new violation of the service account name constraints defined by the plugin

rra:
Hm, yeah, actually, that's persuasive. I'll put up another PR to drop this again.

PR #581, opened 2019-09-03
Fix audit modal on service account view

cauthu:
The service.html template was updated in https://github.com/dropbox/merou/pull/576, but the
role_user_view.py was erroneously not updated to pass the new variable audit_members_infos to
it. Now we do it properly.


cauthu:
> Stamping to unbreak. Also a good impetus to add a test? :)

yea but i think the right way to test this is with selenium in an integration test. and then also, we are getting rid of old-style service accounts.

for those reasons i think it's not worth building a test. thoughts?

wrodriguez-dbx:
> i think the right way to test this is with selenium in an integration test. 
Yea -- luckily we already have the itest framework for the [service account page](https://github.com/dropbox/merou/blob/master/itests/pages/service_accounts.py), so tacking on the audit modal would be (hopefully) not a huge amount of work

> and then also, we are getting rid of old-style service accounts.
True, though given that we've been saying this since forever and still don't have any hard plans to kill them, probably safe to assume old-style role users will be around for a while longer :P

cauthu:
Yea ok will create an itest

PR #584, opened 2019-10-05
Change the stats logging API

cauthu:
Instead of generic `log_rate()` and `log_gauge()` functions, use specific stats reporting functions like `log_request()` for reporting information about a handled request, `log_graph_update_duration()` for reporting a graph update latency.

Add two tests to fe and api handler tests for ensuring the correct reporting of the request metadata.


grouper/background/background_processor.py
```diff
@@ -147,11 +147,9 @@ def run(self):
 
                     session.commit()
 
-                stats.log_gauge("successful-background-update", 1)
-                stats.log_gauge("failed-background-update", 0)
+                stats.log_background_run(True)
```
rra: `success=True` here so that it's obvious what the boolean is for. `self.plugins.log_background_run`

grouper/background/background_processor.py
```diff
@@ -147,11 +147,9 @@ def run(self):
...
             except Exception:
-                stats.log_gauge("successful-background-update", 0)
-                stats.log_gauge("failed-background-update", 1)
+                stats.log_background_run(False)
```
rra: `success=False`

grouper/database.py
```diff
@@ -43,11 +43,9 @@ def run(self):
                 with closing(Session()) as session:
                     self.graph.update_from_db(session)
 
-                stats.log_gauge("successful-db-update", 1)
-                stats.log_gauge("failed-db-update", 0)
+                stats.log_periodic_graph_update(True)
```
rra: `success=True`, `get_plugin_proxy().log_periodic_graph_update`

grouper/database.py
```diff
@@ -43,11 +43,9 @@ def run(self):
...
             except Exception:
-                stats.log_gauge("successful-db-update", 0)
-                stats.log_gauge("failed-db-update", 1)
+                stats.log_periodic_graph_update(False)
```
rra: `success=False`

grouper/api/handlers.py
```diff
@@ -77,23 +77,12 @@ def initialize(self, *args, **kwargs):
...
-        stats.log_rate("duration_ms_{}".format(self.__class__.__name__), duration_ms)
-
-        # log response status code
+        handler = self.__class__.__name__
         response_status = self.get_status()
-
-        stats.log_rate("response_status_{}".format(response_status), 1)
-        stats.log_rate("response_status_{}_{}".format(self.__class__.__name__, response_status), 1)
+        duration_ms = int((datetime.utcnow() - self._request_start_time).total_seconds() * 1000)
+        stats.log_request(handler, response_status, duration_ms)
```
rra: `self.plugins.log_request`

grouper/stats.py
```diff
@@ -1,24 +1,43 @@
...
-
 from grouper.plugin import get_plugin_proxy
 
 
-def set_defaults():
-    # type: () -> None
-    instance = task_id()
-    if instance is None:
-        instance = 0
+def log_background_run(success):
```
rra: Delete all of these (and I think you may be able to get rid of the entire file as a result). All new code should call the plugin proxy directly that it was configured with as part of initialization. Old code should call `get_plugin_proxy()` directly. There shouldn't be any need for this additional level of indirection.

tests/api/handlers_test.py
```diff
@@ -413,3 +416,38 @@ def test_public_keys(session, users, http_client, base_url):  # noqa: F811
     assert rows[0]["fingerprint"] == "e9:ae:c5:8f:39:9b:3a:9c:6a:b8:33:6b:cb:6f:ba:35"
     assert rows[0]["fingerprint_sha256"] == "MP9uWaujW96EWxbjDtPdPWheoMDu6BZ8FZj0+CBkVWU"
     assert rows[0]["comment"] == "some-comment"
+
+
+@pytest.mark.gen_test
+def test_request_logging(session, users, http_client, base_url):  # noqa: F811
+    """Test that the api request handlers properly log stats"""
+    # It does not appear possible to add a mock plugin to the global plugins here and expect it to
+    # be called by the request handler: the server side will create its own plugin proxy
```
rra: It's created in the `api_app` fixture. This code is ugly because it should be converted to an itest, but for now in legacy code you can get at the plugin proxy by injecting the `app` fixture and then adding plugins to `app.plugins`. That will let you get rid of the mocking, which will make this more straightforward.

tests/fe/handlers_test.py
```diff
@@ -865,3 +866,38 @@ def test_group_request_cancelled(session, users, groups, http_client, base_url):
 
     request = Request.get(session, requester_id=user.id, requesting_id=group.id)
     assert request.status == "cancelled"
+
+
+@pytest.mark.gen_test
+def test_request_logging(session, users, http_client, base_url):  # noqa: F811
+    """Test that the fe request handlers properly log stats"""
+    # It does not appear possible to add a mock plugin to the global plugins here and expect it to
+    # be called by the request handler: the server side will create its own plugin proxy
```
rra: Same here, inject the `app` fixture and reference `app.plugins`.

grouper/fe/util.py
```diff
@@ -208,18 +205,10 @@ def on_finish(self):
...
-        stats.log_rate("duration_ms_{}".format(self.__class__.__name__), duration_ms)
-
-        # log response status code
+        handler = self.__class__.__name__
+        duration_ms = int((datetime.utcnow() - self._request_start_time).total_seconds() * 1000)
         response_status = self.get_status()
-
-        stats.log_rate("response_status_{}".format(response_status), 1)
-        stats.log_rate("response_status_{}_{}".format(self.__class__.__name__, response_status), 1)
+        stats.log_request(handler, response_status, duration_ms)
```
rra: `self.plugins.log_request`

tests/fe/handlers_test.py
rra: Oh, actually, looking at this again, that won't work because `plugins` is initialized in `initialize()` and is per-handler rather than per-app. You're going to have to instead call `get_plugin_proxy()` to get the plugin proxy and then inject the new plugin (but that should work).

tests/api/handlers_test.py
rra: Ugh, this is more complicated than I thought now that I've looked at it more closely, since we do inject a proper separate `PluginProxy` to the API server right now. I think you need to also modify `tests/fixtures.py` in `api_app` to call `set_global_plugin_proxy` for right now until we come up with a better way.

cauthu:
> Mostly okay but there's an unnecessary extra level of indirection that makes things more complicated and that adds more uses of the deprecated `get_plugin_proxy`. See detailed comments.

Ah thanks! I hadn't noticed the plugins was passed into the app.

cauthu:
Thanks!

PR #585, opened 2019-10-07
log exception from database thread properly

herb:


rra:
Ugh, sorry about the late night exception. Somehow my grep missed that one.

PR #587, opened 2019-10-10
Clean up grouper.util typing

rra:
Remove an unnecessary covariant flag from a type variable and add
type annotations to the rest of the functions.

### DIFF
--- a/grouper/util.py
+++ b/grouper/util.py
@@ -6,19 +6,22 @@ import threading
 from typing import TYPE_CHECKING, TypeVar
 
 if TYPE_CHECKING:
+    from argparse import Namespace
     from grouper.settings import Settings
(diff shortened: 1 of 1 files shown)


grouper/util.py
```diff
@@ -6,19 +6,22 @@
...
 _TRUTHY = {"true", "yes", "1", ""}
 
 
 def qp_to_bool(arg):
+    # type: (str) -> bool
     return arg.lower() in _TRUTHY
 
 
 def get_loglevel(args, base=None):
+    # type: (Namespace, int) -> int
```
wrodriguez-dbx: I believe `base` should be `Optional[int]`

rra: Good catch, indeed.

PR #588, opened 2019-10-10
Clean up group itests

rra:
Add typing to the Selenium page wrappers, convert all group tests
to the new test format, and separate tests of the group list page
from tests of the page for a specific group (and move tests of
joining a group out of the latter to the existing file for those
tests).


itests/fe/group_join_test.py
```diff
@@ -11,6 +11,31 @@
     from tests.setup import SetupTest
 
 
+def test_request_to_join_group(tmpdir, setup, browser):
```
wrodriguez-dbx: How about instead moving the tests from the this file all into `group_view_test`? Having e.g. `test_request_to_join_group` and `test_remove_member` be in different files feels not great

rra: Hm, it feels correct to me. This is all of the tests that exercise the group join form, which `test_remove_member` does not do (it only exercises the group view page).

PR #589, opened 2019-10-29
Don't allow multiple join requests to the same group

rra:
If the user already has a pending request to join a group, don't
allow them to request membership again for their user (but still
allow them to request membership for groups they own).  Change the
first prompt, for the member, to be empty if the user themselves
already has a pending request but can still request membership on
behalf of some other group.  Add a warning explaining that this is
the case.

### DIFF
--- a/grouper/fe/handlers/group_join.py
+++ b/grouper/fe/handlers/group_join.py
@@ -14,7 +14,7 @@ from grouper.models.user import User
 from grouper.user_group import get_groups_by_user
 
 if TYPE_CHECKING:
-    from typing import Any, Mapping, List, Optional, Tuple, Union
+    from typing import Any, Mapping, List, Optional, Set, Tuple, Union
 
 
 class GroupJoin(GrouperHandler):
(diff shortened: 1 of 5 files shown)


grouper/fe/handlers/group_join.py
```diff
@@ -173,15 +186,13 @@ def _get_member(self, member_choice):
 
         return self.session.query(resource).filter_by(name=member_name, enabled=True).one()
 
-    def _get_choices(self, group):
-        # type: (Group) -> List[Tuple[str, ...]]
-        # This returns List[Tuple[str, str]], but mypy is confused by the * 2 syntax.
+    def _get_choices(self, group, members, user_is_member):
```
wrodriguez-dbx: I think the logic would be a bit clearer here if instead of `members` we passed in only the group members, as right now it's not super obvious until you read through the code why `user_is_member` is separate. Would also make the type and usage of `members` much simpler (could be as simple as `Set[str]` with just a single line of preprocessing)

grouper/fe/templates/group-join.html
```diff
@@ -21,6 +21,13 @@
                 <h3 class="panel-title">Join Group</h3>
            </div>
             <div class="panel-body">
+                {% if user_is_member %}
+                <div class="alert alert-warning">
+                    You are either already a member of this group or have a pending membership
+                    request. You can also request membership on behalf of a group, but be sure this
+                    is what you intended to do.
```
wrodriguez-dbx: Maybe even be explicit about requesting membership on behalf of a group _usually_ being incorrect?

PR #593, opened 2019-11-05
Convert service account permission grant to a usecase

rra:
Since this now uses the graph properly, it also allows delegating
permissions that are inherited by the group from a parent, rather
than forcing the permission to be granted directly to the group.

I'm doing something a little unusual here in the handler by stashing
some information into the handler object in the post() method that's
then used by the callback methods.  Ideally, everything in the
callback comes from the use case, but in this case the group owner
and the list of permissable permissions to grant are quite strange
to return as part of every error message from the usecase, and (in
the case of the list of grantable permissions) potentially expensive
to recalculate.  This seemed less awkward than forcing the usecase
to calculate and pass them around.

### DIFF
--- a/grouper/repositories/permission_grant.py
+++ b/grouper/repositories/permission_grant.py
@@ -11,6 +11,7 @@ from grouper.entities.permission_grant import (
     ServiceAccountPermissionGrant,
     UniqueGrantsOfPermission,
 )
+from grouper.entities.service_account import ServiceAccountNotFoundException
 from grouper.models.base.constants import OBJ_TYPES
 from grouper.models.group import Group
 from grouper.models.group_edge import GroupEdge
@@ -47,13 +48,29 @@ class GraphPermissionGrantRepository(PermissionGrantRepository):
         # type: (str, str, str) -> None
         self.repository.grant_permission_to_group(permission, argument, group)
 
+    def grant_permission_to_service_account(self, permission, argument, service):
+        # type: (str, str, str) -> None
+        self.repository.grant_permission_to_service_account(permission, argument, service)
+
     def group_grants_for_permission(self, name, include_disabled_groups=False):
         # type: (str, bool) -> List[GroupPermissionGrant]
         return self.repository.group_grants_for_permission(name)
 
-    def service_account_grants_for_permission(self, name):
-        # type: (str) -> List[ServiceAccountPermissionGrant]
-        return self.repository.service_account_grants_for_permission(name)
+    def permission_grants_for_group(self, name):
+        # type: (str) -> List[GroupPermissionGrant]
+        group_details = self.graph.get_group_details(name)
+        permissions = []
+        for permission_data in group_details["permissions"]:
+            permission = GroupPermissionGrant(
+                group=name,
+                permission=permission_data["permission"],
+                argument=permission_data["argument"],
+                granted_on=datetime.utcfromtimestamp(permission_data["granted_on"]),
+                is_alias=permission_data["alias"],
+                grant_id=None,
+            )
+            permissions.append(permission)
+        return permissions
 
     def permission_grants_for_service_account(self, name):
         # type: (str) -> List[ServiceAccountPermissionGrant]
@@ -100,6 +117,10 @@ class GraphPermissionGrantRepository(PermissionGrantRepository):
         # type: (str) -> List[ServiceAccountPermissionGrant]
         return self.repository.revoke_all_service_account_grants(permission)
 
+    def service_account_grants_for_permission(self, name):
+        # type: (str) -> List[ServiceAccountPermissionGrant]
+        return self.repository.service_account_grants_for_permission(name)
+
     def service_account_has_permission(self, service, permission):
         # type: (str, str) -> bool
         for grant in self.permission_grants_for_service_account(service):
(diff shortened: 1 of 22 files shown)


grouper/repositories/permission_grant.py
```diff
@@ -172,35 +207,66 @@ def group_grants_for_permission(self, name, include_disabled_groups=False):
...
+        graph behavior.  Use with caution until that is fixed.
+        """
+        now = datetime.utcnow()
+        group = Group.get(self.session, name=name)
+        if not group or not group.enabled:
             return []
-        grants = (
+        group_ids = [group.id]
+
+        # Now, get the parent groups of that group and so forth until we run out of levels of the
```
wrodriguez-dbx: While it is awkward in the repository world, could we pull this logic out into its own helper? Having such a large and perfectly identical chunk of code be duplicated kinda hurts my soul :P

grouper/repositories/service_account.py
```diff
@@ -84,6 +84,27 @@ def mark_disabled_user_as_service_account(self, name, description="", mdbset="")
...
+        return owner
+
     def service_account_exists(self, name):
         # type: (str) -> bool
         return SQLServiceAccount.get(self.session, name=name) is not None
+
+    def service_account_is_enabled(self, name):
+        # type: (str) -> bool
+        user = SQLUser.get(self.session, name=name)
+        return bool(user and user.is_service_account and user.enabled)
```
wrodriguez-dbx: Huh, out of curiosity why is the `bool()` necessary? Quick poking and Python3 seems to understand that `obj and bool and ...` is `bool`

grouper/repositories/permission_grant.py
rra: Yeah, I realized that after I went home. Now fixed in a subsequent diff in this same PR.

grouper/usecases/grant_permission_to_service_account.py
```diff
@@ -0,0 +1,155 @@
...
+
+        valid, error = self.permission_service.is_valid_permission_argument(permission, argument)
+        if not valid:
+            assert error
+            self.ui.grant_permission_to_service_account_failed_invalid_argument(
+                permission, argument, service, error
+            )
+            return
+
+        # If the actor is a permission admin, they can grant any permission to the service account.
```
wrodriguez-dbx: Oof, yea, validation checks are something I've had a lot of trouble finding a good place for in hexagonal. Having this duplicated login with `can_grant_permissions_for_service_account` just so we can call the appropriate callbacks feels yucky, but I don't see a better structure given the conventions we're following :/

itests/fe/service_accounts_test.py
```diff
@@ -126,3 +131,99 @@ def test_permission_grant_revoke(tmpdir, setup, browser):
...
+        assert len(permission_rows) == 1
+        permission = permission_rows[0]
+        assert permission.permission == "some-permission"
+        assert permission.argument == "*"
+
+        # The button doesn't show for someone who can't manage the service account.
+        with pytest.raises(NoSuchElementException):
+            permission.click_revoke_button()
+
+    # Add the user to the group so that the revoke button will show up, and then revoke it before
```
wrodriguez-dbx: Very unimportant: I'd argue that testing this by hitting the URL directly is both a little easier and very marginally more correct, as in theory the current method could fail on some weird auto-update behavior with the endpoint remaining unauthed. Practically though doesn't matter.

grouper/service_account.py
```diff
@@ -58,51 +50,6 @@ def _check_machine_set(service_account, machine_set):
         raise BadMachineSet(str(e))
 
 
-def can_create_service_account(session, actor, group):
-    # type: (Session, User, Group) -> bool
-    return actor.is_member(group.my_members())
-
-
-def create_service_account(session, actor, name, description, machine_set, owner):
```
wrodriguez-dbx: Yay :)

tests/repositories/permission_grant_test.py
```diff
@@ -38,3 +38,32 @@ def test_permission_grants_for_user(setup):
...
+        setup.grant_permission_to_group("other-perm", "arg", "grandparent")
+        setup.grant_permission_to_group("other-perm", "*", "grandparent")
+        setup.grant_permission_to_group("perm", "two", "two-group")
+        setup.grant_permission_to_group("another-perm", "foo", "np-group")
+
+    # Check the returned permissions.
+    permission_grants = permission_grant_repository.permission_grants_for_group("one-group")
+    assert sorted([(p.permission, p.argument) for p in permission_grants]) == [
+        ("other-perm", "*"),
+        ("other-perm", "arg"),
```
wrodriguez-dbx: Ah, so guessing that the service account permission grant UI would display duplicates like this? Not that it particularly matters

wrodriguez-dbx:
Ye, though I believe that right now the validation doesn't check the list at all (rather than being a redundant check). Still ultimately agree that the conceptual simplicity of always having a valid form trumps any other ugliness, regardless of whether we end up getting the WTForms niceness.

grouper/usecases/grant_permission_to_service_account.py
rra: I tried to find some way to use a helper function, but the deeper problem is that the logic is very slightly different because it has to take the argument into account in the actual change but not in the pre-check (since the argument is not known in the pre-check). But that's only relevant if the user is not a permission admin, since permission admins are allowed to grant permissions not held by the group.

I'm sure there's some better structure here that allows sharing more of the code, but I couldn't come up with it and decided to live with this mess. (I think it's likely that we're going to change this logic in the future anyway.)

itests/fe/service_accounts_test.py
rra: I completely agree and had started down that path and then discovered that Python Selenium doesn't allow you to do a test-initiated POST, so there's no way to hit the URL directly (since it's POST-only, since it mutates state) without having something to click on. I'll add a comment to make that clearer to the next person to look at this code.

tests/repositories/permission_grant_test.py
rra: Yeah, it currently does. There's an open Jira ticket to fix that; not sure if I'll get to that this week.

PR #596, opened 2019-11-05
Improve username validation

rra:
Allow dashes in domain names, and check that the RHS of the username
is a superficially valid domain (has at least two parts, final TLD is
at least two characters long, no empty segments of domain).

### DIFF
--- a/tests/constants_test.py
+++ b/tests/constants_test.py
@@ -1,18 +1,25 @@
 import re
 
-from grouper.constants import USERNAME_VALIDATION
+from grouper.constants import SERVICE_ACCOUNT_VALIDATION, USERNAME_VALIDATION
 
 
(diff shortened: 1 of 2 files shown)


tests/constants_test.py
```diff
@@ -0,0 +1,27 @@
...
+    ]
+    for user in good_users:
+        assert re.match(regex, user)
+
+    bad_users = [
+        "nodomain",  # no domain part
+        "bad@domain",  # invalid domain part
+        "foo@u.s",  # single character top-level domain
+        "johnny.appleseed@a.co",  # period in username
+        "jackbrown+new@a.co",  # gmail style aliases
```
wrodriguez-dbx: Out of curiosity, is there a reason we don't allow these?

PR #599, opened 2019-11-06
Remove all mypy Text types

rra:
Text is no longer necessary since Grouper is Python 3 only and we
can just use str.  Also convert all the files I had to touch to the
new-style type annotation syntax.

### DIFF
--- a/grouper/email_util.py
+++ b/grouper/email_util.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 import smtplib
 from datetime import datetime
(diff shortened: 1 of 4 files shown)


rra:
The main difference is that you want `from __future__ import annotations` or Python will try to resolve the types at run time, not only during type checking. With that in place, it should be identical; it's just the preferred Python 3 syntax, and it's present properly in the Python parse tree which should make some tooling easier.

The benefits are really just cleanliness and looking more like a Python 3 code base rather than anything concrete.

PR #600, opened 2019-11-07
Catch error from adding groups as non-members

rra:
Attempting to add a group as a role other than member previously
threw an uncaught exception.  Add a test, catch the exception
properly, and show it as an alert on the page.

### DIFF
--- a/itests/fe/group_join_test.py
+++ b/itests/fe/group_join_test.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import TYPE_CHECKING
 
 import pytest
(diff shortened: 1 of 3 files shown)


itests/fe/group_join_test.py
```diff
@@ -113,3 +111,38 @@ def test_require_clickthru(tmpdir, setup, browser):
...
+    with frontend_server(tmpdir, "gary@a.co") as frontend_url:
+        browser.get(url(frontend_url, "/groups/parent-group/join"))
+        join_page = GroupJoinPage(browser)
+
+        join_page.set_member("some-group")
+        join_page.set_reason("Testing")
+        join_page.submit()
+
+        group_page = GroupViewPage(browser)
+        print(group_page.root.page_source)
```
wrodriguez-dbx: Debugging print?

rra: Good catch, thanks!

PR #601, opened 2019-11-07
Remove old group permission request page

rra:
The new generic permission request page can handle permission
requests restricted to a specific group and implements another
version of the same logic as the old group permission request page.
Remove the latter so that there's only one version of this page.


rra:
It's discoverable from the page of any permission. If you go to the page for a permission, there's a Request this permission button at the top. (This was Branden's Hack Week project last year.)

PR #602, opened 2019-11-07
Don't allow editing the role of a member group

rra:
Limit the choices of role for a member that is a group to only
member, and disable the field.  This avoids uncaught exceptions when
someone tries to change the role to something else.

Add mypy typing to the relevant frontend handler and fix a few other
issues.

Use the standard Selenium exception for NoSuchElementException so
there aren't two exceptions with the same name floating around.

### DIFF
--- a/grouper/fe/handlers/group_edit_member.py
+++ b/grouper/fe/handlers/group_edit_member.py
@@ -1,4 +1,7 @@
+from __future__ import annotations
+
 from datetime import datetime
+from typing import TYPE_CHECKING
 
 from grouper.audit import assert_can_join, UserNotAuditor
 from grouper.fe.forms import GroupEditMemberForm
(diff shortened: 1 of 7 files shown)


grouper/fe/handlers/group_edit_member.py
```diff
@@ -11,22 +14,30 @@
...
 from grouper.user import user_role
 
+if TYPE_CHECKING:
+    from typing import Any, Optional
+
 
 class GroupEditMember(GrouperHandler):
-    def get(self, group_id=None, name=None, name2=None, member_type=None):
-        group = Group.get(self.session, group_id, name)
+    def get(self, *args: Any, **kwargs: Any) -> None:
```
wrodriguez-dbx: Out of curiosity, why use `args`/`kwargs` here?

rra: Sad Tornado nonsense. The Tornado typeshed declares the type of these methods as this generic type, so you either have to add type ignore comments or you have to live with it and pull things out of `kwargs` with explicit typing. I really hate type ignore comments, so decided to live with this. The same is done in all the other Tornado handlers that have been modified since we started doing mypy.

PR #605, opened 2019-11-07
Remove unused parameter to notification function

rra:
notify_nonauditor_promoted didn't do anything with the set of
group names that explained why the user was promoted.  Remove the
unused parameter.


rra:
Never mind, it's used in the test in a way that I missed since it's buried in a mock.

PR #609, opened 2019-11-07
Modernize all API itests

rra:
Convert all remaining API itests to the modern test framework and
retire the old fixtures.  Required adding support for setting an
email address for a group during group creation.

### DIFF
--- a/itests/api/service_accounts_test.py
+++ b/itests/api/service_accounts_test.py
@@ -1,69 +1,75 @@
+from __future__ import annotations
+
 from typing import TYPE_CHECKING
 
 from groupy.client import Groupy
 
-from itests.fixtures import api_client, async_api_server  # noqa: F401
 from itests.setup import api_server
-from tests.fixtures import (  # noqa: F401
-    graph,
-    groups,
-    permissions,
-    service_accounts,
-    session,
-    standard_graph,
-    users,
-)
 
 if TYPE_CHECKING:
     from py.local import LocalPath
     from tests.setup import SetupTest
 
 
-def test_get_service_accounts(api_client, users, service_accounts):  # noqa: F811
-    role_users = [username for username, u in users.items() if u.role_user]
-    assert len(role_users) > 0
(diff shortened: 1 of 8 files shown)


itests/api/service_accounts_test.py
```diff
@@ -1,69 +1,74 @@
...
     from tests.setup import SetupTest
 
 
-def test_get_service_accounts(api_client, users, service_accounts):  # noqa: F811
-    role_users = [username for username, u in users.items() if u.role_user]
-    assert len(role_users) > 0
+def test_get_service_accounts(tmpdir: LocalPath, setup: SetupTest) -> None:
+    with setup.transaction():
+        setup.create_role_user("role@a.co")
+        setup.create_service_account("service@a.co", "team-sre")
```
wrodriguez-dbx: For technically full parity with what was tested before, might as well throw in a `setup.create_user` to be sure non-role users aren't returned as well?

itests/api/service_accounts_test.py
```diff
@@ -1,69 +1,74 @@
...
-    assert role_user.permissions == []
-    assert role_user.metadata == {}
+def test_get_role_user(tmpdir: LocalPath, setup: SetupTest) -> None:
+    with setup.transaction():
+        setup.create_role_user("role@a.co")
+
+    with api_server(tmpdir) as api_url:
+        api_client = Groupy(api_url)
+        role_user = api_client.service_accounts.get("role@a.co")
+        assert sorted(role_user.groups) == ["role@a.co"]
```
wrodriguez-dbx: Huh, why's this different? Was it just a bug in the old fixtures? That `tests/fixtures.py` code does seem suspect, though I haven't dug in -- if it is messed up could we also fix it since it isn't removed here?

rra: It's a bug in `tests/fixtures.py` that didn't create role users properly. I fixed that bug, but the fix is in some later PR that was stacked behind this one.

itests/api/service_accounts_test.py
rra: Ah, yes, good idea. Will do that.

PR #611, opened 2019-11-07
Use frozen dataclasses instead of NamedTuples

rra:
The syntax is much nicer, the code is shorter, and dataclasses
support optional attributes and (not yet used) inheritance.

Also modernize the code for all grouper.entities.* classes.

### DIFF
--- a/grouper/entities/permission.py
+++ b/grouper/entities/permission.py
@@ -1,27 +1,31 @@
-from datetime import datetime
-from typing import NamedTuple
+from __future__ import annotations
 
-Permission = NamedTuple(
-    "Permission",
-    [
-        ("name", str),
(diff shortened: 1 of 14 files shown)


grouper/entities/permission.py
```diff
@@ -1,27 +1,29 @@
...
-    [
-        ("name", str),
-        ("description", str),
-        ("created_on", datetime),
-        ("audited", bool),
-        ("enabled", bool),
-    ],
-)
-
-# The actions a user can perform on a permission.
```
wrodriguez-dbx: Nit: Could we keep this comment? :)

rra: Yup, good idea. Done.

PR #613, opened 2019-11-08
Update dependencies

rra:
Update versions to match what's used internally at Dropbox, and
update development dependencies that don't have to be kept in sync
to their latest versions.  Apply changes made by the new black.

Remove dependencies that are no longer needed since Python 2 isn't
supported.  Remove the development dependency on coverage, which
was there for convenience but isn't automatically used.

Update mypy and fix new issues that it found.  Convert the affected
files to new-style type annotations.


grouper/ctl/group.py
```diff
@@ -11,28 +13,28 @@
...
 
 @ensure_valid_groupname
-def group_command(args, settings, session_factory):
-    # type: (Namespace, CtlSettings, SessionFactory) -> None
+def group_command(args: Namespace, settings: CtlSettings, session_factory: SessionFactory) -> None:
     session = session_factory.create_session()
     group = session.query(Group).filter_by(groupname=args.groupname).scalar()
     if not group:
-        logging.error("No such group %s".format(args.groupname))
+        logging.error("No such group %s", args.groupname)
```
wrodriguez-dbx: Nit: Could be f-string because why not

wrodriguez-dbx: Ah, I see now the note on #616

PR #618, opened 2019-11-09
Add test for group join request with Unicode reason

rra:
This previously failed when generating the email message with an
encoding error.  It now works, probably because Grouper is pure
Python 3.

### DIFF
--- a/itests/fe/group_join_test.py
+++ b/itests/fe/group_join_test.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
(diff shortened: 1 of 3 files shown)


itests/fe/group_join_test.py
```diff
@@ -145,3 +147,17 @@ def test_group_join_group_as_owner(tmpdir: LocalPath, setup: SetupTest, browser:
...
+
+def test_request_join_unicode(tmpdir: LocalPath, setup: SetupTest, browser: Chrome) -> None:
+    with setup.transaction():
+        setup.add_user_to_group("gary@a.co", "some-group", "owner")
+
+    with frontend_server(tmpdir, "cbguder@a.co") as frontend_url:
+        browser.get(url(frontend_url, "/groups/some-group/join"))
+        page = GroupJoinPage(browser)
+
+        page.set_reason("защото причини")
```
wrodriguez-dbx: изглежда законно

PR #620, opened 2019-11-12
Modernize all frontend itests

rra:
Convert all remaining frontend itests to the modern test framework
and retire the old fixtures.


itests/fe/permission_requests_test.py
```diff
@@ -1,187 +1,178 @@
...
+    from selenium.webdriver import Chrome
+    from tests.setup import SetupTest
+
+
+def create_permission_requests(setup: SetupTest) -> None:
+    """Create a permission requesting scenario.
+
+    Set up a permission requesting scenario in which REQUESTING_USER has both inbound and outbound
+    requests that they should be able to see on the requests page. Do this because the itest
+    infrastructure visits all pages as cbguder.
```
wrodriguez-dbx: Nits: `REQUESTING_USER` not a thing anymore, and no longer true that itests can only be as `cbguder`

itests/fe/permission_requests_test.py
```diff
@@ -1,187 +1,178 @@
...
+    with frontend_server(tmpdir, "cbguder@a.co") as frontend_url:
+        browser.get(url(frontend_url, "/permissions/requests?status=pending"))
+        page = PermissionRequestsPage(browser)
+
+        # Check the request rows.
+        requested = [row.requested for row in page.request_rows]
+        expected = ["perm.hasgranter, a", "perm.nogranter, a", f"{PERMISSION_ADMIN}, a"]
+        assert sorted(requested) == sorted(expected)
+
+        # Check the status change rows.
```
wrodriguez-dbx: Likely not worth the effort, but it should be feasible to have `PermissionRequestsPage` be smarter about status change rows to avoid this messiness, right? Specifically having `StatusChangeRow`s included in each appropriate `RequestViewRow`

rra: Yeah, it's a larger project than I want to tackle right now, but I made a note for myself for later.

itests/fe/permission_requests_test.py
rra: Thanks, fixed.

PR #623, opened 2019-11-14
Add typing to grouper.permissions

rra:
Convert a namedtuple to a dataclass object and get rid of some
dead code.  Some compromises had to be made with the typing because
this code does some odd things, like using a singleton object as a
key in an otherwise uniform dict.

### DIFF
--- a/grouper/permissions.py
+++ b/grouper/permissions.py
@@ -551,14 +551,25 @@ def can_approve_request(
     owner: User,
     group_ids: Optional[Set[int]] = None,
     owners_by_arg_by_perm: Optional[Dict[object, Dict[str, List[Group]]]] = None,
-) -> Set[int]:
+) -> bool:
+    """Determine whether the given owner can approve a permission request.
+
+    Args:
+        session: Database session
+        request: Pending permission request
+        owner: User who may or may not be able to approve the request
+        group_ids: If given, the IDs of the groups of which the user is a member (solely so that we
+            can avoid another database query if this information is already available)
+        owners_by_arg_by_perm: List of permission granters by permission and argument (solely so
+            that we can avoid another database query if this information is already available)
+    """
     owner_arg_list = get_owner_arg_list(
         session, request.permission, request.argument, owners_by_arg_by_perm
     )
     if group_ids is None:
         group_ids = {g.id for g, _ in get_groups_by_user(session, owner)}
 
-    return group_ids.intersection([o.id for o, arg in owner_arg_list])
+    return bool(group_ids.intersection([o.id for o, arg in owner_arg_list]))
 
 
 def get_requests(
--- a/grouper/permissions.py
+++ b/grouper/permissions.py
@@ -1,7 +1,10 @@
+from __future__ import annotations
+
 import re
-from collections import defaultdict, namedtuple
+from collections import defaultdict
+from dataclasses import dataclass
 from datetime import datetime
-from typing import TYPE_CHECKING
+from typing import cast, TYPE_CHECKING
 
 from sqlalchemy import asc
 from sqlalchemy.exc import IntegrityError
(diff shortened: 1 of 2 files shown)


grouper/permissions.py
```diff
@@ -146,23 +123,23 @@ def grant_permission(session, group_id, permission_id, argument=""):
...
+        permission: A Permission object being granted
+        argument: Must match constants.ARGUMENT_VALIDATION
 
     Throws:
         AssertError if argument does not match ARGUMENT_VALIDATION regex
     """
-    assert re.match(
-        ARGUMENT_VALIDATION + r"$", argument
-    ), "Permission argument does not match regex."
+    assert re.match(ARGUMENT_VALIDATION + r"$", argument), "Invalid permission argument"
```
wrodriguez-dbx: Should have a `r"^"` at the start as well to be totally comprehensive, right?

grouper/permissions.py
```diff
@@ -252,20 +229,20 @@ def get_groups_by_permission(session, permission):
     )
 
 
-def filter_grantable_permissions(session, grants, all_permissions=None):
-    """For a given set of PERMISSION_GRANT permissions, return all enabled
-    permissions that are grantable.
+def filter_grantable_permissions(
+    session: Session, grants: List[Any], all_permissions: Optional[Dict[str, Permission]] = None
```
wrodriguez-dbx: Ah, the `Any` because they're raw sqlalchemy objects?

grouper/permissions.py
```diff
@@ -285,34 +262,37 @@ def filter_grantable_permissions(session, grants, all_permissions=None):
...
 
 
-def get_owners_by_grantable_permission(session, separate_global=False):
-    """
-    Returns all known permission arguments with owners. This consolidates
-    permission grants supported by grouper itself as well as any grants
-    governed by plugins.
+def get_owners_by_grantable_permission(
+    session: Session, separate_global: bool = False
+) -> Dict[object, Dict[str, List[Group]]]:
```
wrodriguez-dbx: As much as it hurts, I wonder if it wouldn't be less ugly to set `GLOBAL_OWNERS` to some string that isn't a valid permission name?

grouper/permissions.py
```diff
@@ -382,28 +362,34 @@ def _reduce_args(perm_name, args):
...
 
-def get_owner_arg_list(session, permission, argument, owners_by_arg_by_perm=None):
-    """Return the grouper group(s) responsible for approving a request for the
-    given permission + argument along with the actual argument they were
-    granted.
+def get_owner_arg_list(
+    session: Session,
+    permission: Permission,
+    argument: str,
+    owners_by_arg_by_perm: Optional[Dict[object, Dict[str, List[Group]]]] = None,
```
wrodriguez-dbx: oof

grouper/permissions.py
```diff
@@ -550,25 +536,22 @@ def create_request(session, user, group, permission, argument, reason):
...
 
 
-def can_approve_request(session, request, owner, group_ids=None, owners_by_arg_by_perm=None):
+def can_approve_request(
+    session: Session,
+    request: PermissionRequest,
+    owner: User,
+    group_ids: Optional[Set[int]] = None,
+    owners_by_arg_by_perm: Optional[Dict[object, Dict[str, List[Group]]]] = None,
+) -> Set[int]:
```
wrodriguez-dbx: Nit: Mind throwing a quick docstring on this function? Don't think it's super clear from name + args what it actually does. (Admittedly it's fairly straightforward to look at the code, but reading English is faster :) )

Feel free to decline of course, particularly because the behavior here might actually be harder to explain in English than with the code

grouper/permissions.py
rra: `re.match` is implicitly anchored at the start of the string (as opposed to `re.search`, which is not). I find this confusing in Python as well.

grouper/permissions.py
rra: Yeah, and not even full model objects, but tuples. And SQLAlchemy itself doesn't appear to be typed, so it's not clear what to use for a type other than `Any`. This will go away once we have a repository layer wrapping all of the SQLAlchemy code.

grouper/permissions.py
rra: It probably would, but I didn't want to make a behavior change like that, since it stands a high chance of introducing some obscure bug. I decided to leave it alone until we can fix it properly (which would involve returning permissions for global owners in a separate parallel structure and bundling them both together with a dataclass object).

grouper/permissions.py
rra: Good idea. It also uncovered that it only needs to return a bool.

PR #624, opened 2019-11-14
Cancel audit warning messages after group disable

rra:
If an audited group is disabled, cancel all the audit warning
messages in addition to marking the audit as complete.

When disabling a role user, both mark the audit as complete and
cancel all the audit warning messages.

Add a way to set an environment variable to disable headless mode
when running Selenium tests, since it helps with debugging new
tests.


grouper/fe/templates/service.html
```diff
@@ -26,7 +26,8 @@
     <!-- Enable/Disable button. -->
```
wrodriguez-dbx: Nit: could we rename this file to `role-user.html` while we're here?

itests/fe/audits_test.py
```diff
@@ -16,6 +18,65 @@
...
+        audit_modal = page.get_audit_modal()
+        audit_modal.click_close_button()
+        page.wait_until_audit_modal_clears()
+        page.click_disable_button()
+        modal = page.get_disable_modal()
+        modal.confirm()
+
+        assert page.subheading == "some-group (disabled)"
+
+    # And now all of the email messages should be marked sent except the non-async one.
```
wrodriguez-dbx: Ah, non-async meaning immediate, which is still an `AsyncNotification`. Was confused for a sec :)

grouper/fe/templates/service.html
rra: Good idea. Done.

itests/fe/audits_test.py
rra: Ah, yes, clarified the comment.

PR #628, opened 2019-11-16
Allow downgrading your own role

rra:
Permit an owner to downgrade their role in a group, but not upgrade
(so np-owner cannot become owner).  Ensure that this doesn't allow
someone to remove the last owner by editing their role.

This redoes the exception handling in Group.add_member() slightly
since otherwise it was spewing errors into the logs when someone
tried to downgrade the role as the last owner.  This isn't an
exceptional case, just a known error, so don't use the noisy
flush_transaction decorator (which is also a little silly here since
AuditLog.log() commits the transaction anyway), and instead manually
wrap the call and roll back if needed.

### DIFF
--- a/grouper/fe/handlers/group_edit_member.py
+++ b/grouper/fe/handlers/group_edit_member.py
@@ -29,9 +29,6 @@ class GroupEditMember(GrouperHandler):
         if not group:
             return self.notfound()
 
-        if self.current_user.name == user:
-            return self.forbidden()
-
         members = group.my_members()
         my_role = user_role(self.current_user, members)
         if my_role not in ("manager", "owner", "np-owner"):
@@ -50,15 +47,7 @@ class GroupEditMember(GrouperHandler):
         if not edge:
             return self.notfound()
 
-        form = GroupEditMemberForm(self.request.arguments)
-        form.role.choices = [["member", "Member"]]
-        if my_role in ("owner", "np-owner") and member_type != "group":
-            form.role.choices.append(["manager", "Manager"])
-            form.role.choices.append(["owner", "Owner"])
-            form.role.choices.append(["np-owner", "No-Permissions Owner"])
-        else:
-            form.role.render_kw = {"readonly": "readonly"}
-
+        form = self._get_form(user, my_role, member_type)
         form.role.data = edge.role
         form.expiration.data = edge.expiration.strftime("%m/%d/%Y") if edge.expiration else None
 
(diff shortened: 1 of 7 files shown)


grouper/fe/handlers/group_edit_member.py
```diff
@@ -156,3 +134,22 @@ def post(self, *args: Any, **kwargs: Any) -> None:
...
+        cannot make themselves owners.  The role of a group cannot be changed from member.
+        """
+        form = GroupEditMemberForm(self.request.arguments)
+        form.role.choices = [["member", "Member"]]
+        if my_role in ("owner", "np-owner") and member_type != "group":
+            form.role.choices.append(["manager", "Manager"])
+            form.role.choices.append(["np-owner", "No-Permissions Owner"])
+            if my_role == "owner":
+                form.role.choices.append(["owner", "Owner"])
+        elif member_type == "group" or my_role == "manager":
```
wrodriguez-dbx: Manager -> member is a downgrade, right? Are we not allowing it just because managers shouldn't be able to make any role changes?

grouper/fe/handlers/group_edit_member.py
```diff
@@ -156,3 +134,22 @@ def post(self, *args: Any, **kwargs: Any) -> None:
...
+        """Get the form with possible role options filled in.
+
+        Owners can edit themselves to downgrade their membership, but not upgrade their
+        membership.  So owners can make themselves members, np-owners, or managers, but np-owners
+        cannot make themselves owners.  The role of a group cannot be changed from member.
+        """
+        form = GroupEditMemberForm(self.request.arguments)
+        form.role.choices = [["member", "Member"]]
+        if my_role in ("owner", "np-owner") and member_type != "group":
+            form.role.choices.append(["manager", "Manager"])
```
wrodriguez-dbx: Isn't np-owner -> manager technically not a downgrade, as it'd give them permissions? Unless our help text is wrong about `a manager is a member who is additionally allowed to process requests to join a group`

grouper/fe/handlers/group_edit_member.py
rra: Yeah, good call, this logic wasn't really correct. I think I've now cleaned it up.

grouper/fe/handlers/group_edit_member.py
rra: Yes. I had completely forgotten about that. Thank you!

PR #629, opened 2019-11-16
Flesh out typing for User and Permission

rra:
Requires fixing various places that were assuming that the return
value of Group.get was not None (mostly by adding asserts).  Found
one actual bug and a test setup problem with the legacy tests.

### DIFF
--- a/grouper/repositories/service_account.py
+++ b/grouper/repositories/service_account.py
@@ -44,7 +44,7 @@ class ServiceAccountRepository:
         # type: (str, str, str, str) -> None
         group = Group.get(self.session, name=owner)
         if not group:
-            raise GroupNotFoundException(group)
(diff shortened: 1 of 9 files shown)


grouper/repositories/service_account.py
```diff
@@ -44,7 +44,7 @@ def create_service_account(self, name, owner, machine_set, description):
         # type: (str, str, str, str) -> None
         group = Group.get(self.session, name=owner)
         if not group:
-            raise GroupNotFoundException(group)
+            raise GroupNotFoundException(name)
```
wrodriguez-dbx: `name` -> `owner` looks like

rra: Good catch. Thank you!

PR #630, opened 2019-11-22
Introduce a new method of template handling

rra:
Template handling in the frontend has caused problems because there
is no systematic validation that the correct variables are passed into
the template when rendering it, or that those variables have the
correct type.  Handling of alerts is also somewhat tedious.

Introduce a new wrapper dataclass around each template and use it in
all handlers that have been rewritten to use usecases.  This provides
type and argument validation and hides more of the machinery of
generating the template namespace.

Take advantage of this opportunity to add automatic handling of form
errors, turning them into alerts.

Move the definition of Alert to a separate file to reduce circular
dependencies between grouper.fe.templates and grouper.fe.util.  (There
is still a circular dependency for type checking since GrouperHandler
has a method that takes a BaseTemplate, and BaseTemplate has to get
data from the handler, but it's not a true source circular dependency.)

Stash the constructed form in the handler object in more cases, which
allows reuse in error callbacks without reconstructing the form.  As a
bonus, this means validate() has always been called, so it's safe to
append to field errors.

Remove a stray debugging print in the service account integration test.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -280,6 +280,27 @@ provides a pre-constructed use case factory to its handlers, created as
 part of initialization of the UI.  For tests, repository, service, and use
 case factories are provided as attributes on the `SetupTest` object.
 
+Templates
+---------
+
+There is a new mechanism for wrapping templates for the frontend (the `fe`
+directory) in `grouper.fe.templates`.  Each template has a corresponding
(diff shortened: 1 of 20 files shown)


rra:
See https://www.eyrie.org/~eagle/journal/2019-11/001.html for more information about the thought process here.

rra:
Yup, good idea. Done.

PR #631, opened 2019-12-11
Test that template dataclasses match template needs

rra:
For each template wrapped in a dataclass, test that the dataclass
defines all of the variables that the form needs.

This unfortunately doesn't expand macros, so we have to whitelist
all the macros.  It also doesn't handle inclusion of other files,
so in-line the forms into the templates (which is probably a good
idea anyway).

Rename the permission macro to permission_link, since permission is
used elsewhere as a variable.

Modernize mypy typing of grouper.templating (no modifications were
needed here in the end, but it seemed pointless to throw out the
work).

### DIFF
--- a/grouper/fe/templates/permissions.html
+++ b/grouper/fe/templates/permissions.html
@@ -1,5 +1,5 @@
 {% extends "base.html" %}
-{% from 'macros/ui.html' import permission, paginator, dropdown with context %}
+{% from 'macros/ui.html' import permission_link, paginator, dropdown with context %}
 
 {% block subtitle %}
     {% if audited_permissions %}
@@ -51,7 +51,7 @@
             <tbody>
             {% for perm in permissions %}
                 <tr class="permission-row">
-                    <td class="permission-name">{{ permission(perm) }}</td>
+                    <td class="permission-name">{{ permission_link(perm) }}</td>
                     <td class="permission-description">{{ perm.description|default("", True) }}</td>
                     <td class="permission-created-on">{{ perm.created_on | print_date}}</td>
                 </tr>
--- a/tests/fe/templates_test.py
+++ b/tests/fe/templates_test.py
@@ -1,0 +1,59 @@
+from __future__ import annotations
+
+import os
+from dataclasses import fields, is_dataclass
+from typing import TYPE_CHECKING
+
+from jinja2.meta import find_undeclared_variables
+
+import grouper.fe
+from grouper.fe.settings import FrontendSettings
+from grouper.fe.templates import BaseTemplate
+from grouper.fe.templating import FrontendTemplateEngine
+
+if TYPE_CHECKING:
+    from typing import Set
+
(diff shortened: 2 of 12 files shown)


tests/fe/templates_test.py
```diff
@@ -0,0 +1,59 @@
...
+from grouper.fe.settings import FrontendSettings
+from grouper.fe.templates import BaseTemplate
+from grouper.fe.templating import FrontendTemplateEngine
+
+if TYPE_CHECKING:
+    from typing import Set
+
+# These fields are automatically added to the template namespace for all render invocations in
+# BaseTemplate or BaseTemplateEngine and thus don't need to be provided by the template dataclass.
+DEFAULT_FIELDS = {"ROLES", "alerts", "is_active", "static_url", "update_qs", "xsrf_form"}
```
wrodriguez-dbx: I think there's a potential failure case that wouldn't be caught: if one of these is no longer set but a template requires it, the test will still pass. Maybe pull out into constants (lower the chance one of these is randomly no longer set) or pull out the default namespace construction in `BaseTemplate` and `BaseTemplateEngine` into explicitly testable functions?
Alternatively, that failure case might just be unlikely enough that we don't care :)

Separately, should `"TYPES"` also be in here? https://github.com/dropbox/merou/blob/cc70dd41850884e1fc112125dfc693c2458f4f73/grouper/templating.py#L49

tests/fe/templates_test.py
```diff
@@ -0,0 +1,59 @@
+from __future__ import annotations
+
+import os
+from dataclasses import fields, is_dataclass
+from typing import TYPE_CHECKING
+
+from jinja2.meta import find_undeclared_variables
+
+import grouper.fe
```
wrodriguez-dbx: Does this take care of importing all the template classes so `BaseTemplate.__subclasses__()` finds them? I never remember details of Python import behavior

grouper/fe/templates/permissions.html
```diff
@@ -1,5 +1,5 @@
 {% extends "base.html" %}
-{% from 'macros/ui.html' import permission, paginator, dropdown with context %}
+{% from 'macros/ui.html' import permission_link, paginator, dropdown with context %}
```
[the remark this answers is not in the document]
rra: It unfortunately needs `with context` for `update_qs`.

tests/fe/templates_test.py
rra: It relies on the fact that they're all defined in the same file. I'll add an explicit comment to that effect to `grouper.fe.templates`.

tests/fe/templates_test.py
rra: Will add `TYPES`. I thought about breaking this up into separate testable functions, but the problem I ran into is that `grouper.fe.templating` wants a `GrouperHandler`, which are really difficult to instantiate in a test suite. I think this will work for now; we've almost never changed those variables, and hopefully anyone who does will do a grep to see what else uses them.

PR #637, opened 2020-03-03
Fix the cleanup_audits script

FlakkenTime:
the sqlite used in dev does not enforce a foreign key constraint. We need to cleanup both tables.

### DIFF
--- a/oneoffs/cleanup_audits.py
+++ b/oneoffs/cleanup_audits.py
@@ -27,7 +27,7 @@ class AuditCleanup(BaseOneOff):
 
             # now pull the list of audit members that belong to this audit.
             for audit in audits:
-                print(
+                logging.info(
                     "Would delete Audit: ID={}, Complete={}, started_at={}, ends_at={}".format(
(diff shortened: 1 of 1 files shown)


oneoffs/cleanup_audits.py
```diff
@@ -1,19 +1,21 @@
...
+from grouper.models.audit_member import AuditMember
 from grouper.oneoff import BaseOneOff
 
 
 class AuditCleanup(BaseOneOff):
     def run(self, session, dry_run=True):
         # Pull the audits that are expiring on 2020-03-13 00:00:00. None of these have been
-        # completed. About half of the ones for the 31st have already been completed
+        # completed as of my checks. About half of the ones for the 31st have already
+        # been completed
```
wrodriguez-dbx: Nit: Preference to just say that the hardcoded 2020-03-13 is the date we had to delete when this happened, rather than baking in additional Dropbox-specific and right-now-specific info about completion progress.

PR #640, opened 2020-04-30
added dry_run arg to user_created()

ph0enix49:
This is to support dry_run args passed to user_created() in plugins


wrodriguez-dbx:
Closing as this isn't currently being worked on.

PR #642, opened 2020-05-21
Avoid infinite loop when checking auditor status

wrodriguez-dbx:
Not adding an explicit test case for this, as it triggers when there's a cyclic group membership and cyclic group membership aren't something we're generally robust/testing against. Felt bad to leave this known infinite loop case though.

### DIFF
--- a/grouper/audit.py
+++ b/grouper/audit.py
@@ -73,6 +73,7 @@ def assert_controllers_are_auditors(group: Group) -> None:
         cur_group = queue.pop()
         if cur_group in checked:
             continue
+        checked.add(cur_group)
         details = graph.get_group_details(cur_group)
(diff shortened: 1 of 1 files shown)


grouper/audit.py
```diff
@@ -73,6 +73,7 @@ def assert_controllers_are_auditors(group: Group) -> None:
         cur_group = queue.pop()
         if cur_group in checked:
             continue
+        checked.add(cur_group)
         details = graph.get_group_details(cur_group)
         for chk_user, info in details["users"].items():
             if chk_user in checked:
```
rohansharma, line 79: nit: we could also add this check before we append to the queue to avoid processing groups multiple times, but I guess efficiency isn't super critical here

PR #646, opened 2020-09-10
extend create_service_account to support initial user metadata

cugini-dbx:
Implemented at the usecase layer, but not exposed in the UI service account creation dialog.

### DIFF
--- a/grouper/repositories/service_account.py
+++ b/grouper/repositories/service_account.py
@@ -60,6 +60,7 @@ class ServiceAccountRepository:
         # Set initial user metadata fields if present.
         if initial_metadata is not None:
             for key, value in initial_metadata.items():
+                # TODO: move this to use the hexagonal architecture model.
                 set_user_metadata(self.session, user.id, key, value)
 
         # Create the linkage to the owner.
--- a/grouper/repositories/service_account.py
+++ b/grouper/repositories/service_account.py
@@ -7,9 +7,11 @@ from grouper.models.group import Group
 from grouper.models.group_service_accounts import GroupServiceAccount
 from grouper.models.service_account import ServiceAccount as SQLServiceAccount
 from grouper.models.user import User as SQLUser
+from grouper.user_metadata import set_user_metadata
 
 if TYPE_CHECKING:
     from grouper.models.base.session import Session
+    from typing import Dict, Optional
 
 
 class ServiceAccountRepository:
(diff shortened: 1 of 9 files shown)


grouper/repositories/service_account.py
```diff
@@ -52,8 +54,15 @@ def create_service_account(self, name, owner, machine_set, description):
...
         service.add(self.session)
 
-        # Flush the account to allocate an ID, and then create the linkage to the owner.
+        # Flush the account to allocate an ID.
         self.session.flush()
+
+        # Set initial user metadata fields if present.
+        if initial_metadata is not None:
+            for key, value in initial_metadata.items():
+                set_user_metadata(self.session, user.id, key, value)
```
cugini-dbx, line 63: This is probably not the "cleanest" solution, and eventually should be cleaned up by adding a `UserMetadata` service and repository later.

wrodriguez-dbx: Same thought; mixing in non-hexagonal code like this is a bit unfortunate. Not the hugest deal in this case I think since the `set_user_metadata` function is pretty clean -- it's pretty much exactly what a `UserMetadata` repo would be doing anyway -- but could you leave a comment or TODO noting it?

wrodriguez-dbx: Ah relatedly, `set_user_metadata` uses a raw assert here to check the key against a regex: https://github.com/dropbox/merou/blob/198a3a9cba56170f3896dd91ee3dddc5f7c11286/grouper/user_metadata.py#L55

Probably worth making that a nicer catchable exception

tests/usecases/create_service_account_test.py
```diff
@@ -90,6 +92,42 @@ def test_success(setup):
...
+    assert len(initial_metadata) == len(actual_items)
+    for key, value in initial_metadata.items():
+        assert key in actual_items
+        assert initial_metadata[key] == actual_items[key]
+
+    usecase = setup.usecase_factory.create_create_service_account_usecase("gary@a.co", mock_ui)
+    usecase.create_service_account(
+        "another_service@svc.localhost", "some-group", "machine-set", "description"
+    )
+    assert mock_ui.mock_calls[-1] == call.created_service_account(
```
wrodriguez-dbx, line 121: Not sure this check adds value since it's covered by the basic `test_success` case, but :shrug:

wrodriguez-dbx (approved):
Preemptively approving for when the tests pass :)

PR #648, opened 2020-09-17
disallow setting user metadata fields that do not have options

cugini-dbx:
Options are derived from the metadata_options field in the settings.
This prevents user_metadata set by integrations/admins from being
changed. At some point we may want to add more configuration options
around user metadata fields, but this seems reasonable for now.

### DIFF
--- a/grouper/fe/handlers/user_metadata.py
+++ b/grouper/fe/handlers/user_metadata.py
@@ -58,7 +58,13 @@ class UserMetadata(GrouperHandler):
             metadata_key, DEFAULT_METADATA_OPTIIONS
         )
 
-        self.render("user-metadata.html", form=form, user=user, metadata_key=metadata_key)
+        self.render(
+            "user-metadata.html",
+            form=form,
+            user=user,
+            is_enabled=known_field,
+            metadata_key=metadata_key,
+        )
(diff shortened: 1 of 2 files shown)


grouper/fe/handlers/user_metadata.py
```diff
@@ -92,6 +98,7 @@ def post(self, *args: Any, **kwargs: Any) -> None:
                 form=form,
                 user=user,
                 metadata_key=metadata_key,
+                set_enabled=known_field,
```
wrhall, line 101: I find `set_enabled` to be a bit unintuitive as a name. Would `is_enabled` be a better fit?

wrhall: (Probably my issue is with how overloaded the word `set` is)

wrhall (approved):
Seems simple enough

grouper/fe/handlers/user_metadata.py
cugini-dbx: Yeah I agree, I struggled a bit with naming that. I like `is_enabled` better.

wrodriguez-dbx (approved):
LGTM

PR #649, opened 2020-10-03
Make permission requests expand to permission aliases

yasaswyk:
Currently, grantable permissions don't take any permission alias into account. So when looking to see who can grant a permission where the `grouper.permission.grant` is given as an alias, that request goes to grouper admin instead of the appropriate owner of the group. This will fix that bug.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,7 @@ RUN apt-get install -y chromium-driver
 RUN apt-get install -y procps unzip wget
 RUN apt-get install -y libcurl4-openssl-dev libssl-dev
 
+ENV DB mysql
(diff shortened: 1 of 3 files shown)


wrodriguez-dbx (approved):
It's unfortunate to not have test coverage here, particularly around the behavior you care about (permission requests getting routed according to aliased `grouper.permission.grant`s). If you're feeling up to it adding coverage would be great, but given that this functionality still needs to be ported to new architecture I don't think it's strictly necessary.

Mind putting a little text in the PR description about the motivation for this change? So we don't get confused down the line if this becomes relevant or breaks something :)

PR #653, opened 2020-11-18
Expose Audit information about permissions

m-ynk:
Expose whether grouper permissions are audited or not by relevant helper methods which are then used by the Grouper API to expose this information.

### DIFF
--- a/grouper/graph.py
+++ b/grouper/graph.py
@@ -566,7 +566,11 @@ class GroupGraph:
         # type: (str, bool) -> Dict[str, Dict[str, Any]]
         """ Get a permission and what groups and service accounts it's assigned to. """
         with self.lock:
-            data = {"groups": {}, "service_accounts": {}}  # type: Dict[str, Dict[str, Any]]
+            data = {
+                "groups": {},
+                "service_accounts": {},
+                "audited": {},
+            }  # type: Dict[str, Dict[str, Any]]
 
             # Get all mapped versions of the permission. This is only direct relationships.
             direct_groups = set()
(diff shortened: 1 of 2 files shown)


rohansharma (approved):
lgtm!

grouper/graph.py
```diff
@@ -611,6 +615,13 @@ def get_permission_details(self, name, expose_aliases=True):
                         else:
                             data["service_accounts"][account] = {"permissions": [details]}
 
+            # Add permission audit value
+            permission_audited = False
+            for p in self._permissions.values():
```
lfaraone, line 620: why are we enumerating `.values()` here? for reference: 
```
        self._permissions = {}  # type: Dict[str, Permission]
```

grouper/graph.py
```diff
@@ -611,6 +615,13 @@ def get_permission_details(self, name, expose_aliases=True):
...
+                if p.name == name:
+                    permission_audited = p.audited
+
+            data["audited"] = {"audited": permission_audited}
```
lfaraone, line 624: this would mean `data["audited"]["audited"]` is a bool? can you provide some context as to what the thinking was here, as the double-nesting seems a bit strange.

grouper/graph.py
m-ynk: sure! makes sense to iterate over the keys instead and only extract the relevant Permission value.

grouper/graph.py
m-ynk: Agreed that it's a bit odd! I just wanted to keep the return values consistent with the function type signature.
# type: (str, bool) -> Dict[str, Dict[str, Any]]
I can simplify if this is confusing

grouper/graph.py
lfaraone: does `self._permissions[name].audited` work? we can just set this inline on L572 then.

grouper/graph.py
lfaraone: Given that this becomes API surface area we need to support "forever", externally logical (e.g. just make it a bool value, not a nested dict) beats keeping one type signature the same

PR #656, opened 2020-12-29
Add ability to search for permission by it's argument

yasaswyk:
This diff allows searching for `perm=arg` in the search bar; this will result in showing only the results which are granted the `perm`  permission with the argument `arg`

### DIFF
--- a/grouper/fe/handlers/permission_view.py
+++ b/grouper/fe/handlers/permission_view.py
@@ -39,8 +39,8 @@ class PermissionView(GrouperHandler, ViewPermissionUI):
 
     def get(self, *args: Any, **kwargs: Any) -> None:
         name = self.get_path_argument("name")
-        permission_arg = self.get_argument("permission_arg", None)
+        argument = self.get_argument("argument", None)
         usecase = self.usecase_factory.create_view_permission_usecase(self)
         usecase.view_permission(
-            name, self.current_user.username, audit_log_limit=20, permission_arg=permission_arg
+            name, self.current_user.username, audit_log_limit=20, argument=argument
         )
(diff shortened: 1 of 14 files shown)


grouper/fe/handlers/search.py
```diff
@@ -25,13 +27,16 @@ def get(self):
...
+        permission_query = query.split("=")[0] if "=" in query else query
         permissions = (
             self.session.query(
                 label("type", literal("Permission")),
                 label("id", Permission.id),
                 label("name", Permission.name),
             )
-            .filter(Permission.enabled == True, Permission.name.like("%{}%".format(query)))
+            .filter(
+                Permission.enabled == True, Permission.name.like("%{}%".format(permission_query))
```
wrodriguez-dbx, line 38: Note that because the argumented search only triggers when there's only a single result, you won't be able to do an argumented search for any permission whose name is contained in another permission or group name. For example, if there are permissions `foo` and `foo.bar`, trying to search for `foo=baz` will turn up both permissions and thus will not have the special argumented search behavior.

Honestly I think it's fine, the feature is still an improvement and it'll be possible to manually tack on `?permission_arg=baz` on the permission view page. Alternatively could change the behavior by making the search be for an exact match if there is a `=`, but that behavior doesn't deem ideal either

grouper/fe/handlers/permission_view.py
```diff
@@ -39,5 +39,8 @@ def viewed_permission(
 
     def get(self, *args: Any, **kwargs: Any) -> None:
         name = self.get_path_argument("name")
+        permission_arg = self.get_argument("permission_arg", None)
```
wrodriguez-dbx, line 42: How about using just `argument` instead of `permission_arg`? Easier to guess if you're trying to manually edit the URL :D

wrodriguez-dbx (approved):
With a few comments

grouper/fe/handlers/permission_view.py
yasaswyk: Done

PR #657, opened 2021-01-06
Paginate the permissions page

yasaswyk:
This PR splits the permissions granted to a group and service account into 2 different pages. It also paginates them so that the UI doesn't die when where are a lot of elements.

Transferring the URL args from the permissions page to the group/service account page to keep it consistent with https://github.com/dropbox/merou/pull/656

Screen Shot 2021-01-14 at 5 58 07 PM
Screen Shot 2021-01-05 at 10 44 03 PM
Screen Shot 2021-01-05 at 10 44 12 PM


grouper/fe/templates.py
```diff
@@ -86,14 +87,42 @@ def _get_form_alerts(self, errors: Dict[str, List[str]]) -> List[Alert]:
...
 
     template: InitVar[str] = "permission.html"
 
 
+@dataclass(repr=False, eq=False)
+class PermissionGroupGrantsTemplate(BaseTemplate):
+    permission: Permission
+    access: PermissionAccess
+    grants: List[GroupPermissionGrant]
+    audit_log_entries: List[AuditLogEntry]
```
wrodriguez-dbx, line 101: Slight potential for confusion because the audit log entries will be for the permission generally (so not just for grants to groups), but that's the least of the problems with the audit log :D

grouper/fe/templates/permission-group.html
```diff
@@ -0,0 +1,176 @@
...
+{% from 'macros/ui.html' import audit_log_panel, paginator, dropdown with context %}
+
+{% block subtitle %} permission {{ permission.name }}{% endblock %}
+
+{% block heading %}
+Permissions
+{% endblock %}
+
+{% block subheading %}
+{{ permission.name }} {{total}} permission(s)
```
wrodriguez-dbx, line 11: I think `permission(s)` -> `grant(s)` since that's what the page is displaying now. (Same for the permission-service-account template below.)

grouper/fe/templates/permission-group.html
```diff
@@ -0,0 +1,176 @@
...
+{% endblock %}
+
+{% block headingbuttons %}
```
wrodriguez-dbx, line 14: Not totally convinced that the grant-listing pages should have these buttons (or maybe even the audit log) -- seems easy to confuse the grants pages with the main permission page otherwise?

grouper/fe/templates/permission.html
```diff
@@ -58,56 +58,30 @@
             <table class="table table-elist">
                 <thead>
                     <tr>
-                        <th class="col-sm-2">Group</th>
-                        <th class="col-sm-5">Argument</th>
+                        <th class="col-sm-2">Permission Grants</th>
+                        <th class="col-sm-5">Type</th>
```
wrodriguez-dbx, line 62: Don't think this second column is necessary at all, given that the first column already specifies the type

grouper/fe/templates/permission.html
```diff
@@ -58,56 +58,30 @@
...
-                    <tr class="no-service-grants">
-                        <td colspan="2" class="text-center"><em>This permission is not mapped to
-                        any service accounts. To grant this permission to a service account, visit
-                        the service account page.</em></td>
-                    </tr>
-                {% endif %}
-                </tbody>
+                <tr class="permission-grant-row">
+                    <td class="grant-group">
+                        <a href="{{transfer_qs('/permissions/', permission.name, '/groups')}}">Permission Group Grants</a>
```
wrodriguez-dbx, line 69: How about `Grants to Groups` (and `Grants to Service Accounts`) instead? A little clearer IMO

grouper/repositories/permission_grant.py
```diff
@@ -54,7 +61,38 @@ def grant_permission_to_service_account(self, permission, argument, service):
 
     def group_grants_for_permission(self, name, include_disabled_groups=False, argument=None):
         # type: (str, bool, Optional[str]) -> List[GroupPermissionGrant]
-        return self.repository.group_grants_for_permission(name, include_disabled_groups, argument)
+        grants = self.repository.group_grants_for_permission(
```
wrodriguez-dbx, line 64: Out of curiosity, any particular reason for this change? Debugging artifact?

grouper/repositories/permission_grant.py
```diff
@@ -54,7 +61,38 @@ def grant_permission_to_service_account(self, permission, argument, service):
...
+
+    def group_paginated_grants_for_permission(
+        self, name, pagination, include_disabled_groups=False, argument=None
+    ):
+        # type: (str, Pagination, bool, Optional[str]) -> PaginatedList[GroupPermissionGrant]
+        grants = self.repository.group_grants_for_permission(
+            name, include_disabled_groups, argument
+        )
+
+        print(grants)
```
wrodriguez-dbx, line 77: Debug print :)

grouper/repositories/permission_grant.py
```diff
@@ -54,7 +61,38 @@ def grant_permission_to_service_account(self, permission, argument, service):
 
     def group_grants_for_permission(self, name, include_disabled_groups=False, argument=None):
         # type: (str, bool, Optional[str]) -> List[GroupPermissionGrant]
-        return self.repository.group_grants_for_permission(name, include_disabled_groups, argument)
+        grants = self.repository.group_grants_for_permission(
+            name, include_disabled_groups, argument
+        )
+        return grants
+
+    def group_paginated_grants_for_permission(
```
wrodriguez-dbx, line 69: Nit: can we add an optional pagination argument to `group_grants_for_permission` instead of making a separate function? Feels a bit cleaner, but feel free to disagree
(And same for the equivalent service account functions below, and for the equivalent functions in the service/usecase layers instead of the repository layer)

grouper/repositories/permission_grant.py
```diff
@@ -24,13 +25,19 @@
...
     from typing import Dict, Iterable, List, Optional
 
 
 class GraphPermissionGrantRepository(PermissionGrantRepository):
     """Graph-aware storage layer for permission grants."""
 
+    SORT_FIELD = {
+        PermissionGrantSortKey.GROUP: "group",
+        PermissionGrantSortKey.SERVICE_ACCOUNT: "service_account",
+    }
```
wrodriguez-dbx, line 39: Hmm, not sure how I feel about using the sort keys in this way. I think we could instead have only a single sort key ("name" probably) and simplify things that way (e.g. removes the need for this dict at all). There are separate methods below anyways, so it's not like you actually need the sort key to distinguish between group grants vs. service account grants.

grouper/repositories/permission_grant.py
yasaswyk: Having an optional arg would make the return value a `Union`, which makes it a little annoying by introducing more casts

grouper/repositories/permission_grant.py
yasaswyk: Changing it to name involves changing the data class and lots of other places, which is a little annoying. I refactored to avoid the dict. Let me know if that's unsatisfactory

grouper/repositories/permission_grant.py
```diff
@@ -56,6 +58,33 @@ def group_grants_for_permission(self, name, include_disabled_groups=False, argum
...
+    ):
+        # type: (str, Pagination, bool, Optional[str]) -> PaginatedList[GroupPermissionGrant]
+        grants = self.repository.group_grants_for_permission(
+            name, include_disabled_groups, argument
+        )
+
+        if pagination.sort_key != PermissionGrantSortKey.NONE:
+            grants = sorted(
+                grants,
+                key=lambda p: getattr(p, pagination.sort_key.name.lower()),
```
wrodriguez-dbx, line 72: Will this break with a `sort_key` of "service_account"?

grouper/entities/pagination.py
```diff
@@ -62,3 +62,9 @@ class ListPermissionsSortKey(Enum):
     NONE = "none"
     NAME = "name"
     DATE = "date"
+
+
+class PermissionGrantSortKey(Enum):
```
wrodriguez-dbx, line 67: I do still think this is a bit unfortunate -- namely it makes it possible to have a sort key of "group" when actually looking at service account permission grants or vice versa -- but not the biggest deal. How about putting a comment here noting that this may get used for requests for only one of group grants / service account grants (and thus the value might not always make sense)?

wrodriguez-dbx (approved):

grouper/entities/pagination.py
yasaswyk: I've split the sortkey to avoid this problem

PR #662, opened 2021-03-11
Catch db engine creation failures

m-ynk:
Failures during db engine creation end up leaking the url parameter in the exception logs.
The url parameter is sensitive and can contain credentials to connect to the database.

Catch and throw a Grouper specific exception which masks any sensitive information.

### DIFF
--- a/grouper/models/base/session.py
+++ b/grouper/models/base/session.py
@@ -3,7 +3,10 @@ import logging
 from typing import TYPE_CHECKING
 
 from sqlalchemy import create_engine
+from sqlalchemy.exc import ArgumentError, OperationalError
 from sqlalchemy.orm import Session as _Session, sessionmaker
+
+from grouper.settings import InvalidSettingsError
 
 if TYPE_CHECKING:
     from sqlalchemy.engine import Engine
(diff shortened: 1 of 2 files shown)


grouper/models/base/session.py
```diff
@@ -31,7 +34,12 @@ def wrapper(self, *args, **kwargs):
 
 def get_db_engine(url):
     # type: (str) -> Engine
-    return create_engine(url, pool_recycle=300)
+    try:
+        engine = create_engine(url, pool_recycle=300)
+    except (ArgumentError, OperationalError):
+        logging.exception("Can't create database engine.")
```
wrodriguez-dbx, line 40: The lack of detail here would make debugging such an exception quite difficult. I'm not thinking of a great way to safely log anything though, in case of arbitrarily malformed values of `url` (that still contain the URL)...

wrodriguez-dbx (approved):
Stamping (it's a necessary fix), but I do wish there were a way to improve the debug-ability here.

grouper/models/base/session.py
m-ynk: Yeah, I played around various ideas for a bit but I couldn't think of a solution to safely catch and sanitize all outputs.

PR #668, opened 2021-06-01
Add use case for programmatically adding permissions to groups

rohansharma:
See https://jira.dropboxer.net/browse/PSEC-174. 

Still in the RFC stage. One quick question: should we tighten the permissiveness here even more?

### DIFF
--- a/grouper/usecases/grant_permission_to_group.py
+++ b/grouper/usecases/grant_permission_to_group.py
@@ -79,7 +79,6 @@ class GrantPermissionToGroup:
 
         actor_grantable_perms = []  # type: List[Tuple[str, str]]
         if self.service_account_service.service_account_exists(self.actor):
-            owner = self.service_account_service.owner_of_service_account(self.actor)
             p = self.service_account_service.permissions_grantable_by_service_account(self.actor)
             actor_grantable_perms = p  # line length :( if you see a nicer way, do it
         else:  # actor is not a service account, and is thus a normal user
@@ -112,7 +111,10 @@ class GrantPermissionToGroup:
                 allowed = True
                 break
         if not allowed:
-            message = "Permission denied"
+            message = (
+                "Permission denied. Actor {actor} does not have the ability to"
+                "grant permission {permission} with argument {argument}."
+            ).format(actor=self.actor, permission=permission, argument=argument,)
             self.ui.grant_permission_to_group_failed_permission_denied(
                 permission, argument, group, message
             )
--- a/grouper/usecases/grant_permission_to_group.py
+++ b/grouper/usecases/grant_permission_to_group.py
@@ -120,15 +120,11 @@ class GrantPermissionToGroup:
 
         with self.transaction_service.transaction():
             try:
-                self.group_service.grant_permission_to_group(
-                    permission, argument, group
-                )
+                self.group_service.grant_permission_to_group(permission, argument, group)
             except PermissionNotFoundException:
                 # It should be impossible to hit this exception. In order to get this far, the
                 # perm must be on the list of perms the actor can grant, and thus must exist.
                 # Leaving the logic here however in case that changes in the future.
-                self.ui.grant_permission_to_group_failed_permission_not_found(
-                    permission, group
-                )
+                self.ui.grant_permission_to_group_failed_permission_not_found(permission, group)
                 return
         self.ui.granted_permission_to_group(permission, argument, group)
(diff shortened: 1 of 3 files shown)


grouper/usecases/grant_permission_to_group.py
```diff
@@ -0,0 +1,130 @@
...
+        self.transaction_service = transaction_service
+
+    def permissions_grantable(self):
+        # The actor can grant a permission to the group if the actor independently has the ability to
+        # to grant the permission, as a permission admin or because of grants of
+        # grouper.permission.grants.
+
+        actor_grantable_perms = []  # type: List[Tuple[str, str]]
+        if self.service_account_service.service_account_exists(self.actor):
+            owner = self.service_account_service.owner_of_service_account(self.actor)
```
wrodriguez-dbx, line 82: Nit: unused I think?

grouper/usecases/grant_permission_to_group.py
```diff
@@ -0,0 +1,130 @@
...
+            return
+
+        allowed = False
+        grantable = self.permissions_grantable()
+        for grantable_perm, grantable_arg in grantable:
+            if grantable_perm == permission and matches_glob(grantable_arg, argument):
+                allowed = True
+                break
+        if not allowed:
+            message = "Permission denied"
```
wrodriguez-dbx, line 115: A little more detail here to help debugging down the line could be useful -- something like "Actor {} does not have the ability to grant permission {} with argument {}" or "To grant a permission to a group, the user must either be a permission admin or have an appropriate grouper.permission.grant permission"

tests/usecases/grant_permission_to_group_test.py
```diff
@@ -0,0 +1,198 @@
...
+from grouper.constants import ARGUMENT_VALIDATION, PERMISSION_ADMIN, PERMISSION_GRANT
+from grouper.entities.permission_grant import GroupPermissionGrant
+
+if TYPE_CHECKING:
+    from tests.setup import SetupTest
+
+
+def test_permissions_grantable(setup):
+    # type: (SetupTest) -> None
+    with setup.transaction():
```
wrodriguez-dbx, line 14: Nit: a few of these are unused (the service accounts, `permless@a.co` at least)

tests/usecases/grant_permission_to_group_test.py
```diff
@@ -0,0 +1,198 @@
...
+        "zorkian@a.co", mock_ui
+    )
+    expected = sorted([(perm, "*") for perm in all_permissions])
+    assert usecase.permissions_grantable() == expected
+
+    usecase = setup.usecase_factory.create_grant_permission_to_group_usecase("gary@a.co", mock_ui)
+    assert usecase.permissions_grantable() == []
+
+
+def test_success(setup):
```
wrodriguez-dbx, line 55: For completeness, could add a success case of granting where the actor is a service account

tests/usecases/grant_permission_to_group_test.py
```diff
@@ -0,0 +1,198 @@
...
+    usecase.grant_permission_to_group("some-permission", "argument", "some-group")
+    assert mock_ui.mock_calls == [
+        call.grant_permission_to_group_failed_permission_denied(
+            "some-permission", "argument", "some-group", ANY
+        )
+    ]
+    _assert_sane_permission_denied_message(mock_ui)
+
+    # Add the user to the group, but don't delegate the permission to the group.  The user should
+    # now be able to grant permissions in general, but not this permission in particular.
```
wrodriguez-dbx, line 171: I think some of these comments are from the grant to service account usecase and don't apply here? E.g. being a member of the group in this case does not confer the ability to grant any permissions whatsoever

wrodriguez-dbx (commented):
Generally LGTM. Two high level things though:
 - Adding better support for direct permission grants to groups presumably means we'll be making heavier use of such direct grants. Direct grants, however, have a big flaw at the moment: they don't leave any record of the reason for granting the permission. It'd be nice to see that fixed before expanding usage so much. (One hacky way to fix would be to have this usecase create a perm request behind the scenes and then approve it with a reason; the more consistent but more complex fix would be to add a reason field to all perm grants.)
 - We should convert any existing permission-grant-to-group logic (whatever the UI uses in particular) to use this new usecase for consistency.

grouper/usecases/grant_permission_to_group.py
rohansharma: Oops, you're right.

rohansharma:
Thanks @wrodriguez-dbx for the review!

Yup, I'm all for making sure a reason field eventually gets added. We should unblock https://jira.dropboxer.net/browse/PSEC-174 for the moment though, as their use case is somewhat limited.

wrodriguez-dbx (approved):
Chatted a bit more offline. The high-level questions don't actually block this PR since it only adds the initial usecase which ought to be done regardless.

PR #672, opened 2021-10-19
Update error message when failing to grant a perm with empty argument

wrodriguez-dbx:
The dropbox having a suggested argument in parens but not auto-filling the argument field confuses approximately everyone. Update the error message to at least be more clear about what failed.

Same problem as https://github.com/dropbox/merou/pull/671, just a different approach.


wrodriguez-dbx:
What the new error messages look like
Screen Shot 2021-10-19 at 3 26 58 PM
:
Screen Shot 2021-10-19 at 3 27 15 PM

PR #674, opened 2022-03-25
update dependencies for security alert and build error

jeichenhofer:


jeichenhofer:
`py.path` is in [maintenance mode](https://py.readthedocs.io/en/latest/path.html) and on track to be deprecated. When we upgrade to the non-vulnerable version of the library, this includes the updated naming convention to the [`LocalPath` object](https://py.readthedocs.io/en/latest/_modules/py/_path/local.html#LocalPath) as a private member.

jeffpuchalski (approved):

PR #683, opened 2022-11-18
introduce basic regex log redaction

jeichenhofer:
In response to the recent incident regarding database password leaks, this PR adds a log formatter to all python loggers which will match plaintext db url strings and redact them from the log output. It does not eliminate the risk, but it adds an additional layer of protection against accidental logging of sensitive credentials.

### DIFF
--- a/grouper/log_redact.py
+++ b/grouper/log_redact.py
@@ -2,6 +2,7 @@ from dataclasses import dataclass
 import logging
 import re
 
+
 @dataclass
 class RedactionPattern:
     # regular expression to match for redaction
@@ -9,12 +10,9 @@ class RedactionPattern:
     # sensitive string will be replaced by this "redacted_message" in the log output
     redacted_message: str
 
+
 class RedactingFormatter(logging.Formatter):
-    """
-    RedactingFormatter masks sensitive regular expressions from log entries.
-    
-    Configure the sensitive log regular expressions in settings.log_redact_patterns.
-    """
+    """RedactingFormatter masks sensitive regular expressions from log entries."""
 
     _REDACT_PATTERNS = [
         RedactionPattern(  # mysql url
@@ -31,12 +29,17 @@ class RedactingFormatter(logging.Formatter):
         ),
     ]
 
-    def _filter(self, log_string: str) -> str:
+    def _redact(self, log_string: str) -> str:
         new_string = log_string
         for pattern in RedactingFormatter._REDACT_PATTERNS:
-            new_string = re.sub(pattern.regex, f"<{pattern.redacted_message}>", new_string)
+            new_string = re.sub(
+                pattern.regex,
+                f"<{pattern.redacted_message}>",
+                new_string,
+                flags=re.IGNORECASE,
+            )
         return new_string
-    
+
     def format(self, record: logging.LogRecord) -> str:
-        original_log_message = logging.Formatter.format(self, record)
-        return self._filter(original_log_message)
+        original_log_message = super().format(record)
+        return self._redact(original_log_message)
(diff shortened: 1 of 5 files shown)


grouper/setup.py
```diff
@@ -51,15 +53,13 @@ def build_arg_parser(description):
...
+    log_level = get_loglevel(args)
 
     sa_log = logging.getLogger("sqlalchemy.engine.base.Engine")
 
     logging.basicConfig(level=log_level, format=log_format)
     if log_level < 0:
         sa_log.setLevel(logging.INFO)
+
+    for handler in logging.root.handlers:
+        handler.setFormatter(RedactingFormatter(log_format))
```
jussi-dbx, line 65: If you wanted to retain the original formatters, you could use logic similar to [this blog post](https://relaxdiego.com/2014/07/logging-in-python.html) that sets up the formatter with all the other formatters, and applies those first before applying the redactions. You would also need to adjust the logic to `init` the class + the `format` function to take into account what formatters the `RedactingFormatter` was initialized with.

grouper/settings.py
```diff
@@ -79,6 +79,7 @@ def __init__(self):
         self.http_proxy_host = None  # type: Optional[str]
         self.http_proxy_port = None  # type: Optional[int]
         self.log_format = "%(asctime)-15s\t%(levelname)s\t%(message)s  [%(name)s]"
+        self.log_redact_patterns = [] # type: List[str]
```
jussi-dbx, line 82: This seems unused?

grouper/log_redact.py
```diff
@@ -0,0 +1,42 @@
...
+    # regular expression to match for redaction
+    regex: str
+    # sensitive string will be replaced by this "redacted_message" in the log output
+    redacted_message: str
+
+class RedactingFormatter(logging.Formatter):
+    """
+    RedactingFormatter masks sensitive regular expressions from log entries.
+    
+    Configure the sensitive log regular expressions in settings.log_redact_patterns.
```
jussi-dbx, line 16: Aren't these hardcoded now?

grouper/log_redact.py
```diff
@@ -0,0 +1,42 @@
...
+        RedactionPattern(  # anything with password=
+            r"password=\S+",
+            "PASSWORD_REDACTED",
+        ),
+    ]
+
+    def _filter(self, log_string: str) -> str:
+        new_string = log_string
+        for pattern in RedactingFormatter._REDACT_PATTERNS:
+            new_string = re.sub(pattern.regex, f"<{pattern.redacted_message}>", new_string)
```
jussi-dbx, line 37: I believe both SQL server and MySQL connection strings are not case sensitive, so you might want to use `flags=re.IGNORECASE` here

grouper/settings.py
jeichenhofer: Yeah. Leftover from when I had it in the config/settings files. I'll remove it.

grouper/log_redact.py
jeichenhofer: Yeah. I'll rewrite the comment. I had originally thought to allow this in settings files, but I figured this would be easier and we'd gain little benefit from configuring it in the configs.

grouper/log_redact.py
jeichenhofer: Good call. Done.

grouper/setup.py
jeichenhofer: I've looked into this quite a bit, and the typing system makes it very complicated. While the native python type system will gladly work with the `get_attr` approach to "proxying" the objects, mypy refuses it (with good reason). I'm going to stick with an approach similar to what I have here, since it works with the more common object-oriented type system that mypy presents at the surface.

akorzo-dbx (approved):

